PR #1, opened 2020-05-30
Python Wrapper for VMEC

zhucaoxiang:
@landreman @ejpaul
I cleaned up my implementation of python wrapper for VMEC. I tested it on the PPPL cluster with python 3.7. Right now, there are not too many interfaces to different data, but should be straightforward to extend.

I didn't use a submodule to include VMEC sources, because it is impossible to include one subfolder from a repo as a submodule. I think it would be acceptable to download VMEC from STELLOPT first.


zhucaoxiang:
Brief documentation can be found at https://github.com/hiddenSymmetries/simsopt/blob/f2py/modules/VMEC/README.md

PR #2, opened 2020-06-08
Advanced interface with VMEC indata namelist

zhucaoxiang:
Leave f90wrap only to interface VMEC indata namelist.


zhucaoxiang:
Merge it for now as it was also tested by Paul.

PR #10, opened 2020-10-27
MPI and array-valued objectives

landreman:
This update has 2 features. The first is MPI-parallelized finite-different gradient computation, very similar to stellopt. You can now have multiple VMEC jobs running concurrently, each using multiple processors. For now the MPI communication is coded into simsopt directly; we may want to abstract out the communication later to make things more flexible.

The second update is that the functions going into a LeastSquaresTerm no longer need to return scalars - they can also return 1D numpy arrays. The idea is that if we have a bunch of related terms in the objective function, like the Bmn's for quasisymmetry, they can be represented by a single LeastSquaresTerm.


mbkumar (approved):
Hello Matt,

Everything looks good to me.

mbkumar:
Thank you Matt!

PR #11, opened 2020-11-11
Simplified API, refactor, Garabedian

landreman:
This update has 3 main improvements:
- Users can now specify a least-squares objective function in a less verbose way using standard tuples (function, goal, weight) instead of LeastSquaresTerm objects.
- Some refactoring was done so Dofs and LeastSquaresProblem do not depend at all on MPI or the least-squares solver. This new structure is better because we can now tinker with the communication and optimization algorithms without touching the fundamental Dofs and LeastSquaresProblem classes.
- The Garabedian surface representation was implemented, and it is now possible to optimize in the space of Garabedian coefficients instead of VMEC RBC/ZBS coefficients.


mbkumar (approved):
Looks good to me. I am cleaning up the previous versions of some classes. Will incorporate this push request beforeadding thm.

PR #12, opened 2020-11-11
Cleanup of least squares problem code

mbkumar:
Cleaned up the code in least-squares-problem code. Changes introduced are:

1)  LeastSquaresTerm init signature is modified to make it work with tuples better. 
2) x setter in LeastSquaresProblem.
3) Checkers in utils.py cleaned up.

@landreman, Change 1 would simplify the code to instantiate the LeastSquaresTerm quite a bit. No need for bloated validation.


landreman:
These changes seem fine to me. Thanks for taking a look.

If LeastSquaresProblem now treats the state vector x as a `@property` with a setter instead of using a `set()` function, we might consider making the same change to `Dofs.set()` for consistency.

PR #16, opened 2021-01-23
SPEC interface

landreman:
This update adds an interface to SPEC (via subprocess call) using the python packages `py_spec` and `pyoculus`. This is the version of the code that was successfully used for eliminating the magnetic islands in the Aten configuration.


landreman:
The CI tests are failing here because a "secret" is needed to check out the SPEC repo (since it's private), and secrets are not available when a workflow is triggered from a forked repository: https://docs.github.com/en/actions/reference/encrypted-secrets#using-encrypted-secrets-in-a-workflow
All the tests pass in the CI in my fork.

PR #22, opened 2021-03-05
MPI improvements

landreman:
Removed `from mpi4py import MPI` from most modules, with MPI process instead printed to log using utils.mpi.log(). Better handling of MPI for Boozer: only group leaders call booz_xform. Log file is written out by `least_squares_mpi_solve`.


mbkumar:
@landreman 
The CI tests are failing because of errors in installing booz_xform. Could you please fix it?

mbkumar (approved):
@landreman,

The code looks good.

Bharat

landreman:
Hmm, I don't know why the python 3.7 CI job failed in the booz_xform install. The python 3.8 job installed booz_xform fine and failed where it was supposed to (SPEC install). I re-ran the 3.7 one and it installed booz_xform fine. Must have been a fluke with pypi or something.

PR #28, opened 2021-03-11
More curve & biot savart features and testing

florianwechsung:
A few more tests for the biot savart kernel,  ability to only compute derivative wrt to B, and least squares fitting for curves


landreman:
With the new curverzfourier and curvexyzfourier, should fouriercurve.cpp/.py and magneticaxis.cpp/.py be removed?

florianwechsung:
Yeah good point, this is a left over from me copying files manually from simsgeo. I have now switched fully to simsopt, so these sort of things should not happen any more in the future.

PR #33, opened 2021-03-19
WIP: DOF refactoring

mbkumar:
Introduction
----------------
This pull request introduces refactored DOF and Optimizable classes that use graph links to connect multiple Optimizable classes. Both encapsulation and inheritance are used to define the optimization problems in a modular fashion. The final optimizable problem is also going to be an object of a subclass of the Optimizable class and is nothing but a graph of Optimizable objects. Encapsulation is used to hide all the Optimizable objects and the related data without burdening the final Optimizable problem with a vast amount of data and methods. Inheritance is used to redefine the base class behavior where appropriate and to introduce new functionality if desired. Judicious usage of encapsulation and inheritance is a trade-off left for users. 

I am pushing this into the main repository even though many things are still be implemented to prevent too much divergence from the main branch and to introduce the framework for everyone.

Changes Introduced:
-----------------------------
1. Degrees of freedom are stored as pandas DataFrame under a _DOF_ class and the capabilities of pandas are exploited for masking the unwanted data. The names of DOF are also supplied at this level and all aspects of the DOFs are condensed into this class.
2. _Optimizable_ classes and subclasses can be used to define unique optimizable blocks. They can take other _Optimizable_ objects as inputs and form the final optimization problem automatically without needing to reformulate the problem at the end.
3. Introduced framework for an _Optimizable_ object to return outputs from multiple methods to calling objects. The methods that return objectives have to be named (mandatory) by the class developer for end-user's convenience. At instantiation, the child _Optimizable_ object can supply the parent Optimizable objects along with names of required functions. Alternatively, the child _Optimizable_ object can directly specify the methods of the parent Optimazable objects. 
4. For a single parent _Optimizable_ object, each child _Optimizable_ object can ask for a different set of functions as return values. The parent _Optimizable_ object handles the connection between methods and child Optimizable objects.
5. Example classes are implemented in the _new_functions.py_ module, and are nothing but a reformulation of similar classes in _functions.py_ module
6. _Surface_ classes (based on Matt's Surface class) are rewritten to use the new framework.
7. _LeastSquaresProblem_ class is reimplemented with the new framework. LeastSquareTerm is not used anymore. The tuples approach from Matt for LeastSquaresProblem is implemented with a class method.
8. _least_squares_serial_solve_ reimplemented to use the new _LeastSquaresProblem_ class
8. _surface_volume_area_ in _examples_ folder reimplemented

Things pending
-----------------------------
1. Reimplement additional problems in _examples_ folder using the new framework. 
2. Reimplement Spec and Vmec adapter interfaces
2. Document the classes in detail and write language-agnostic documentation on this approach for general users (Bindel's suggestion)
4. Enhanced caching than the simple strategy used currently (Bindel's suggestion)
5. Implementation of derivatives (Will wait on this for a bit).


mbkumar:
1. Introduced framework for an Optimizable object to return outputs from multiple methods to calling objects. The methods that return objectives have to be named (mandatory) by the class developer for end-user's convenience. At instantiation, the child Optimizable object can supply the parent Optimizable objects along with names of required functions. Alternatively, the child Optimizable object can directly specify the methods of the parent Optimazable objects.
2. For a single parent Optimizable object, children objects can ask for different set of functions as return values. The parent Optimizable object handles the connection between methods and child Optimizable objects

The original PR comment is updated.

mbkumar:
Reimplemented least_squares_serial_solve and the surface_volume_and_area example using the new Optimizable class. All the tests are passing.

landreman (approved):

PR #36, opened 2021-03-23
Boozer surfaces

andrewgiuliani:
This pull request introduces:

1. computation of `surfaceXYZFourier` surfaces, where `(phi,theta)` correspond to the Boozer angles on the surface, by solving a nonlinear least squares problem in `boozersurface.py`:
- `minimizeBoozerScalarizedLBFGS` solves a scalarized constrained optimization problem using LBFGS, where 
```
min 0.5* || f(x) ||^2_2 + 0.5 * constraint_weight * (label - labeltarget)^2
+0.5*constraint_weight * (y(varphi=0,theta=0) -0)^2+0.5*constraint_weight * (z(varphi=0,theta=0) -0)^2
where || f(x)||^2_2 is the sum of squares of the Boozer residual at
a set of quadrature points.
```
The final two terms in the objective function are boundary conditions to ensure that the parametrization of the surface is unique.

- `minimizeBoozerScalarizedNewton` solves the same scalarized constrained optimization problem as above, but instead of LBFGS, Newton's method is used.

- `minimizeBoozerConstrainedNewton` solves the full constrained optimization problem using Newton's method.
```
min 0.5*|| f(x) ||^2_2
subject to 
label - labeltarget = 0
y(varphi=0,theta=0) - 0 = 0
z(varphi=0,theta=0) - 0 = 0
```

3. The user is free to choose either a geometric constraint, e.g. area, or a toroidal flux constraint.
4. `objectives.py` is split into `coilobjectives.py` and `surfaceobjectives.py`

Let's use this PR to further discuss the above additions to the code.


mbkumar:
Hello Andrew,

The tests are failing in CI. You are importing a new module called ipdb. If it is an external library, could you please add to setup.py requires argument?

Thanks
Bharat

mbkumar:
Hello Andrew,

simsopt adheres to Google style guideline. https://google.github.io/styleguide/pyguide.html
Could you please run a linter such as pylint or open your code in pycharm and fix any style issues identified by them?

Bharat

florianwechsung:
Thank you @andrewgiuliani  for all these additions. I should have time tomorrow to look at this in detail before we merge this.

florianwechsung:
The tests currently fail because the test data files can't be found. You probably need to do some magic with the path so that the files are guaranteed to be found. Have a look at `tests/mhd/test_vmec.py` and `tests/mhd/__init__.py` for some inspiration.

andrewgiuliani:
Hi @florianwechsung and @mbkumar, thanks for the comments.  I've just pushed a commit that partially addresses the feedback, more to come.

mbkumar:
In surface_test_helpers.py, you can replace the line
`filename = os.path.join(TEST_DIR, 'NCSX_test_data/NCSX_coil_coeffs.dat')` with
`filename = TEST_DIR / 'NCSX_test_data' / 'NCSX_coil_coeffs.dat'` and do away with importing os module.

And please don't use linux folder specifier such as  `'NCSX_test_data/NCSX_coil_coeffs.dat'`.  The former is more general and works with windows also.

landreman:
It looks like the CI is failing because some of the new tests cannot find `surface_test_helpers.py` from where the tests are run, in the `tests` directory. I believe one solution would be to add the following at the top of `tests/geo/test_boozersurface.py` and `test_surface_objectives.py`:
~~~~
import sys, os
sys.path.append(os.path.dirname(__file__))
~~~~
Feel free to suggest a more elegant solution if there is one.

mbkumar:
On the failing tests, another option would be to to use
from .surface_test_helpers import get_ncsx_data,get_surface,
get_exact_surface
instead of
from surface_test_helpers import get_ncsx_data,get_surface,
get_exact_surface

The need to add a dot arises because there is an __init__.py file present,
which makes the geo folder a package.

Bharat Medasani

Engineer

florianwechsung:
Okay I think this one is ready if @mbkumar and @landreman are also happy with it. I added a script `examples/boozer` that illustrates how to use this stuff as well.

florianwechsung (approved):

mbkumar:
Thanks for getting the tests successful.

I have few requests for this PR. The majority of them are syntactic in nature. 

1) Please bring down the line lengths to 80. If not, at least split the ultra-long lines I see few times.
2) Change _curveobjectives.py_ to _curve_objectives.py_ and update the corresponding imports and test file names. Additional file names to change are _surfaceobjectives.py_, _surfacexyztensorfourier.py_, _surfacexyzfourier.py_, _curvexyzfourier.py_, _coilcollection.py_, and _boozersurface.py_
3) If using space on one side of an operator, please use space on the other side of the operator as well. The following line is hard to read.
`J = np.concatenate( (J,  np.sqrt(constraint_weight) *dl[None,:] ,np.sqrt(constraint_weight) *dry[None,:],np.sqrt(constraint_weight) *drz[None,:]  ),  axis = 0)` 
Please change it to
`J = np.concatenate((J,  np.sqrt(constraint_weight) * dl[None, :], np.sqrt(constraint_weight) * dry[None, :], np.sqrt(constraint_weight) * drz[None, :]),  axis=0)`
4) Still seeing camel cased variable names occassionally. For example, line 211 in _test_boozersurface.py_: `boozerSurface = BoozerSurface(bs, s, tf, tf_target)`
5) `get_surface` function in _test_surface_taylor.py_ could go into main code.
6) Please change the tests class name `Testing` in _test_surface_taylor.py_ to a more useful name. 
7) Please delete line 299 `print('wtf')` in _test_surface_taylor.py_. It look unprofessional for a scientific code.

florianwechsung:
Thank you for looking over this Bharat. I have taken care of 1, 3, 4, 6, and 7.

ad 2): This pretty much applies to all the files in `simsopt.geo` and doesn't really have much to do with boozer surfaces / this specific pull request. If we want to make this change, I think it should be done separately. I've been working on a PR that allows us to enforce a selection of specific coding styles that we all agree on (e.g. white spaces etc). If we all agree that all of `simsopt.geo` should be renamed, then that seems like a more suitable PR to do so.

ad 5) I agree that we have these sort of `get_surface` and `get_curve` functions in a couple of places in the tests, but they are reasonably specific to the individual tests, and for now it would be a bit of a pain to generalise them. In the future it may be nice to have some helper functions in the main code that quickly create certain types of surfaces/coils/stellarator configurations. But I don't think it's crucial for now.

landreman:
Looks good to me overall. A few suggestions:

1. It would be good to have an overall docstring for the new BoozerSurface class, so people know what this class is for.
2. In `examples/boozer` it would be nice to have a docstring at the top that gives an overview of what is being demonstrated in this example.
3. The different Curve and Surface objects have some documentation in the C++, but it does not appear in the corresponding python classes or in the sphinx docs. Copying or moving the documentation from the C++ to the python should make it appear in the sphinx docs. (Or we could try to use `doxygen` with `breathe` to scrape comments from C++.) This might be beyond the scope of this PR, but I mention it now since there is new documentation in the C++ for e.g. `SurfaceXYZTensorFourier`.
4. You can pull in master, then add the new boozer example to `examples/run_examples` so it is run in the CI.
5. In the tests, it would be nice to have comments in each test to explain what that test is checking.

andrewgiuliani:
Hi Matt,

@florianwechsung and I have made the suggested changes to this PR.  It's now ready to be merged.

-Andrew

PR #37, opened 2021-03-29
Helical and Toroidal Magnetic Field Classes

rogeriojorge:
This pull request introduces two new classes of magnetic fields:

- A "Helical Field" produced by helical coils. This coils are created using a new HelicalCurve class.
- A "Toroidal Field" that simply introduces a 1/R dependence of B.

For this, a MagneticField class was created, for which the BioSavart class and the type of magnetic fields above inherit.


landreman:
Thanks for working on this.

Some unit tests are needed, as in `tests/geo/`. For `CurveHelical` a good test could be to verify that everything matches a CurveXYZFourier corresponding to the identical shape, as in the Mathematica notebook you showed me. The test could verify that gamma matches, and possibly that some derivatives of gamma match. For unit tests of ToroidalField, it is probably sufficient to spot check B and its derivatives at one point, for some values of B0 and R0 different from 1.

There should probably be a class that represents a sum of MagneticField objects, specified as a list to the constructor. To compute B (or A or derivatives thereof) it just delegates to each item in the list, then sums. Then HelicalField should be removed and represented as one of these: `MagneticFieldSum([<ToroidalField instance>, <BiotSavart(CurveHelical) instance>]`.

Presently the CI tests are failing due to a known issue, related to permissions for SPEC. I should address that so we can see if tests pass for incoming pull requests. For some reason the CI workflow didn't run in @rogeriojorge 's fork, not sure why.

rogeriojorge:
The magnetic field classes were refactored and both unit tests and documentation using docstrings were added

src/simsopt/geo/magneticfieldclasses.py
```diff
@@ -0,0 +1,116 @@
...
+                np.multiply(np.divide(self.B0*self.R0,R**4),-2*np.multiply(x,y)),
+                0*R))
+            dB_by_dX3=np.vstack((0*R,0*R,0*R))
+
+            dToroidal_by_dX=np.array([dB_by_dX1,dB_by_dX2,dB_by_dX3]).T
+
+            self._dB_by_dX = dToroidal_by_dX
+
+        if compute_derivatives >= 2:
+            self._d2B_by_dXdX = None # To be implemented
```
florianwechsung, line 50: Please raise a NotImplementedError here.

src/simsopt/geo/magneticfieldclasses.py
```diff
@@ -0,0 +1,116 @@
...
+                             [self.Phiparsed.diff(R).diff(Z),(self.Phiparsed.diff(Phi)/R).diff(Z),self.Phiparsed.diff(Z).diff(Z)]])
+        
+    def compute(self, points, compute_derivatives=0):
+        r   = np.sqrt(np.power(points[:,0],2) + np.power(points[:,1],2))
+        z   = points[:,2]
+        phi = np.arctan2(points[:,1],points[:,0])
+        self._B = [self.Blambdify(r[i],z[i],phi[i]) for i in range(len(r))]
+
+        if compute_derivatives >= 1:
+            self._dB_by_dX = [self.dBlambdify_by_dX(r[i],z[i],phi[i]) for i in range(len(r))]
```
florianwechsung, line 82: Please raise an error in case a hessian is required

src/simsopt/geo/magneticfieldclasses.py
```diff
@@ -0,0 +1,116 @@
...
+        alpha = np.sqrt(self.r0**2 + np.power(r,2) - 2*self.r0*rho)
+        beta  = np.sqrt(self.r0**2 + np.power(r,2) + 2*self.r0*rho)
+        k     = np.sqrt(1-np.divide(np.power(alpha,2),np.power(beta,2)))
+        self._B = np.array([
+            [self.Inorm*point[0]*point[2]/(2*alpha[i]**2*beta[i]*rho[i]**2)*((self.r0**2+r[i]**2)*ellipe(k[i]**2)-alpha[i]**2*ellipk(k[i]**2)),
+             self.Inorm*point[1]*point[2]/(2*alpha[i]**2*beta[i]*rho[i]**2)*((self.r0**2+r[i]**2)*ellipe(k[i]**2)-alpha[i]**2*ellipk(k[i]**2)),
+             self.Inorm/(2*alpha[i]**2*beta[i])*((self.r0**2-r[i]**2)*ellipe(k[i]**2)+alpha[i]**2*ellipk(k[i]**2))]
+            for i,point in enumerate(points)])
+
+## Next magnetic field classes to implement
```
florianwechsung, line 113: This should not be in the source code. Maybe open an issue if you want to keep track of our todos

tests/geo/test_magneticfields.py
```diff
@@ -0,0 +1,115 @@
...
+        Bfield    = ToroidalField(R0test,B0test)
+        Bfield.set_points(points)
+        B1        = Bfield.B()
+        dB1_by_dX = Bfield.dB_by_dX()
+        # Bfield analytical
+        B2        = np.array([(B0test*R0test/(point[0]**2+point[1]**2))*np.array([-point[1], point[0], 0.]) for point in points])
+        dB2_by_dX = np.array([(B0test*R0test/((point[0]**2+point[1]**2)**2))*np.array([[2*point[0]*point[1], point[1]**2-point[0]**2, 0],[point[1]**2-point[0]**2, -2*point[0]*point[1], 0],[0,0,0]]) for point in points])
+        # Verify
+        assert np.allclose(B1, B2)
+        assert np.allclose(dB1_by_dX, dB2_by_dX)
```
florianwechsung, line 31: Could add a check for symmetry of gradient and also that divergence is zero

tests/geo/test_magneticfields.py
```diff
@@ -0,0 +1,115 @@
...
+        coils     = [CurveHelical(101, 2, 5, 2, 1., 0.3) for i in range(2)]
+        coils[0].set_dofs(np.concatenate(([np.pi/2,0],[0,0])))
+        coils[1].set_dofs(np.concatenate(([0    ,0],[0,0])))
+        currents  = [-2.1e5,2.1e5]
+        Bhelical  = BiotSavart(coils, currents)
+        # Set up toroidal field
+        Btoroidal1 = ToroidalField(1.,1.)
+        Btoroidal2 = ToroidalField(1.2,0.1)
+        # Set up sum of the two
+        Btotal    = MagneticFieldSum([Bhelical,Btoroidal1,Btoroidal2])
```
florianwechsung, line 44: We want to be able to just write `Btotal = Bhelical + Btoroidal1 + Btoroidal2` instead of this

florianwechsung (changes requested):
This looks pretty good already. I made some comments in the code that should be addressed.

src/simsopt/geo/magneticfieldclasses.py
rogeriojorge: Issue opened

tests/geo/test_magneticfields.py
rogeriojorge: Check added for the divergence of the magnetic field. Could you be specific about which symmetry you were looking for?

florianwechsung: For the fields that we get from the Biot Savart law for coils, we know that the 3x3 matrix `A = \nabla B` is symmetric. Is this the case here as well?

rogeriojorge: I am not aware of such a symmetry.

P.S: Happy Easter!

florianwechsung:
Thanks for making all these changes!

- I noticed that you kept the magnetic potential `A` (`\curl A = B`) in `BiotSavart` and didn't move it into `MagneticField`. Is that for a reason? Can we not compute `A` for the circular coils? If possible, it would be good to add that, since we need it for calculations such as the toroidal flux and I believe Elizabeth also uses `A` in a bunch of her stuff.

- Could you also add the Hessian of B for the circular coil? We need the hessian for the near-axis quasi-symmetry optimisation and also for a few computations involving Boozer surfaces. It would be nice if those would all work with your new classes.

tests/geo/test_magneticfields.py
landreman: This symmetry should indeed be present here. This is a vacuum field so B = grad phi, hence grad B = grad grad phi which is manifestly symmetric.

rogeriojorge: Right! Test added

rogeriojorge:
The magnetic field is now computing the magnetic vector potential instead of those functions being only on Bio-Savart.
The circular coil class right now is a very simple example taken from https://ntrs.nasa.gov/citations/20010038494
I will try to add the derivatives of B this week following that article, I'll post here when they're done.

florianwechsung:
If computing the derivatives by hand is too big of a pain, you could also ask JAX to compute those for you automatically.

rogeriojorge:
Derivatives are now computed and tested for the circular coil case

rogeriojorge:
The circular coil magnetic field has now been updated to its desired behavior: calculate the magnetic field at a point given a circular coil with a center and a normal vector. The only inputs needed for the normal vector are its spherical angles theta and phi relative to the cartesian z axis due to the inherent azimuthal symmetry of the coil.

landreman:
This looks good to go to me. There have been a few commits merged into hiddenSymmetries/master, so it would be good to merge those into your fork and make sure the CI tests still pass in your fork. @florianwechsung are you happy with this PR, particularly the `MagneticFieldSum` system for combining different types of magnetic fields? (The motivation is that for a particular problem of interest to Rogerio and me, we want to add a pure toroidal field to a field from coils.)

florianwechsung (approved):

rogeriojorge:
Suggestions/improvements added after the developer's meeting.
Changes from v1.0.0 were also merged in.

landreman:
In a few places there is `ellipek2=[ellipe(k[i]**2) for i, point in enumerate(points)]`. But `ellipe` and `ellipk` are already vectorized, so `ellipek2 = ellipe(k**2)` should work and be faster, I believe. Similarly, in `magneticfieldclasses.py` lines 156-240, can't all or most of those `for i, point in enumerate(points)` loops be eliminated and sped up in the same way by removing `[i]`?

Also in `magneticfieldclasses.py` line 11, `pass` is unnecessary.

rogeriojorge:
> Similarly, in `magneticfieldclasses.py` lines 156-240, can't all or most of those `for i, point in enumerate(points)` loops be eliminated and sped up in the same way by removing `[i]`?

Can you give an example on one line where such a speed up could be implemented?

landreman:
For example, I believe lines 163-170 can be replaced by
~~~~
            dBxdx = (self.Inorm*points[:, 2]*(ellipkk2*alpha**2*((2*points[:, 0]**4 + gamma*(points[:, 1]**2 +
                    points[:, 2]**2))*r**2 + self.r0**2*(gamma*(self.r0**2 + 2*points[:, 2]**2) -
                    (3*points[:, 0]**2 - 2*points[:, 1]**2)*rho**2)) + ellipek2*(-((2*points[:, 0]**4 + gamma*(points[:, 1]**2 +
                    points[:, 2]**2))*r**4) + self.r0**4*(-(gamma*(self.r0**2 + 3*points[:, 2]**2)) + (8*points[:, 0]**2 - points[:, 1]**2)*rho**2) -
                    self.r0**2*(3*gamma*points[:, 2]**4 - 2*(2*points[:, 0]**2 + points[:, 1]**2)*points[:, 2]**2 *
                    rho**2 + (5*points[:, 0]**2 + points[:, 1]**2)*rho**4))))/(2*alpha**4*beta**3*rho**4)
~~~~
For performance it is best to let numpy vectorization handle all the loops rather than have loops in python.

rogeriojorge:
New merged changes from main repository and Matt's suggestions implemented

landreman:
Thanks for those edits. Upon closer examination, there are a bunch of other explicit for loops in `magneticfieldclasses.py`. Could you search that file for each occurrence of `for` and check whether the loop can use numpy vectorization / sums / matrix multiplication instead, as in your last commit? It may also be possible to eliminate the explicit `for` loops in `curvehelical.py` lines 11-12.

landreman (approved):

PR #42, opened 2021-04-09
Faster Biot Savart

florianwechsung:
This PR makes two main changes:

- Compile with `-march=native` when compiled locally. Compile with `-march=ivybridge`  when compiled on the CI for our wheels. Ivy Bridge supports avx but is old enough that we can probably rely on this wheel working everywhere.

- If compiled with AVX512F instructions available, then we use a Fast Inverse Square root trick to speed up the computation of `1/||diff||`, since both inverses and square roots are pretty slow on current AVX512 hardware.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -170,8 +170,14 @@ jobs:
       run: |
         pip install mpi4py
 
+    - name: Install gcc 11
+      run: sudo apt-get install gcc-11 g++-11
+
     - name: Install simsopt package
       run: pip install -v .
+      env:
+        CC: gcc-11
+        CXX: g++-11
 
     - name: Run mpi unit tests
       if: "contains(matrix.test-type, 'unit')"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -170,14 +170,8 @@ jobs:
       run: |
         pip install mpi4py
 
-    - name: Install gcc 11
-      run: sudo apt-get install gcc-11 g++-11
-
     - name: Install simsopt package
       run: pip install -v .
-      env:
-        CC: gcc-11
-        CXX: g++-11
 
     - name: Run mpi unit tests
       if: "contains(matrix.test-type, 'unit')"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,7 +44,7 @@ set(XTENSOR_USE_OPENMP 0)
 
 
 pybind11_add_module(${PROJECT_NAME}
-    src/simsgeopp/python.cpp src/simsgeopp/biot_savart.cpp src/simsgeopp/biot_savart_derivative.cpp
+    src/simsgeopp/python.cpp src/simsgeopp/biot_savart_py.cpp src/simsgeopp/biot_savart_vjp_py.cpp
     )
 set_target_properties(${PROJECT_NAME}
     PROPERTIES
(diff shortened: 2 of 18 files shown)


florianwechsung:
Here are some benchmarks on a modern Xeon Platinum CPU. Depending on problem size and number of derivatives, we get a 2x-3x speedup.

```
# march=x86-64 (old default)

Number of derivatives: 0
         N Time (in ms) Gigainteractions/s
     10000      0.03184            0.31407
   1000000        3.131            0.31939
 100000000        313.5            0.31898
Number of derivatives: 1
         N Time (in ms) Gigainteractions/s
     10000      0.12139           0.082379
   1000000       11.921           0.083886
 100000000         1192           0.083893
Number of derivatives: 2
         N Time (in ms) Gigainteractions/s
     10000      0.24222           0.041285
   1000000        23.71           0.042176
 100000000       2355.3           0.042457

# march=ivybridge (new default on CI)

Number of derivatives: 0
         N Time (in ms) Gigainteractions/s
     10000      0.02043            0.48948
   1000000        2.012            0.49702
 100000000        200.8            0.49801
Number of derivatives: 1
         N Time (in ms) Gigainteractions/s
     10000      0.07475            0.13378
   1000000        7.193            0.13902
 100000000        720.3            0.13883
Number of derivatives: 2
         N Time (in ms) Gigainteractions/s
     10000       0.1397           0.071582
   1000000       13.532           0.073899
 100000000       1345.7           0.074311

# march=native (new default for local installation)
 Number of derivatives: 0
         N Time (in ms) Gigainteractions/s
     10000      0.02685            0.37244
   1000000        1.943            0.51467
 100000000        191.5            0.52219
Number of derivatives: 1
         N Time (in ms) Gigainteractions/s
     10000      0.06174            0.16197
   1000000        3.645            0.27435
 100000000        362.9            0.27556
Number of derivatives: 2
         N Time (in ms) Gigainteractions/s
     10000      0.22623           0.044203
   1000000       10.436           0.095822
 100000000         1029           0.097182

# march=native + fast inverse square root avx512
 Number of derivatives: 0
         N Time (in ms) Gigainteractions/s
     10000      0.02157            0.46361
   1000000         1.38            0.72464
 100000000        138.7            0.72098
Number of derivatives: 1
         N Time (in ms) Gigainteractions/s
     10000      0.05904            0.16938
   1000000        3.424            0.29206
 100000000        340.3            0.29386
Number of derivatives: 2
         N Time (in ms) Gigainteractions/s
     10000      0.21749           0.045979
   1000000        9.531            0.10492
 100000000        972.7            0.10281
```

florianwechsung:
I also took this opportunity to clean up the biot savart code a bit to separate the version that uses `xtensor-python` from the one that just relies on `xtensor`.

florianwechsung:
```
Number of derivatives: 0
         N Time (in ms) Gigainteractions/s
     10000       0.0107            0.93458
   1000000         0.99             1.0101
 100000000           98             1.0204
Number of derivatives: 1
         N Time (in ms) Gigainteractions/s
     10000       0.0305            0.32787
   1000000         2.78            0.35971
 100000000          280            0.35714
Number of derivatives: 2
         N Time (in ms) Gigainteractions/s
     10000       0.0949            0.10537
   1000000         8.65            0.11561
 100000000          852            0.11737
```

I think this is now ready to merged.

mbkumar:
@florianwechsung Thanks for working on this PR. I agree with your changes to build system that goes for march=native optimization for local builds. 
Regarding march flag for CI, I am also bringing Matt (@landreman) into this discussion.

The reason I choose march=x86-64 is to have a base optimization that works on all processors. However, this flag is very baseline in the sense it supports all the processors from early 2000's when AMD introduced 64 bit extensions for x86 architecture. Since we don't have to support such old processors, I thought we could do better, but was not able to figure out if choosing a specific platform such as march=ivybridge would also support the latest AMD processors.  The gcc docs are not clear on that. This point would have been moot and we could have used march=ivybridge if AMD didn't have competitive processors. But now AMD is leading and the new PPPL cluster is using AMD processors.  So it didn't make sense to choose any manufacturer specific flag. I also checked some other codes that supply pre-built binaries and the selected default is x86-64. 

After googling about this a little bit, I found [this phoronix page](https://www.phoronix.com/scan.php?page=news_item&px=GCC-11-x86-64-Feature-Levels). The author is a well known guy in the  hardware and software optimization space (owner of the website). According to the info on the page, newer versions of gcc (v11+) and clang (v12+) now support different versions of x86-64 optimization flags which work for a broad set of processors from both intel and amd. So we could specify march=x86-64 or x86-64-{v2, v3, v4}. x86-64-v4 features are supported only by intel processors as of now, but all AMD Zen+ processors support x86-64-v3. 

So my recommendation would be to use the gcc/clang compilers to v11+/v12+ in conjuction with march=x86-64-v3.

florianwechsung:
I believe `march=ivybridge` (which has AVX but not AVX2) lies somewhere between `x86-64-v2` (no AVX) and `x86-64-v3` (AVX, AVX2, and FMA). 

Having AVX, AVX2 and FMA makes the code _significantly_ faster, so I would be happy to assume `x86-64-v3` as default.

mbkumar (approved):

PR #44, opened 2021-04-12
Automatic linting

florianwechsung:
As discussed during the April 12 SIMSOPT meeting, I have created a little helper script that checks and automatically enforces pep8 rules.

For now I have only commited the script. Let's discuss which rules we want to enforce, and once we decided on a set of rules, I will run the script and commit the resulting code changes.

To try out the script and check the effect on the code

```
chmod u+x run_autopep
./run_autopep
git diff # to view the changes
git reset --hard HEAD # to undo them (warning: this will undo any changes to run_autopep as well)
```

For a list of rules, please check https://pycodestyle.pycqa.org/en/latest/intro.html#error-codes    
For now I have excluded a lot of the rules and only enforce those that are not very controversial (mostly related to whitespace).

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -26,37 +26,10 @@ jobs:
         test-type: [linting]
 
     steps:
-    # First print out lots of information. We do this in separate 
-    # "name" blocks because otherwise the output gets mixed together
-    # in the github actions log.
-    
-    - name: Print user and group id
-      run: |
-        set -ex
-        id
-
-    - name: PWD
-      run: |
-        set -ex
-        pwd
-
-    - name: ls -l
-      run: |
-        set -ex
-        ls -l
-
     - uses: actions/checkout@v2
       # If we want submodules downloaded, uncomment the next 2 lines:
       #with:
       #  submodules: true
-
-    # We must run actions/checkout@v2 before downloading and building VMEC, since checkout deletes the contents of the directory.
-    # For now we use the mattland fork of VMEC. Once changes are pushed to mbkumar's fork, we can use that instead.
-    - name: ls -l again
-      run: |
-        set -ex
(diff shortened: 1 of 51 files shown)


mbkumar:
Hi Florian,

Thanks for the script. Let's start the discussion on the excluded error codes.

My personal choice is **not to** keep E1, E3, and W in the exclude list. I am OK with the rest.

Why do I want to use E1, E3, and W?
1. I actually care about many of the errors that fall under E1. For example, E101 means the probability of subtle bugs in the code due to mixing tabs and spaces. If you have in mind a specific error code that falls under E1, let's use that instead of E1, which is too broad for general exclusion.
2. Python uses a sensible # of blank lines as function and class delimiters. Those lines provide clear visible clues for quick glancing of code. A linter fixing E3 doesn't change the code on any line, just deletes or add blank lines. So having E3 in exclude list doesn't make sense to me.
3. W are not errors, so let's keep them. I also care about W6 (deprecation warnings).

mbkumar:
Since we'd like to do lint checking for PR's, could you please add line 10 and 11 to CI as well?

florianwechsung:
I'm happy to include E1 and E3. Regarding the warnings: flake8 will return 1 if warnings are triggered, which would fail the CI, that's why I excluded those. I did a bit of googling, but couldn't figure out how to get flake8 to print the warnings but return 0 still.

mbkumar:
OK. In that case, let's keep W in the exclude list and remove only E1 and E3.

florianwechsung:
Sounds good!

Matt @landreman, any strong feelings from your side about this?

landreman:
No strong feelings from my side.

mbkumar (changes requested):

mbkumar (approved):

PR #45, opened 2021-04-12
least_squares_problem.py moved to optimizable subpackage

mbkumar:
The module least_squares_problem.py, which defines LeastSquaresProblem is moved to solve sub-package from core sub-package. Since the  implementation of the solvers is tightly integrated with the methods defined in LeastSquaresProblem, the new location is a natural fit.


landreman:
After mulling this over, I think LeastSquaresProblem makes more sense where it is in core rather than in solve. Here is my reasoning. What exactly is core for? Core is about *defining* optimization problems. And indeed the LeastSquaresProblem class is for defining a certain class of optimization problems. Once a user defines a problem, they could choose to solve it with one of the simsopt/scipy solvers but they don't have to. They could solve it with an outside solver package, or they might want to just plot the function without solving the problem. In these cases they would still need LeastSquaresProblem, and it seems sensible that they should not expect to use use the simsopt.solve module. What do you think?

mbkumar:
I have a different picture in mind.  Core folder is defining the modules
for fundamental classes of simsopt. However they are not routinely used by
end users.  For example the Dofs and Optimizable classes are not needed by
end users.  Those classes are mostly used by developers to implement
Optimizable subclasses. So in future my plan is to move core to _core to
signify end users don't have to know much about it. Of course power users
could access it. This will follow standard python paradigm of hiding
fundamental stuff by prefixing them with underscore, but making them always
accessible.

Regarding least squares,  we could change the folder name to least squares
and keep the three modules as it is.

Alternatively,  we could introduce a new folder named "optimizables" or
"objectives" and move LeastSquaresProblem class implementation there. In
that folder,  we could define various cost function classes in addition
least squares that are used by end users.

What do you think?

landreman:
I see, it is a good point that core/dofs.py is not user-facing, and so it should perhaps be hidden away. The same is true of core/util.py. While core/functions.py is mostly used for unit tests and so could perhaps be moved to /tests, on the other hand maybe core.functions.Rosenbrock should be in some user-facing location as a simple test problem. core.optimizable.Optimizable or core.optimizable.optimizable() is needed by any user who wants to optimize some outside code or define a custom optimizable object, so those should be user-facing.

You mentioned the idea of a "least_squares" folder. At some point we may want to further develop machinery for optimization problems that are not of least-squares form. E.g. minimize the loss of alpha particles, or minimize the reactor volume, without squaring anything. Already solve.serial_solve has the serial_solve() function for such general problems. So the question is then should we have even more modules, for the least-squares and non-least-squares versions of problems and solvers. That seems like too many modules to me.

I would actually be quite happy if LeastSquaresProblem, Optimizable, optimizable(), and the solvers were imported into the root simsopt module, to make imports less verbose. I'm guessing you prefer otherwise.

Or, I kind of like the last approach you suggested - here's a version of it:

simsopt._core module: contains stuff to hide from users. Contains dofs.py and util.py.

simsopt.problems or simsopt.objectives module: contains user-facing tools for defining optimization problems. Contains optimizable.py, least_squares_problem.py (possibly renamed least_squares.py for brevity?), and functions.py (some of which could be moved to /tests)

simsopt.solve module: contains serial_solve.py and mpi_solve.py. These could potentially be renamed serial.py and mpi.py for brevity, since they are already within simsopt.solve?

mbkumar:
Sounds good.  I am OK with importing important classes to top level,
instead of "from xyz import *" blanket imports.

mbkumar:
On Fri, Apr 16, 2021 at 5:39 PM Matt Landreman ***@***.***>
wrote:

> I see, it is a good point that core/dofs.py is not user-facing, and so it
> should perhaps be hidden away. The same is true of core/util.py. While
> core/functions.py is mostly used for unit tests and so could perhaps be
> moved to /tests, on the other hand maybe core.functions.Rosenbrock should
> be in some user-facing location as a simple test problem.
> core.optimizable.Optimizable or core.optimizable.optimizable() is needed by
> any user who wants to optimize some outside code or define a custom
> optimizable object, so those should be user-facing.
>
Yes, there is one function/class we need to expose from this folder. We
could import into top level.

> You mentioned the idea of a "least_squares" folder. At some point we may
> want to further develop machinery for optimization problems that are not of
> least-squares form. E.g. minimize the loss of alpha particles, or minimize
> the reactor volume, without squaring anything. Already solve.serial_solve
> has the serial_solve() function for such general problems. So the question
> is then should we have even more modules, for the least-squares and
> non-least-squares versions of problems and solvers. That seems like too
> many modules to me.
>
> I would actually be quite happy if LeastSquaresProblem, Optimizable,
> optimizable(), and the solvers were imported into the root simsopt module,
> to make imports less verbose. I'm guessing you prefer otherwise.
>
I am fine with this.

mbkumar:
@landreman
>  core.optimizable.Optimizable or core.optimizable.optimizable() is needed by any user who wants to optimize some outside code or define a custom optimizable object, so those should be user-facing.

On the above point, my expectation is that the people using Optimizable class are simsopt developers because it is used mainly to write a subclass to develop an interface. So, we should expose only core.optimizable.optimizable() function for end-users to convert their functions or classes to Optimizable object on the fly. 

I took another look at the _**optimizable**_ function. One drawback I found is that you are modifying the input object. So what if a user doesn't want their objects to be modified? We should have an option such as `input=True`, so modifications happen in-place. Otherwise, a deep-copy is used to generate a new object.  The function name is also  not very informative. How about we change it to _**make_optimizable**_?

I'd also like to keep optimizable.py in _core subpackage. I'll do a top-level import for **_make_optimizable_** function to make it visible for end users.

landreman:
> On the above point, my expectation is that the people using Optimizable class are simsopt developers because it is used mainly to write a subclass to develop an interface. So, we should expose only core.optimizable.optimizable() function for end-users to convert their functions or classes to Optimizable object on the fly.
> 
> I took another look at the _**optimizable**_ function. One drawback I found is that you are modifying the input object. So what if a user doesn't want their objects to be modified? We should have an option such as `inplace=True`, so modifications happen in-place. Otherwise, a deep-copy is used to generate a new object. The function name is also not very informative. How about we change it to _**make_optimizable**_?
> 
> I'd also like to keep optimizable.py in _core subpackage. I'll do a top-level import for **_make_optimizable_** function to make it visible for end users.

This all sounds fine to me.

mbkumar:
The proposed changes have been implemented.
Changes made are:

1. least_squares_problem.py renamed to least_squares.py
2. core renamed to _core
3. least_squares.py and functions.py moved to newly created `objectives` subpackage.
4. imports are updated everywhere.
5. optimizable function renamed to `make_optimizable`.
6. Docs updated with the new API

 Only thing left to be implemented is the inplace option for make_optimizable function. But that could part of another PR. I think this is ready to be merged.

landreman (approved):

PR #48, opened 2021-04-14
Simplified logging interface introduced

mbkumar:
This PR introduces logging functionality that is simple and consistent to use for both serial and mpi jobs.

For serial jobs: 
```python 
from simstop.util.log import initialize_logging
initialize_logging(filename="serial.log")
for i in range(2):   # Some log dumping
    logging.info("Hello (times %i) from serial job" % (i+1))
```
serial.log:
```bash
2021-04-14 12:31:00,089 root-INFO: Hello (times 1) from serial job
2021-04-14 12:31:00,089 root-INFO: Hello (times 2) from serial job
```

For mpi jobs:
```python
from mpi4py import MPI
from simstop.util.log import initialize_logging

comm = MPI.COMM_WORLD
initialize_logging(mpi=True, filename="mpi.log")
for i in range(2):
    logging.warning("Hello (times %i) from mpi job)
```
Resulting mpi.log with 2 MPI processes:
```bash
[rank 0/2] 2021-04-14 12:32:35,898 root-WARNING: Hello (times 1) from mpi job
[rank 1/2] 2021-04-14 12:32:35,898 root-WARNING: Hello (times 1) from mpi job
[rank 0/2] 2021-04-14 12:32:35,911 root-WARNING: Hello (times 2) from mpi job
[rank 1/2] 2021-04-14 12:32:35,911 root-WARNING: Hello (times 2) from mpi job
```

The `initialize_logging` function imported from `simsopt.util.log` module uses dictconfig method to initialize logging. The default logging handler and formatting options are stored in `simsopt/util/log_config.yaml` file.

For optional MPI, in addition to _mpi4py_, _mbkumar/python-mpi-logger_ repo is also installed.

Credits: The MPI logging functionality uses MPILogger package from Richard Shaw, Dept. of Physics, Univ. of British Columbia. I modified the MPILogger to get it work with python3.x.

`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,10 +10,11 @@ build/
 build-cmake/
 dist/
 pip-wheel-metadata/
-src/SIMSOPT.egg-info/
+*simsopt.egg-info/
 .DS_Store
 .coverage
 .tox
+*fort.9
 	
 # f90wrap related
 src.*/
@@ -40,6 +41,10 @@ spec*.sp.h5
 
 # examples
 examples/VMEC/
+examples/*.log
 
 # Sphinx related files
 docs/build/
+
+# IDE
+.idea
--- a/examples/logger_example.py
+++ b/examples/logger_example.py
@@ -17,4 +17,4 @@ except:
 if comm is not None:
     initialize_logging(mpi=True, filename='mpi.log')
     for i in range(2):
-        logging.warning("Hello (times %i) from rank %i (of %i)" % (i+1, comm.rank, comm.size))
+        logging.warning("Hello (times %i) from mpi job" % (i+1))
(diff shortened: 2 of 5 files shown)


landreman:
Since this logger spawns one MPI process that does only logging, we'll have to see how well it interacts with the other MPI machinery used in simsopt when used on HPC systems. For instance we may want to oversubscribe, e.g. have 33 MPI processes on 32 cores, where the logging process isn't actually taking much CPU time. This will take some time to try out so I don't necessarily want to hold up this PR while we try it out. In the meantime let's not remove the existing simsopt.util.mpi.log() function, which does not add an MPI process and does not introduce these new dependencies.

To simplify installation I'd like to minimize dependencies as much as possible, and this PR introduces 3. Can we remove monty? Here it doesn't seem to be doing anything more than opening a file, right?

mbkumar:
Sure. I tested it only on my local machine without any real simulations. As Florian suggested, we could convert the print statements to logging statements present in a production type simulation and check.

 I removed the dependency on monty.

landreman (approved):

PR #53, opened 2021-04-21
Better handling of failures in the objective function

### LINKED ISSUE #51: Bug in mpi_solve.py
Hello Matt,

There is a bug in _mpi_solve.py_ at line 209. The `prob` variable was not defined anywhere in that scope. Could you please fix it and write a test case that covers that part of the code?

Thank you,
Bharat

landreman:
This pull request addresses @florianwechsung 's comment in slack last week that the solver should not catch every error that occurs in the objective function. This pull request also fixes #51.

Now, there is a custom exception `ObjectiveFailure` that is raised by Vmec and Spec when they fail to converge. Only this specific exception is now caught during optimization. The error catching is done in Dofs rather than `least_squares_serial_solve` and `least_squares_mpi_solve` to eliminate duplication. Moreover, an optional argument `fail` to the constructor of `Dofs` or `LeastSquaresProblem` allows the user to customize the large value that is used to represent failures, or to turn off this error catching by setting `fail=None`. Tests have been added in which intentional failures of the objective function occur.


mbkumar (approved):
Looks good

PR #55, opened 2021-05-01
New magnetic field class: Reiman model

rogeriojorge:
Magnetic field taken from section 5 of Reiman, Greenside (1986), "Calculation of three-dimensional MHD equilibria with islands and stochastic regions", Computer Physics Communications 43 (157-167)
This field has the advantage of allowing an exact calculation of the island width which can be used for island width optimization (example: https://arxiv.org/abs/2102.04497)

What still needs to be added:
- Continuous integration tests
- Optimization example


rogeriojorge:
Unit tests added. The new magnetic field class is now ready for review.

florianwechsung (approved):

florianwechsung:
Merged, thanks!

Issue #64, opened 2021-05-12
Delete DOF class


mbkumar:
The current DOF class is not serving any purpose. Remove it.

mbkumar:
Addressed in PR #67

Issue #66, opened 2021-05-12
Move MPILogger code to simsopt


mbkumar:
Because MPILogger is not hosted on pypi, pypi is rejecting simsopt uploads. MPILogger package is a single file with a permissive license. Add the file to simsopt with license on top of it.

mbkumar:
Addressed in PR 67

PR #67, opened 2021-05-12
Cleanup and refactoring

mbkumar:
1. This PR addresses #66 by merging mbkumar/python-mpi-logger into simsopt and removes the optional dependency so that simsopt could be uploaded to pypi. To test run examples/logger_example.py with mpirun and check mpi.log
2. Github actions wheel workflow is modified to upload the wheels to pypi instead of test.pypi.
3. dofs, state type properties that are synonymous with x type properties in Optimizable class definition are removed (Fixes issue #68).
4. DOF class that is not used anywhere removed (Fixes issue #64).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,7 +120,7 @@ jobs:
       run: |
         cd SPEC
         # In the SPEC Makefile, we need to set BUILD_ENV:
-        cat Makefile | sed 's/BUILD_ENV=intel/BUILD_ENV=gfortran_ubuntu/' > Makefile1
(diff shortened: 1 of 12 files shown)


landreman (commented):
Looks fine except that linting and CI are failing. I believe the CI is failing due to a change in SPEC yesterday, in which a question mark was added in the SPEC Makefile here:
https://github.com/PrincetonUniversity/SPEC/blob/master/Makefile#L62
This line of our CI needs to be modified accordingly
https://github.com/hiddenSymmetries/simsopt/blob/mpi_logger/.github/workflows/ci.yml#L123

mbkumar:
Hello Matt,

Addressed the failing tests.

landreman (approved):

PR #73, opened 2021-05-17
Magnetic well metric

landreman:
This is a measure of MHD stability that can be computed directly from VMEC and which is often included in STELLOPT/ROSE optimization.


src/simsopt/mhd/vmec.py
```diff
@@ -399,6 +399,10 @@ def load_wout(self):
         #self.wout.volume = f.variables['volume_p'][()]
         f.close()
 
+        self.s_full = np.linspace(0, 1, self.wout.ns)
```
mbkumar, line 402: For the sake of code readability, is it possible to expand 's_full' and 's_half' to something more descriptive?

landreman:
Done.

mbkumar (approved):
Thank you!

PR #75, opened 2021-05-24
MagneticField refactoring

florianwechsung:
This PR introduces the following changes

- The abstract `MagneticField` is now implemented in C++. The interface for the user is essentially the same.
- For certain calculations is handy to work in cylindrical coordinates. I have introduced `set_points_cyl` and `set_points_cart` so that the user can use either cylindrical or cartesian coordinates to set the points where the field should be evaluated. Also added a `B_cyl()` function to return the field in cartesian coordinates.
- `BiotSavart` is now implemented in C++
- `InterpolatedMagneticField` takes a field, interpolates it on a regular grid and then allows for fast evaluations of this field. This is useful for doing things like poincare plots or particle tracing, when field evaluation is the bottleneck.
- I also used this chance to do some general refactoring to split the C++ code into more separate compilation units, to reduce compile time whenever a change on the C++ side is made.


mbkumar:
Hello Florian,

Since this is a big set of changes, shall we do a 1 hour session where you can explain the changes to me? That way I could process the code much faster.

landreman:
As part of this refactoring, what do you think of the idea of renaming `simsgeopp` everywhere to `simsoptpp`? A lot of "simsgeopp" is related to magnetic fields rather than geometric objects, and soon "simsgeopp" will have other stuff like guiding center integration.

src/simsopt/geo/biotsavart.py
```diff
@@ -1,27 +1,55 @@
...
+    where :math:`\mu_0=4\pi 10^{-7}` is the magnetic constant.
+
+    """
 
     def __init__(self, coils, coil_currents):
         assert len(coils) == len(coil_currents)
+        self.currents_optim = [sopp.Current(c) for c in coil_currents]
+        self.coils_optim = [sopp.Coil(curv, curr) for curv, curr in zip(coils, self.currents_optim)]
+        MagneticField.__init__(self)
+        sopp.BiotSavart.__init__(self, self.coils_optim)
```
mbkumar, line 24: coils_optim is set in a derived class and then passed as an argument to base class. Kind of unusual. What is the logic behind this?

src/simsopt/geo/biotsavart.py
```diff
@@ -54,58 +82,66 @@ def compute_A(self, points, compute_derivatives=0):
...
+            self._ddA = sum(self.coil_currents[i] * self._d3A_by_dXdXdcoilcurrents[i] for i in range(len(self.coil_currents)))
         return self
 
-    def compute(self, points, compute_derivatives=0):
-        assert compute_derivatives <= 2
-
-        self._dB_by_dcoilcurrents = [np.zeros((len(points), 3)) for coil in self.coils]
+    def _A_impl(self, A):
+        self.compute_A(compute_derivatives=0)
+        A[:] = self._A
```
mbkumar, line 92: Passing a variable and filing it is more of C/C++ style. I am guessing we are doing it this way for efficiency's sake. Is there a python approach that is equally efficient? Also, slicing can create a copy. Please make sure there are no gotcha's here.

src/simsopt/geo/magneticfield.py
```diff
@@ -1,139 +1,133 @@
...
+        RPhiZ[:, 1] = Phi.flatten()
+        RPhiZ[:, 2] = Z.flatten()
+
+        self.set_points_cyl(RPhiZ)
+        vals = self.B()
+        contig = np.ascontiguousarray
+        gridToVTK(filename, X, Y, Z, pointData={"B": (contig(vals[:, 0]), contig(vals[:, 1]), contig(vals[:, 2]))})
+
 
 class MagneticFieldMultiply(MagneticField):
```
mbkumar, line 67: Do we need a class to do this? Can't we use use numpy multiply operation? Another approach could be to write a class method.

src/simsopt/geo/magneticfield.py
```diff
@@ -1,139 +1,133 @@
...
+        A[:] = self.scalar*self.Bfield.A()
+
+    def _dA_by_dX_impl(self, dA):
+        dA[:] = self.scalar*self.Bfield.dA_by_dX()
+
+    def _d2A_by_dXdX_impl(self, ddA):
+        ddA[:] = self.scalar*self.Bfield.d2A_by_dXdX()
 
 
 class MagneticFieldSum(MagneticField):
```
mbkumar, line 101: Same question for this class as well?

mbkumar (commented):
Some minor comments mainly to understand the logic.

src/simsopt/geo/biotsavart.py
florianwechsung: Yeah, we're doing this so that the memory management can be done manually on the C++ side and so that we don't have to malloc a new array every time.

According to https://stackoverflow.com/questions/47181092/numpy-views-vs-copy-by-slicing, slicing only creates a copy when non-contiguous data is accessed, so I think we're good here.

src/simsopt/geo/biotsavart.py
florianwechsung: A coil object essentially just consists of a curve together with a current object. Once the new graph structure is fully ready, that's how we want user to do coil optimisation. It will naturally allow for scenarios where several coils have the same current for example. The constructor to Biot Savart will then just take a list of coils.

However, since that's not ready yet, I don't want to break the old interface, and for now the user facing biot savart class takes separate curves and currents, and then internally creates the corresponding coil objects.

src/simsopt/geo/magneticfield.py
florianwechsung: Discussed on slack. For future reference: the point is to support lazy evaluation of expressions such as `B = B0 + alpha * B1`

mbkumar (approved):

PR #78, opened 2021-05-30
Coverage with tox

mbkumar:
This PR achieves coverage using tox to run pytest and generate coverage reports. Coverage is aggregated over python versions 3.7, 3.8, and 3.9. The coverage report is uploaded to codecov.io and displayed as a badge.
Along the way fixed few issues with top-level imports of MPI stuff (don't do it) and skipping tests if SPEC is not installed.

Things to do for the future:
1. The pytests are essentially duplicating the running of the unit tests in ci workflow. One thing we could do is to get rid of running unit tests in the ci workflow. 
2. Since the installation of VMEC and SPEC takes a long time, we could try to combine both workflows into one. 

Both the alternatives are worth pursuing. But for the time being, the PR is ready to merge.


landreman:
Awesome. Do I understand correctly that the 73% coverage figure is artificially low because vmec, spec, and booz_xform are not installed in the tox/coverage CI, and so all the vmec/spec/boozer tests are skipped there?

mbkumar:
Yes. Once I merge both the workflows, those tests will also be executed.

florianwechsung:
Thank you for working on this, this is great and already show's us a few places where we can do better.

@rogeriojorge if you look at this file, you can see that some of the MagneticField classes are not fully covered. I think the one using sympy can be fixed by adding sympy to the github action script. For the circular coil there doesn't seem to be a test for the potential, so it would be good to add that.

landreman (dismissed):

florianwechsung (dismissed):

PR #79, opened 2021-06-01
Guiding Center Trajectories and Fieldline tracing

florianwechsung:
This introduces the ability to compute fieldlines as well as particle trajectories using the guiding center approximation. The computation is done on the C++ side for better performance.

Some points of discussion:
- How should this be tested?
- I introduced boost::odeint as dependency. That library is header only and boost is probably available everywhere, but maybe worth discussing whether we should turn this into an optional part of simsopt.


florianwechsung:
I have moved everything Magnetic field related into `simsopt.field`. In terms of tracing, I have added right hand sides for field lines, full orbit, and guiding center. The guiding center rhs assumes `\nabla p = 0` for now (an error is thrown when the general form is requested). We can add the more complex guiding center rhs once we can work with bfields from desc/spec/vmec.

For all three types of tracing, one can pass a list of `phis` at which the state should be recorded, as well as a list of `stopping_criteria` (intended for finding out when the particle is outside of a surface). There is an example in `1_simple/tracing.py`. 

The computation can be parallelized across an MPI communicator if desired.

rogeriojorge:
The features introduced by this PR are great! Thanks!
Although I am still familiarizing myself with all the new functions, I think that I managed to add a test for energy conservation in the tests/field/test_particle.py. The conservation of energy in the full orbit case seems to be fine, with a maximum error in the 8th decimal place. I am having a bit more trouble to it in the guiding center case. @florianwechsung  could you check if there are some normalizations in the charge/velocity that should be taken into account in the test? I suspect that's the reason that some terms might not be canceling.

florianwechsung:
Thank you for adding the tests @rogeriojorge  I got them working with some very minor changes, could you have a look and double check that they make sense? I now get
```
Max Energy Error for full orbit     =  [-8.25014115 -8.36487096 -8.87836965 -8.36625564]                                                 
Max Energy Error for guiding center =  [-10.66350997 -10.37032219 -10.68677456 -10.77267388]                                             Max mu Error for full orbit         =  [-7.94533722 -7.99730043 -5.27184613 -8.08222078]                                                 
Max mu Error for guiding center     =  [-10.36245206 -10.06927661 -10.38571532 -10.47160711]          
```

landreman:
Lots of awesome new features here! Generally I like the division into simsopt.geo vs simsopt.field.

Comments so far:

More unit tests needed generally: e.g. for gc_to_fullorbit_initial_guesses, signed_distance_from_surface, SurfaceClassifier, LevelsetStoppingCriterion, IterationStoppingCriterion

It would be good if the extensive CI also covers the case in which boost is not present.

docstrings needed for compute_fieldlines, SurfaceClassifier, LevelsetStoppingCriterion, get_ncsx_data. The latter should explain whether or not the planar coils are included.

How about including a function to save Poincare/particle tracing results, and another function to load it and plot it?

Testing idea: verify that the field line tracing routine agrees with the magnetic axis from a known magnetic field. Could use the coil config in Cary & Hanson (1986). Rogerio has a simsopt script for these coils.

tracing.py line 24, do you mean velocity (a vector) or speed (a scalar)? In the latter case, how does gc_to_fullorbit_initial_guesses choose the gyro phase? This should be explained in the docstring.

trace_particles and trace_particles_starting_on_axis: 9000 eV is a weird default. How about 3.5 MeV.

The docstring for trace_particles should give explicit expressions for the guiding center equations used, since there are several variations. (E.g. whether or not a parallel drift is included, whether or not the B||* factor is included)

Does signed_distance_from_surface handle both the case in which Gammadash1 cross Gammadash2 points outward and the case in which it points inward? Both cases are possible, right?

It looks like signed_distance_from_surface is using surface.normal(). As I understand it, normal() is not the _unit_ normal. Should signed_distance_from_surface be using the unit normal? It would be good if the base Surface class could return the unit normal. Users might be confused and think that normal() is a unit normal, so perhaps there is a clearer name for normal().

In my experience, plasma folks usually talk about the "parallel velocity" rather than "tangential velocity". So consider renaming "tangential_velocities" -> vpar or v_parallel?

Also, here are some comments more about MagneticField and interpolation code that's already in master, but with some relevance to this PR:

InterpolatedField needs more documentation. How are the interpolation nodes specified? What interpolation methods are used exactly? Documentation is needed both in the python and C++ source.

It would be good to have unit tests for the interpolation machinery, besides just field.test_magneticfieldstest_interpolated_field_convergence_rate / close.

Need user guide documentation for MagneticField and BiotSavart, similar to what is present for surfaces/curves.

In this, the documentation for set_points_cyl and B_cyl should say what order the coordinates are. I presume the order is (R, phi, Z), but it would be good to display this clearly.

Shorten name of magneticfieldclasses.py ? Maybe fields.py? 

Does the interpolation machinery know that the phi coordinate is different from R and Z in that things are periodic in phi?

It's generally confusing that in some parts of simsopt (e.g. tracing.cpp line 155), `phi` is a coordinate which goes up to 2pi, whereas in the Surface parts of simsopt, `phi` means an angle that goes up to 1. VMEC uses angles that go up to 2pi, and it is often convenient to allow the cylindrical coordinate phi and surface parameter phi to be the same.

florianwechsung:
Thank you for the detailed comments. I took care of most of them. A few more detailed responses/questions:

> How about including a function to save Poincare/particle tracing results, and another function to load it and plot it?

I added a routine to create the typical poincare plot in matplotlib. What did you have in mind for saving/loading? The results are just numpy arrays, so saving/loading anything is as easy as `np.save(data, filename)`, and I don't think we need a dedicated function for that.

> Testing idea: verify that the field line tracing routine agrees with the magnetic axis from a known magnetic field. Could use the coil config in Cary & Hanson (1986). Rogerio has a simsopt script for these coils.

Added a test for the NCSX configuration for which we have the magnetic axis.

> trace_particles and trace_particles_starting_on_axis: 9000 eV is a weird default. How about 3.5 MeV.

I added a `simsopt.util.constants` file that includes charges, mass, and energy for some typical particles. I use an alpha particle as default now.

> Does signed_distance_from_surface handle both the case in which Gammadash1 cross Gammadash2 points outward and the case in which it points inward? Both cases are possible, right?

Yes it works in either case. covered in tests now.

> It looks like signed_distance_from_surface is using surface.normal(). As I understand it, normal() is not the unit normal. Should signed_distance_from_surface be using the unit normal? It would be good if the base Surface class could return the unit normal. Users might be confused and think that normal() is a unit normal, so perhaps there is a clearer name for normal().

Added a unitnormal function.

florianwechsung:
Regarding the non PR specific comments:

> InterpolatedField needs more documentation. How are the interpolation nodes specified? What interpolation methods are used exactly? Documentation is needed both in the python and C++ source.

Opened an issue for this #103 

> It would be good to have unit tests for the interpolation machinery, besides just field.test_magneticfieldstest_interpolated_field_convergence_rate / close.

see #103

Need user guide documentation for MagneticField and BiotSavart, similar to what is present for surfaces/curves.

opened #104

> In this, the documentation for set_points_cyl and B_cyl should say what order the coordinates are. I presume the order is (R, phi, Z), but it would be good to display this clearly.

fixed

> Shorten name of magneticfieldclasses.py ? Maybe fields.py?

I agree that a shorter name would be good. We could also separate them into `reiman.py`, `toroidal.py` and `dommaschk.py`. That would be more consistent with how we keep surfaces and curves separate.

> Does the interpolation machinery know that the phi coordinate is different from R and Z in that things are periodic in phi?

It doesn't, since it's a general interpolant on a box in 3d. The only benefit of adding the periodic-in-phi special case would be that we could save ourselves function evaluation on one side of a box domain, which is fairly negligible.  For other interpolants (e.g. ones using a fourier expansion in phi), this is of course different.

> It's generally confusing that in some parts of simsopt (e.g. tracing.cpp line 155), phi is a coordinate which goes up to 2pi, whereas in the Surface parts of simsopt, phi means an angle that goes up to 1. VMEC uses angles that go up to 2pi, and it is often convenient to allow the cylindrical coordinate phi and surface parameter phi to be the same.

I agree that it's a bit confusing. The reasoning for this is twofold:

- in general this is really just a parameter, not a physical angle. take CurveXYZFourier for example, there is no physical angle that \phi corresponds to
- it simplifies integrals: if phi goes from 0 to 2pi, every integral needs to be normalised by 2pi. Where as now, we just write `curve_length = np.mean(np.linalg.norm(curve.gammadash(), axis=1))` etc, no normalisation required.

Going through all curve and surface classes is not something I'm particularly keen on, so I'd only want to change this if people have very strong feelings about it (/ or if someone else volunteers to do it).

landreman:
Thanks for all these improvements. A few more comments:

- Can the interpolated field take advantage of nfp-symmetry and stellarator-symmetry, so phi only needs to cover half a field period rather than all of 2pi?

- Rather than doubling the number of extensive-CI jobs by having "boost: [true, false]" in the matrix, consider adding just 1 or a few runs as in https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-additional-values-into-combinations

- Since the interpolant does not know about periodicity in phi, is there any loss in accuracy for high-order interpolation near the "seam" at phi=0 and 2pi, since the interpolant does not take advantage of data on grid points "after" phi=2pi or "before" phi=0?

- Consider renaming SurfaceClassifier -> SignedDistance, since the latter name is more general and it could be useful for other applications? Also consider moving it from simsopt.field to simsopt.geo since the signed distance function is a geometric concept and not really related to magnetic fields.

florianwechsung:
Thanks @landreman

> Can the interpolated field take advantage of nfp-symmetry and stellarator-symmetry, so phi only needs to cover half a field period rather than all of 2pi?

Added that capability.

> Rather than doubling the number of extensive-CI jobs by having "boost: [true, false]" in the matrix, consider adding just 1 or a few runs as in https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-additional-values-into-combinations

Only added two extra jobs not involving boost now, all other jobs now also have boost.

> Since the interpolant does not know about periodicity in phi, is there any loss in accuracy for high-order interpolation near the "seam" at phi=0 and 2pi, since the interpolant does not take advantage of data on grid points "after" phi=2pi or "before" phi=0?

No, we have same order everywhere. Periodicity doesn't affect this type of interpolant, we always just rely on the values in the current 'cell', no matter if that cell is on the boundary of the r,phi,z cube or not.

> Consider renaming SurfaceClassifier -> SignedDistance, since the latter name is more general and it could be useful for other applications? Also consider moving it from simsopt.field to simsopt.geo since the signed distance function is a geometric concept and not really related to magnetic fields.

The `SurfaceClassifier` is actually specific to computing the distance to a surface, so I don't think it should be renamed. The general concept here is the `LevelsetStoppingCriterium`, and one could create other types of classifiers which can then be passed to that criterium. I have moved the `SurfaceClassifier` into simsopt.geo.

landreman (approved):
Looks great!

florianwechsung:
Thanks @landreman and @rogeriojorge for your input. I merged rogerios branch into this branch, so github requires a fresh review to  merge this one.

landreman (approved):

PR #89, opened 2021-06-06
Cleanup of the examples folder

rogeriojorge:
This pull request organizes the examples folder into subfolders in order to separate input files from main optimization scripts. Additionally, the scripts should have a text at the beginning showing exactly what the script is doing and how to run it. This text should have a uniform style across different files.

An extension .py has been added to all the python files so that run_auto pep could apply proper linting.

Added a README.me to make it easier to navigate the examples folder.

Questions:
- Should complex optimization scripts (such as VMEC + SPEC optimization) be separated from simpler ones (such as the minimization of the curve length) in order not to confuse a newcomer?
- How to make the examples folder easy to navigate for somebody that is trying to run SIMSOPT the first time?


florianwechsung (dismissed):

mbkumar:
Just a heads up, that this PR will clash the PR #96. After one of them is merged, a manual resolution of conflict is needed to merge the other.

landreman:
Thanks for working on this!

If I understand correctly, all the examples that use an input file can only be run from the `examples` directory and not from the directory in which the driver script lives (e.g. `examples/stellopt_scenarios`) because of the way the paths to the input files are hard-coded. This requirement could be confusing to users. To enable the script to be run both from `examples` and from the subdirectories (or anywhere else), how about setting the input file names via something like

    filename = os.path.join(os.path.dirname(__file__), 'input', 'input.1DOF_Garabedian')

Also, some of the benefits of the 1_ 2_ 3_ grouping are lost when there is a separate `stellopt_scenarios` directory: where does `stellopt_scenarios` fall in that numbering? So I'd suggest folding those examples in with the other directories. In my opinion, "simple" and "advanced" are a bit subjective (e.g. does "simplicity" factor in execution time or not?) Users might need to search multiple directories to find an example relevant to their interest. For these reasons I personally like an objective grouping like `1_no_equilibrium_codes`, `2_vmec`, `3_spec`, `4_vmec_and_spec`.

landreman:
Also the 3 examples in the readthedocs docs refer to file locations in the examples folder, so let's update these references in docs/source/example_*

mbkumar:
@rogeriojorge 

In place of using os.path, please use Path. Please take a look at the tests that use input files and how those files are referred to.

landreman (dismissed):
Looks good to me. Since @mbkumar asked for a modification I'll wait to merge it.

mbkumar:
@landreman,
Do we still need the QAS.py in the intermediate examples folder? It is using the oldest style of VMEC import that we abandoned a long time ago? It won't work now and we are not running it anyhow. If we want to run the example, we need to rewrite it.

landreman:
@mbkumar QAS.py can be removed.

PR #93, opened 2021-06-09
Update coil picture in readme

florianwechsung:


florianwechsung:
The reason for this is that the old picture looks a bit odd for people using the dark github background:

image

landreman (approved):

PR #96, opened 2021-06-12
Imports fixed to account for optional packages

mbkumar:
This PR fixes the failing tests when optional packages are not installed. The imports in the source code are adjusted to prevent simsopt becoming inoperable whenever it is not able to import optional packages. And imports in tests are adjusted to prevent  spurious failures of tests. The CI is also updated with a newer strategy matrix containing vmec and spec installation as additional options. run_examples file is split into multiple parts and only specific parts are run depending on which optional packages are installed.


mbkumar:
Based on the feedback, I created a simplified CI that runs on branches and an extensive CI that runs on the master branch and releases. The simplified CI uses a single python version and installs all dependencies. Extensive CI is the full-blown combinatorial CI with three python versions and with optional installs of VMEC and SPEC.

mbkumar:
@florianwechsung 
This is ready to be reviewed.

landreman:
The new split CI seems to be a good solution to the issue we discussed on Monday. It would be nice if a user could have the ability to run the extensive CI once in a while when they choose to, to make sure everything works with/without vmec and with/without spec before they are ready to make a PR. From https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow it sounds like this can be accomplished by adding `workflow_dispatch` to the list of trigger events. Would this addition be ok with you?

mbkumar:
Sure Matt. That's a nice alternative.

Bharat Medasani

Engineer

mbkumar:
I added the workflow_dispatch event for the extensive_ci.yml. However, it needs to be merged into the master branch before one could run it manually. When running manually, one can pick the branch.

landreman (approved):

mbkumar:
An example of manually run CI can be found [here](https://github.com/hiddenSymmetries/simsopt/actions/runs/940869767)

PR #98, opened 2021-06-19
New surface representation by Henneberg et al

landreman:
This PR has an implementation of a new representation for toroidal surfaces proposed by Henneberg et al in https://arxiv.org/pdf/2105.00768.pdf. This representation has about half as many degrees of freedom as in SurfaceRZFourier and is expected to be useful for stellarator shapes.


florianwechsung:
- I think right now if one wants to do anything with this surface, one always has to convert to a `SurfaceRZFourier` first. Can you add the explicit formula for the surface, that way we don't always have to do this transformation (and we would also be able to get derivatives fairly easily)? 
- `SurfaceRZFourier` computes area and volume independently, so I think there is no need for the `area_volume` function that computes both at the same time.

landreman:
Thanks @florianwechsung - I've addressed those two items.

landreman:
@florianwechsung Do I understand correctly that you want SurfaceHenneberg to implement concrete functions for `gamma_impl`, `gammadash1_impl`, `gammadash2_impl`, `dgamma_by_dcoeff_impl`, `dgammadash1_by_dcoeff_impl`, and `dgammadash2_by_dcoeff_impl`? In some of the other surface classes I also see a function `gamma_lin` - what does this do?

landreman:
`gamma_lin` now works, and all tests pass in my fork. OK if we merge this now, and implement `dgamma*_by_dcoeffs_impl` later if needed?

Generally, requiring developers to implement both `gamma_impl` and `gamma_lin` would seem to promote code duplication. Wouldn't it make sense to let developers implement only `gamma_lin`, and have the base `Surface` class include a `gamma_impl` function that just calls `gamma_lin` for the tensor product grid in (theta,phi)? If users want to override that default `gamma_impl` for better performance they could.

florianwechsung:
Yeah that's a good suggestion. `gamma_lin` was an afterthought when Andrew added the flux and crosssection, so we didn't think too much about the structure there. Do you want to make that change in this PR or should we keep this in mind for a separate one?

florianwechsung:
One more comment:

The other surface classes support that either a number of quadrature points in phi and theta is passed, or that a specified list of phis and thetas is passed. This is useful
a) for the taylor tests for gammadash1 and gammadash2 (we want to evaluate gamma at 0, eps, eps/2, eps/4, eps/8 for a taylortest)
b) for surfaces with stellsym or nfp>1, sometimes we only want to work on a part of the surface, and sometimes (for visualisation mainly) we want the full surface.

landreman:
> Yeah that's a good suggestion. `gamma_lin` was an afterthought when Andrew added the flux and crosssection, so we didn't think too much about the structure there. Do you want to make that change in this PR or should we keep this in mind for a separate one?

Let's do that refactoring in a different PR.

> One more comment:
> 
> The other surface classes support that either a number of quadrature points in phi and theta is passed, or that a specified list of phis and thetas is passed. This is useful
> a) for the taylor tests for gammadash1 and gammadash2 (we want to evaluate gamma at 0, eps, eps/2, eps/4, eps/8 for a taylortest)
> b) for surfaces with stellsym or nfp>1, sometimes we only want to work on a part of the surface, and sometimes (for visualisation mainly) we want the full surface.

Support for initializing the points by either an int or array is added in the latest commit.

florianwechsung:
Thanks!
Is it possible to get the coverage report for this? @mbkumar  
Would be good to see if there are any major bits of the code left untested before merging.

landreman:
> Thanks!
> Is it possible to get the coverage report for this? @mbkumar
> Would be good to see if there are any major bits of the code left untested before merging.

There may be a better way to check this, but in the latest "simple CI" run on my fork https://github.com/landreman/simsopt/runs/2951185572?check_suite_focus=true, if you expand "Combine coverage reports" in the main panel, there is a coverage summary for each file. Coverage for the new file surfacehenneberg.py is 97%.

florianwechsung (approved):

florianwechsung:
Happy for this to be merged.

PR #106, opened 2021-07-07
QFM Surface v2

florianwechsung:
Due to changes in master, the tests failed after merging the previous QFM pull request. So we undid the PR and are now opening this one to fix the tests. Lesson learned from this: we should always merge master into a PR branch and check that the tests are passing before we merge a PR.

Elizabeth and I also noticed some strange behavior of the QFM surface computed in the example (it is self intersecting), so maybe let's wait until we know a bit better what's going on there before merging.


src/simsopt/geo/qfmsurface.py
```diff
@@ -0,0 +1,173 @@
+from scipy.optimize import minimize, least_squares, NonlinearConstraint
+import numpy as np
+from simsopt.geo.surfaceobjectives import QfmResidual
+
+
+class QfmSurface():
+
```
mbkumar, line 7: Please add docstrings here explaining the class.

src/simsopt/geo/qfmsurface.py
```diff
@@ -0,0 +1,173 @@
...
+        if derivatives == 1:
+            return val, dval
+
+        d2r = qfm.d2J_by_dsurfacecoefficientsdsurfacecoefficients()
+        d2l = self.label.d2J_by_dsurfacecoefficientsdsurfacecoefficients()
+        d2val = d2r  \
+            + constraint_weight*rl*d2l + constraint_weight*dl[:, None]*dl[None, :]
+
+        return val, dval, d2val
+
```
mbkumar, line 111: Can we have another function `minimize_qfm_penalty_constraints` with LBFGS and SLSQP supplied as choices to an argument called method

```
def minimize_qfm_penalty_constraints(self, tol=1e-3, maxiter=1000,
                                               constraint_weight=1, method=<default of LBFGS/SLSQP>)
```

mbkumar (commented):
Hello Florian, 

I have two minor comments.

Bharat

florianwechsung:
Not my code, so I'm tagging @ejpaul here.

src/simsopt/geo/qfmsurface.py
ejpaul: I added the docstring. 

I am not sure what you want to achieve with minimize_qfm_penalty_constraints. Right now, minimize_qfm_exact_constraints_SLSQP solves a constrained problem while minimize_qfm_penalty_constraints_LBFGS solves an unconstrained penalized problem. For example, constraint_weight is not used for minimize_qfm_exact_constraints_SLSQP.

mbkumar:
OK. I was thinking of scipy.optimize.minimize function where the method is
supplied as an argument. Both SLSQP and LBFGS are supplied via method
argument. However if you think its not worth, that comment can be ignored.

Bharat Medasani

Engineer

mbkumar (approved):

PR #107, opened 2021-07-08
Poloidal Field Magnetic Field class

rogeriojorge:
This PR introduces a new magnetic field class called PoloidalField. This allows more testing in the Poincare and Guiding Center solvers, namely to check that the Poincare plot yields a circular flux surface and that the canonical angular momentum is conserved in an axisymmetric geometry

- Help with the definition of canonical angular momentum. Are the units of both terms the same?


florianwechsung:
How about 
```
points = self.get_points_cyl_ref()
r = points[:, 0]
phi = points[:, 1]
z = points[:, 2]
```
instead of 
```
points = self.get_points_cart_ref()

x = points[:, 0]
y = points[:, 1]
z = points[:, 2]

phi = np.arctan2(y, x)
theta = np.arctan2(z, np.sqrt(x**2+y**2)-self.R0)
r = np.sqrt((np.sqrt(x**2+y**2)-self.R0)**2+z**2)
```

rogeriojorge:
The (r, phi, theta) coordinates are simple toroidal coordinates, which differ from the standard cylindrical coordinates in get_points_cyl_ref(). Therefore, I can't use that function directly in this case.

florianwechsung:
Ah of course, I should read more carefully, make sense!

src/simsopt/field/tracing.py
```diff
@@ -433,7 +433,7 @@ def plot_poincare_data(fieldlines_phi_hits, phis, filename, mark_lost=False, asp
             if data_this_phi.size == 0:
                 continue
             r = np.sqrt(data_this_phi[:, 2]**2+data_this_phi[:, 3]**2)
-            axs[row, col].scatter(r, data_this_phi[:, 4], marker='o', s=0.2, linewidths=0, c=color)
+            axs[row, col].scatter(r, data_this_phi[:, 4], marker='o', s=5, linewidths=0, c=color)
```
florianwechsung, line 436: If you disagree with the default I set here, please make it an optional argument to pass to the function

src/simsopt/field/tracing.py
```diff
@@ -398,7 +398,7 @@ class IterationStoppingCriterion(sopp.IterationStoppingCriterion):
     pass
 
 
-def plot_poincare_data(fieldlines_phi_hits, phis, filename, mark_lost=False, aspect='auto', dpi=300):
+def plot_poincare_data(fieldlines_phi_hits, phis, filename, mark_lost=False, aspect='equal', dpi=300):
```
florianwechsung, line 401: why the change?

src/simsopt/field/tracing.py
rogeriojorge: I tried with the default option and I couldn't see any output. It might be the fact that I only had a few points. For these cases, it could actually be misleading to have a blank plot. As I'm a bit wary to change the function definition and its ramifications throughout, I'll revert it back to 0.2 and let you decide the best way to make it as an optional argument.

src/simsopt/field/tracing.py
rogeriojorge: Having aspect='auto' tends to distort flux surfaces. For example, if the aspect is not 'equal', a circular flux surface will appear elliptical.

rogeriojorge:
The definition of canonical angular momentum has been changed to account for a factor of 2*pi in psi_poloidal. This change resulted in a significant reduction in the variation of the angular momentum, which leads me to believe that the current definition is now correct. The test is now more robust with more particles and for longer times. I believe that this is ready to be merged.

Issue #117, opened 2021-07-29
Visualization optional dependencies


StanczakDominik:
I see Mayavi is used for plotting e.g. here: https://simsopt.readthedocs.io/en/latest/simsopt.geo.html#simsopt.geo.surface.Surface.plot

However, the Docker container does not include it. Would it perhaps make sense to add somewhere around

https://github.com/hiddenSymmetries/simsopt/blob/b3f3c3165d7ad9c333787e1ef2bd7ad08023a3b4/setup.cfg#L62-L69

a list of visualization package (matplotlib, mayavi, have I missed anything?) versions that work with the rest of the package?

Of course, I know mayavi can be a tough cookie to install due to its heavy VTK dependency. Matplotlib, at least, could be good candidate for now, especially for inclusion into the Docker image.

Whatcha think? :)

mbkumar:
@StanczakDominik,

Thanks for the feedback. I added mayavi to the docker container. Mayavi and its requirements are also made optional requirements for simsopt. Matplotlib is made as a mandatory requirement.

Issue #91, opened 2021-06-08
Auto versioning the code


mbkumar:
The current manual increment of the version string is not feasible in the long run. Instead, investigate tools like bumpversion and versioneer to increment the version.

StanczakDominik:
There's also https://github.com/pypa/setuptools_scm/ that lets you pull the version from the git tag. Just leaving it here as another possibility.

mbkumar:
@StanczakDominik 

Thanks! I got setuptools_scm working and PR #119 should add the auto-versioning capability.

PR #119, opened 2021-07-29
Scm autoversion

mbkumar:
Auto-versioning by setuptools_scm package.
After a tag is created, for all subsequent commits, setuptools_scm will generate a new version that is dependent on the version specified in the last tag, the number of new commits, and the commit hash. Works for both wheels and sdist.

Addresses #117 and #91

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,13 +55,13 @@ jobs:
         sudo apt-get install -y build-essential gfortran openmpi-bin libopenmpi-dev libnetcdf-dev libnetcdff-dev liblapack-dev libblacs-mpi-dev libscalapack-mpi-dev libhdf5-dev libhdf5-serial-dev git m4 libfftw3-dev libboost-all-dev
 
     - uses: actions/checkout@v2
-      # If we want submodules downloaded, uncomment the next 2 lines:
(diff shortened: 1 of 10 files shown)


mbkumar:
Added mayavi to the docker build as well. And added optional dependencies for pyevtk, and mayavi and its dependencies. Fixed some bugs in the docker workflow.

mbkumar:
@florianwechsung @landreman 
This PR is ready for merging.

florianwechsung:
Not really an expert on this, but looks sensible. Can you summarise the changes to required/optional packages for visualisation too? just for our own records.

florianwechsung (dismissed):

mbkumar:
> Not really an expert on this, but looks sensible. Can you summarise the changes to required/optional packages for visualisation too? just for our own records.

Sure. 

For visualization, matplotlib is now a mandatory dependency. So it is always available. We can remove the checks for matplotlib. 

vtk, pyevtk, pyqt5, and mayavi are optional dependencies which could be installed along with simsopt by running the command `pip install simsopt[VIS]`

mbkumar:
Updated the docker container image to include pyevtk.

landreman (approved):

Issue #118, opened 2021-07-29
`minimize_curve_length.py` example does not converge to a circle?


StanczakDominik:
To quote:

> Minimize the length of a curve, holding the 0-frequency Fourier mode fixed.
The result should be a circle.

I ran the example using the provided Docker image with `simsopt.__version__` being `0.4.0`, in a Jupyter notebook, then plotted the curve using
```
%pip install matplotlib
%matplotlib notebook
curve.plot()
```

![image](https://user-images.githubusercontent.com/11289391/127473714-b3c23a24-aa79-4267-962f-f50eaeb716b6.png)

It doesn't seem to be particularly circular, no mater how I rotate it! 😉 

<details>
<summary>Print-out from the execution</summary>

```
WARNING:absl:No GPU/TPU found, falling back to CPU. (Set TF_CPP_MIN_LOG_LEVEL=0 and rerun for more info.)

Initial curve dofs:  [ 3.         -0.47167117 -0.14711749 -0.36173382 -0.12887794 -0.19145344
  0.37352108  0.18049693  0.34338664]
Initial curve length:  47.51029349787225
finite difference Jacobian:
[[ -4.16447993   1.08533634 -31.92883693 -17.17364281  -2.05079456
   10.62181397   9.67475973  49.24442454]]
Analytic Jacobian:
[[ -4.16447995   1.08533638 -31.92883691 -17.17364278  -2.05079457
   10.62181394   9.67475967  49.24442449]]
Difference:
[[ 2.91588123e-08 -4.27509244e-08 -1.22633423e-08 -3.29675203e-08
   1.17619621e-08  2.69913691e-08  6.07857036e-08  5.03375546e-08]]
   Iteration     Total nfev        Cost      Cost reduction    Step norm     Optimality   
       0              1         1.1286e+03                                    2.34e+03    
       1              2         6.8936e+02      4.39e+02       8.47e-01       2.22e+03    
       2              4         3.3791e+02      3.51e+02       2.12e-01       8.36e+02    
       3              6         2.6535e+02      7.26e+01       1.06e-01       3.18e+02    
       4              7         2.1860e+02      4.68e+01       2.12e-01       4.04e+02    
       5              9         2.0364e+02      1.50e+01       5.30e-02       1.42e+02    
       6             10         1.9599e+02      7.65e+00       5.30e-02       1.22e+02    
       7             11         1.8839e+02      7.60e+00       1.06e-01       2.24e+02    
[...]

PR #120, opened 2021-07-30
more sensible plotting defaults

florianwechsung:
adresses some of the concerns raised in #118 

- add equal scaling for 3d plots of curves
- set `closed=False` by default for curves

### DIFF
--- a/examples/1_Simple/minimize_curve_length.py
+++ b/examples/1_Simple/minimize_curve_length.py
@@ -14,7 +14,7 @@ The result should be a circle.
 # Create a curve:
 nquadrature = 100
 nfourier = 4
-nfp = 5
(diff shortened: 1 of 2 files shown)


florianwechsung:
I should add that `ax.set_aspect('equal')` is broken in matplotlib (see e.g. this issue https://github.com/matplotlib/matplotlib/issues/15382). The workaround I added works fine for plotting a single curve, but won't give perfect scaling for multiple objects. essentially one just puts a cube around each object.

I don't think we need to implement a fully general solution here (if we did, it should be added to matplotlib directly and not our code), since we support two other forms of visualising curves that don't suffer from this issue.

landreman (approved):

PR #123, opened 2021-08-01
minor tracing improvements

florianwechsung:
- added `phase_angle` option for full orbit
- return first and last position when `forget_exact_path=True`


src/simsopt/field/tracing.py
```diff
@@ -167,7 +170,7 @@ def trace_particles(field: MagneticField, xyz_inits: NDArray[Float],
         if not forget_exact_path:
             res_tys.append(np.asarray(res_ty))
         else:
-            res_tys.append([])
+            res_tys.append(np.asarray([res_ty[0], res_ty[-1]]))
```
rogeriojorge, line 173: Codecov patch is throwing a warning:

Codecov
src/simsopt/field/tracing.py#L173
Added line #L173 was not covered by tests

Perhaps a test with forget_exact_path==True is missing

src/simsopt/field/tracing.py
```diff
@@ -216,8 +220,10 @@ def trace_particles_starting_on_curve(curve, field, nparticles, tmax=1e-4,
             `gc`: general guiding center equations,
             `gc_vac`: simplified guiding center equations for the case :math:`\nabla p=0`,
             `full`: full orbit calculation (slow!)
-        forget_exact_path: return an empty list for the ``res_tys``. To be used when only res_phi_hits
-                           is of interest and one wants to reduce memory usage.
+        forget_exact_path: return only the first and last position of each
+                           particle for the ``res_tys``. To be used when only res_phi_hits is of
+                           interest or one wants to reduce memory usage.
+        phase_angle: the phase angle to use in the case of full orbit calculations
```
rogeriojorge, line 226: It may not be clear to the user what phase_angle is or what impact can it have. Maybe it would be too lengthy to write here. A better place might be in the "concepts" section in the documentation, where we could mention in a single line what are the initial conditions needed by the solver.

rogeriojorge (commented):
Two minor suggestions

florianwechsung:
Thank you @rogeriojorge, I made the requested changes.

rogeriojorge (approved):

PR #126, opened 2021-08-09
small typo fix

tmqian:
suggesting fix for minor grammar typo in Curves documentation

A simsopt.geo.curve.Curve is modelled as a function 
. A curve object stores a list of 
 quadrature points 
 and return**s** all information about the curve, at these quadrature points.


landreman (approved):
Thanks for the fix!

PR #133, opened 2021-08-23
Add missing file to examples script

StanczakDominik:
Hey, this may not be really a helpful PR, but I figured it'd be the best way to ping you all about this :) I was trying to run the examples, in random order, basically, and I stumbled upon `graph_surf_vol_area.py`. Turns out it's not running in the current Docker build. That looked suspicious, wouldn't your CI catch that? Well, it would have, if only this particular file had been included in it!

I suspect #130 may be somewhat superseding that particular example, so, feel free to close this one if that's the case.

### DIFF
--- a/examples/run_serial_examples
+++ b/examples/run_serial_examples
@@ -6,6 +6,7 @@ set -ex
 
 ./1_Simple/just_a_quadratic.py
 ./1_Simple/surface_volume_and_area.py
+./1_Simple/graph_surf_vol_area.py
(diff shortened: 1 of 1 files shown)


mbkumar:
@StanczakDominik,

Thanks for bringing this to our attention.

That example file is a vestige of an earlier implementation of the SurfaceRZFourier class, which was refactored in a major way. The file is deleted in a branch soon to merged into the master .

Issue #134, opened 2021-08-23
Mayavi vs Pyvista, from the docker standpoint


StanczakDominik:
I'm doing my JOSS review out of the provided Docker image and I'm just lamenting the fact that Mayavi keeps crashing my Jupyter server.

From what I can tell, it seems that Mayavi is used only in these two locations:

https://github.com/hiddenSymmetries/simsopt/blob/77c32e08b33c660bfd300ba2c2f5b2fbc9bc3ab3/src/simsopt/geo/surface.py#L21-L50

https://github.com/hiddenSymmetries/simsopt/blob/af234540745b7e37c5d6cc737e74cf65d7d09299/src/simsopt/geo/curve.py#L140-L149

I suspect that if this was (simply, it looks like) rewritten using PyVista, we could then take advantage of https://docs.pyvista.org/extras/docker.html and get this to work from within the notebook! Is there anything I'm missing here? In particular, is there any tighter coupling/investment between Simsopt and Mayavi that I'm missing?

---

Of course, the whole issue of running this out of a Docker container is a side effect of SPEC being private, and the Docker image being the only way I can get my hands on it for review purposes - but that's a rant that I don't want to get into 😅 

---

Let me add that this should in no way be considered a blocker for the JOSS review. Just putting the idea out there.

mbkumar:
@StanczakDominik,
Sorry to hear that Stan. How about increasing the memory of the container to 8 GB and try again in case you are running the container on mac?

@florianwechsung,
could you please look into Stan's suggestion when you have time?

florianwechsung:
Hi @StanczakDominik 

There isn't any tight coupling between SIMSOPT and mayavi, and I'm all onboard with getting rid of mayavi, as it has given me trouble a few times now. 

Most objects at this point can be exported to vtk for plotting purposes (using https://github.com/pyscience-projects/pyevtk), but having some form of visualisation inside jupyter notebooks would be nice. I will have a look at pyvista when I'm back from holiday next week.

landreman:
Resolved by #137

PR #137, opened 2021-08-30
Add more surface plot options

zhucaoxiang:
This is a quick fix to #134.

I have adopted multiple plot engines to plot the toroidal surface in 3D.

What I have changed:
   - add options for using `Matplotlib` and `Plotly`.
   - add `**kwargs` to pass advanced plotting arguments.
 
Something we can discuss:
  - I haven't prepared the options `plot_normal`, `plot_derivative` for the other two engines. It might be messy, especially when using `matplotlib`. 
  - I forgot which docstring format we are using. I looked at other parts and try to follow similar conventions. 
  - The default engine is `matplotlib`, which can be changed to `mayavi` for backward capabilities.


zhucaoxiang:
I have updated the plotting functions for both curves and surfaces, but I have problems testing it on my MacBook #139 . By the way, I didn't found too many examples using `surface.go` (maybe I missed). For example, in this case, I would like to create a toroidal surface and plot it. I don't find an example showing how to do this.

For a demonstration of plotly, you can check this page. https://zhucaoxiang.github.io/CoilPy/examples/Plot_with_plotly.html

florianwechsung:
I'm not sure what you are referring to with `surface.go`. Can you expand?

To create a toroidal surface in simsopt, simply 
```
from simsopt.geo.surfacerzfourier import SurfaceRZFourier
s = SurfaceRZFourier()
```
This defaults to a torus with major radius 1 and minor radius 0.1.

zhucaoxiang:
@florianwechsung Sorry, I was saying `geo.Surface`. But what you provide should enough for a test (after I installed it on my MacBook).

zhucaoxiang:
I think it is now working. Here is a simple example plotting both curves and surfaces using plotly (forgive me that I chose a curve lying inside the surface).
image

I have add some docstring in the source introducing the arguments. I am happy to add some documentation as long as someone can kindly direct me where I should put.

landreman:
This looks excellent. Thanks @zhucaoxiang for working on this. I will try to add a basic unit test for plotting today.

landreman (approved):

PR #140, opened 2021-09-08
Surface and MHD classes converted to graph framework

mbkumar:
This PR converts the surface and MHD modules to the graph framework. 

Changes Implemented:
1) Enhancements to graph Optimizable class to keep order of the dependent functions
2) ObjectiveFailure added to LeastSquaresProblem
3) Tolerances can be specified as part of the solver
4) Surface classes and the corresponding tests are separated into different modules
5) Surface  modules converted to graph framework
6) MHD modules converted to graph framework

Things to be implemented:
1) Jacobian implementation.


tests/mhd/test_integrated_vmec_mpi.py
```diff
@@ -48,7 +48,8 @@ def test_stellopt_scenarios_1DOF_circularCrossSection_varyR0_targetVolume(self):
         logger = logging.getLogger(__name__)
 
         for ngroups in range(1, 1 + MPI.COMM_WORLD.Get_size()):
-            for grad in [False, True]:
+            # for grad in [False, True]:
+            for grad in [False]:
```
landreman, line 52: We'll need to remember to restore this to `[False, True]` once mpi finite difference gradients are working again.

landreman (commented):
Generally this all looks good to me. The one main issue, which we already discussed, is that in `solve.graph_mpi.least_squares_mpi_solve`, mpi finite difference gradients need to work, or else production vmec optimizations will no longer be feasible.

Master needs to be merged in, and I expect there will be a few minor merge conflicts related to `Surface.plot()`

In `tests/solve/test_graph_mpi.py` there are `MpiPartitionTests`. A while ago, these tests moved out of the non-graph `tests/solve/test_mpi.py` to `tests/util/test_mpi_partition.py`

mbkumar:
@landreman,
Do you want to save the jacobians in the same log files? Or can I use a different set of files to save the jacobians if necessary?

mbkumar:
> The issue with the vmec_adjoint.py example was that the Vmec class has a few of its own dofs (e.g. toroidal flux) beyond those owned by the Surface object, so we needed `surf.dofs_free_status` instead of `obj.dofs_free_status` to get an array of the desired length. I just pushed a fix.

I realized that. However using surf.dofs_free_status is not right because that behavior is not natural and a user writing a driver script has to know lot of internals from the dJ method of IotaMetric.

PR #141, opened 2021-09-08
Plotting enhancements

landreman:
* Added tests for `Surface.plot()` and `Curve.plot()`.
* Added option to close the seam in surfaces where the angles wrap from 2pi to 0.
* Separated out a `fix_matplotlib_3d()` function, used by both `Surface.plot()` and `Curve.plot()`, and which might have other uses.
* Tweaks to docstrings for plotting functions.


zhucaoxiang:
It looks good to me. The only thing I would suggest is to use a consistent argument name (and default value) for `close=True` in surface and `closed_loop=False` in curve. This is optional and up to you @landreman .

landreman:
> It looks good to me. The only thing I would suggest is to use a consistent argument name (and default value) for `close=True` in surface and `closed_loop=False` in curve. This is optional and up to you @landreman .

Good point. Fixed.

zhucaoxiang (approved):

PR #143, opened 2021-09-12
QFM Speedup

florianwechsung:
Rewrote the derivative of the QFM objective to first compute the partial derivative of the objective wrt to surface position and normal, and then pass those to the appropriate vjp functions that are already implemented in the C++ code. Results in a speedup of ~3 for the example I timed it on.


ejpaul:
Thanks, Florian, for implementing this! I also observed ~3x speedup in the qfm example on my laptop. Just a few small comments:

- Perhaps rename "num" and "denom" as "J1" and "J2": 
https://github.com/hiddenSymmetries/simsopt/blob/cb08aa92bc731cb03d732e6b6897573abebf5870/src/simsopt/geo/surfaceobjectives.py#L348-L349
- It looks like you did not modify the second derivative calculation. At the moment this functionality is unused, and a significant amount of time in test_surface_objectives.py and test_qfm.py is spent testing the Hessian. What do you think about getting rid of "d2J_by_dsurfacecoefficientsdsurfacecoefficients" and the associated tests? We could bring this code back in the future if we decide to use a Newton method.

florianwechsung:
Thank you for the quick review @ejpaul I made both changes that you suggested. Could you confirm that I didn't miss any traces of the second derivatives?

ejpaul:
Looks good @florianwechsung. Thanks! I guess it is only the codecov that needs to be addressed now.

florianwechsung:
The code coverage has just gone down because the lines that I deleted were previously all tested. We haven't actually generated any untested code, so we can ignore that.

ejpaul (approved):

PR #145, opened 2021-09-21
New Surface quadpoints initialization

landreman:
This PR implements the new methods for initializing Surface quadpoints that we have discussed.

Now, the quadrature points can be initialized via

    s = SurfaceXYZFourier(range="half period", nphi=128, ntheta=64)

or

    s = SurfaceRZFourier(range=SurfaceRZFourier.RANGE_FULL_TORUS, nphi=128, ntheta=64)

where, as shown in these examples, the phi range can be set by either a string or a constant. You can still specify arrays for quadpoints_phi and/or quadpoints_theta directly instead if you like, as previously. The new initialization methods work for and are tested for all Surface subclasses.

The bit of code that handles setting `quadpoints_phi` from `range` and `nphi` is pure python, so nothing was added to C++. It would have been nice to put this new bit of code in the python `Surface` constructor, but I don't think this is possible. The reason is that in the different Surface sub-types, the C++ constructor has to be called before the python constructor (so e.g. `self.mpol` is set before the dof names are set before the Optimizable constructor is called.) Hence a `get_quadpoints` function was introduced in the Surface base class, and it is called in the constructor of each Surface sub-type. See what you think of this implementation.

If you're happy with this implementation, I'll polish some docstrings and other docs before we merge.

A little bit of C++ code was removed since it is now required that `quadpoints_phi` and `quadpoints_theta` are lists and not integers when the C++ Surface constructors are called.

Note that this PR is based on the `graph_surface` branch instead of `master` to minimize merge conflicts with Bharat's work.

### DIFF
--- a/src/simsopt/_core/graph_optimizable.py
+++ b/src/simsopt/_core/graph_optimizable.py
@@ -1189,6 +1189,7 @@ def make_optimizable(func, /, *args, dof_indicators=None, **kwargs):
         Subclass of Optimizable class to create optimizable objects dynamically.
         dof_indicators argument is used to filter out dofs and
         """
+
         def __init__(self, func, /, *args, dof_indicators=None, **kwargs):
 
             self.func = func
--- a/src/simsopt/_core/util.py
+++ b/src/simsopt/_core/util.py
@@ -236,11 +236,13 @@ def nested_lists_to_array(ll):
                 arr[jm, jn] = x
     return arr
 
+
 class WeakKeyDefaultDict(WeakKeyDictionary):
     """
     A simple implementation of defaultdict that uses WeakKeyDictionary as its
     parent class instead of standard dictionary.
     """
+
     def __init__(self, default_factory=None, /, *args, **kwargs):
         self.default_factory = default_factory
         super().__init__(*args, **kwargs)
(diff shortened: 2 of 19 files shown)


florianwechsung:
Thank you, I think this looks good. Just one comment:
I'm not sure how I feel about passing all of `**kwargs` on to `get_quadpoints`. I'm concerned that if someone passes a wrong argument to a surface constructor that has nothing to do with quadpoints, the error message will lead them astray. e.g. if somebody write `Mpol` instead of `mpol`, that will create an error in `get_quadpoints`. It also makes the function signature a bit harder to read, since it doesn't actually mention quadpoints now. I'm not quite sure how to fix this, but if we keep it as is, the docstrings should mention all these arguments that are hidden in `kwargs`.

landreman:
Yeah, I wasn't sure about the `**kwargs` approach, and agree with the issues you point out. I used the `**kwargs` approach since it made the `__init__` methods of each Surface type shorter. But writing out the 5 arguments explicitly is not too bad:

    def __init__(self, nfp=1, stellsym=True, mpol=1, ntor=0,
                 nphi=None, ntheta=None, range="full torus",
                 quadpoints_phi=None, quadpoints_theta=None):
        quadpoints_phi, quadpoints_theta = Surface.get_quadpoints(nfp=nfp,
                                                                  nphi=nphi, ntheta=ntheta, range=range,
                                                                  quadpoints_phi=quadpoints_phi,
                                                                  quadpoints_theta=quadpoints_theta)
        sopp.SurfaceRZFourier.__init__(self, mpol, ntor, nfp, stellsym,
                                       quadpoints_phi, quadpoints_theta)
        ...

I think you convinced me to switch to this approach. I'll implement this, and add to docstrings.

PR #147, opened 2021-09-23
Fix reference cycles between python and C++

florianwechsung:
Kick out py_shared_ptr and instead keep the python instance alive manually by holding a reference to it

See pybind/pybind11#1389 for why py_shared_ptr was needed in the first place, and the comment from May 27 why we may not want to use it (reference cycle)


mbkumar (approved):

mbkumar:
Thanks @florianwechsung

PR #151, opened 2021-10-04
CircularCoil fix

rogeriojorge:
This PR fixes a known issue in the magnetic field class called CircularCoil. It was seen that although the magnetic field and its derivatives were correct at some angles, not all angles were giving sensible results. Furthermore, when testing for a circular arrangement of toroidal coils, periodicity was not being observed.
The rotation matrix was missing important elements. The new rotation matrix is now obtained using the Mathematica function _RotationMatrix_, which gives matrices for rotations of vectors around the origin. The formula, made in such a way that the Bfield as the necessary orientation to match the one with the BioSavart class with the same current, is
`RotationMatrix[phi, {-Sin[theta], -Cos[theta], 0}]`
Theta and Phi angles are spherical angles as often used in mathematics: 
x=rho sin(phi) cos(theta)
y=rho sin(phi) sin(theta
z=rho cos(phi)

### DIFF
--- a/src/simsopt/field/magneticfieldclasses.py
+++ b/src/simsopt/field/magneticfieldclasses.py
@@ -274,19 +274,34 @@ class CircularCoil(MagneticField):
         self.Inorm = I*4e-7
         self.center = center
         if len(normal) == 2:
-            self.normal = [normal[0], -normal[1]]
+            theta = normal[0]
+            phi   = normal[1]
         else:
-            self.normal = [np.arctan2(np.sqrt(normal[0]**2+normal[1]**2), normal[2]), -np.arctan2(normal[0], normal[1])]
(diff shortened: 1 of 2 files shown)


zhucaoxiang:
@rogeriojorge The test passes on my side. But I am trying to find one more example for cross-checking. I have the general rotation matrix using the yaw, pitch angles at https://github.com/zhucaoxiang/CoilPy/blob/master/coilpy/misc.py#L722. I guess we can generate a standard circular coil, c0, with normal=[0,0,1] and generate an arbitrary coil, c1, with a new normal. I can try to reproduce c1 from c0 using my rotation matrix. What do you think?

rogeriojorge:
@zhucaoxiang That seems like a good test. For c1, using your script, could you make a circular coil with an arbitrary normal vector, center, radius and current? Then you could send me the magnetic field at a few points and I can use them as a test for a coil with SIMSOPT's CircularCoil class with the same parameters.

PR #164, opened 2021-11-21
Only initialize MPI if necessary

landreman:
In the master branch of simsopt presently, MPI is initialized whenever any part of simsopt is imported, even just `import simsopt`. This is a problem because some HPC systems, such as Cobra and Raven at IPP, only allow MPI to be initialized when programs are submitted through slurm, not on login nodes. This means simsopt is not presently usable on the login nodes even for non-computationally-expensive scripts. This might also be frustrating for new users, if they try `import simsopt` on a login node and it crashes with an MPI error. This PR fixes this issue by modifying some imports so that MPI is only initialized when it is really necessary, such as for the `MpiPartition` or `Vmec` classes. A test was also added to the CI to verify that `import simsopt` and importing a surface class do not initialize MPI now.

### DIFF
--- a/tests/verify_MPI_not_initialized.py
+++ b/tests/verify_MPI_not_initialized.py
@@ -14,17 +14,11 @@ unit tests, because otherwise one of the unit tests that uses MPI
 would initialize MPI.
 """
 
+import sys
 import simsopt
 from simsopt.geo.surfacerzfourier import SurfaceRZFourier
 
-import mpi4py
-# The next line tells mpi4py to not initialize MPI when mpi4py.MPI is imported:
-mpi4py.rc.initialize = False
+assert "mpi4py.MPI" not in sys.modules, \
+    "Importing simsopt should not initialize MPI"
 
-from mpi4py import MPI
-
-if mpi4py.MPI.Is_initialized():
-    print("ERROR! Importing simsopt should not initialize MPI")
-    exit(1)
-else:
-    print("Verified that importing simsopt does not initialize MPI")
+print("Verified that importing simsopt does not initialize MPI")
--- a/tests/verify_MPI_not_initialized.py
+++ b/tests/verify_MPI_not_initialized.py
@@ -1,0 +1,30 @@
+#!/usr/bin/env python
+
+"""
+Some HPC systems only allow MPI to be initialized when jobs are
+submitted through slurm, not on login nodes.  To make the non-MPI
+parts of simsopt usable on login nodes, then, we want to be sure that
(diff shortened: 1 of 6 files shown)


tests/verify_MPI_not_initialized.py
```diff
@@ -0,0 +1,30 @@
...
+# The next line tells mpi4py to not initialize MPI when mpi4py.MPI is imported:
+mpi4py.rc.initialize = False
+
+from mpi4py import MPI
+
+if mpi4py.MPI.Is_initialized():
+    print("ERROR! Importing simsopt should not initialize MPI")
+    exit(1)
+else:
+    print("Verified that importing simsopt does not initialize MPI")
```
mbkumar, line 30: Instead of this complicated setup, a simple approach would be to check for mpi4py in loaded modules. Top level import of simsopt shouldn't load mpi4py. This can be checked with these few lines of code
```
import sys
assert("mpi4py" not in sys.modules)
```

Lines 20  to end can be replaced with the above two lines.

landreman: Here's the reason the new test was written that way. `FiniteDifference` and `MpiFiniteDifference` are in the same module, with the serial `FiniteDifference` imported when the top-level simsopt module is imported via `least_squares_serial_solve`. So the `@simsoptRequires` in front of `MpiFiniteDifference` requires that simsopt tries to import mpi4py when a user calls `import simsopt` or imports the serial `FiniteDifference`. Importing `mpi4py` does not initialize MPI; usually importing `mpi4py.MPI` does, but not if `mpi4py.rc.initialize=False`. So checking `"mpi4py" in sys.modules` does not generally tell us whether MPI has been initialized. Whereas `mpi4py.MPI.Is_initialized()` gives the answer reliably. In `finite_difference.py` and perhaps elsewhere in the future, it can be useful to `try: import mpi4py` to see if mpi4py has been installed in a module that we still may wish to use on a login node, i.e. without initializing MPI.

mbkumar:
In that case, sys modules should have mpi4py.MPI as one of the modules.
Let me do a few checks regarding the approach I proposed.

mbkumar:
In the current master branch, importing simsopt imports mpi4py and
mpi4py.MPI modules. So checking for mpi4py.MPI should be sufficient. After
installing your branch

>>> import simsopt

>>> import sys

>>> "mpi4py" in sys.modules

True

>>> "mpi4py.MPI" in sys.modules

False

Bharat Medasani

Engineer

mbkumar (approved):

PR #165, opened 2021-12-01
Performance improvements in SurfaceRZFourier

florianwechsung:
I added two tricks to speed up some of the functions in `SurfaceRZFourier`.
1) Use SIMD instructions to parallelise across the angle theta.
2) This parametrization requires the evaluation of 
`         sin(m*theta-n*nfp*phi) and cos(m*theta-n*nfp*phi)`
   for many values of n and m. Since trigonometric functions are expensive,
   we want to avoid lots of calls to sin and cos. Instead, we use the rules
`         sin(a + b) = cos(b) sin(a) + cos(a) sin(b)`
`         cos(a + b) = cos(a) cos(b) - sin(a) sin(b)`
   to write
`         sin(m*theta-(n+1)*nfp*phi) = cos(-nfp*phi) * sin(m*theta-n*nfp*phi) + cos(m*theta-n*nfp*phi) * sin(-nfp*phi)`
`         cos(m*theta-(n+1)*nfp*phi) = cos(m*theta-n*nfp*phi) * cos(-nfp*phi) + sin(m*theta-n*nfp*phi) * sin(-nfp*phi)`
   In our code we loop over n. So we start with n=-ntor, and then we always
   just increase the angle by -nfp*phi.

For very high order surfaces, this speeds up computations such as computing a QFM surface by almost an order of magnitude.


florianwechsung:
Should say that similar ideas could probably be used to speed up computations of the other geometric objects. So if anyone sees bottlenecks in their code related to the other Curve or Surface classes, the same ideas as here can likely be applied.

srlantz:
Could some of these trig formulas get into accuracy issues due to subtraction of near-equal numbers? For example, when a is close to +/- b.

florianwechsung:
I recompute the angles using sin and cos every 5 steps, so errors of this kind don't accumulate.

andrewgiuliani (approved):
neat

Issue #178, opened 2021-12-21
Error with MPI logging 


ejpaul:
I am running into an error with calling initialize_logging() with mpi=True. This can be seen by running the 1_Simple/logging_example.py, whose output is copied below. 

Running 1_Simple/logger_example.py
==================================
Hello (times 1) from serial job
Hello (times 2) from serial job
Traceback (most recent call last):
  File "/Users/elizabeth/opt/miniconda3/envs/vmec2000_conda/lib/python3.8/logging/config.py", line 563, in configure
    handler = self.configure_handler(handlers[name])
  File "/Users/elizabeth/opt/miniconda3/envs/vmec2000_conda/lib/python3.8/logging/config.py", line 723, in configure_handler
    if issubclass(klass, logging.handlers.MemoryHandler) and\
TypeError: issubclass() arg 1 must be a class

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "examples/1_Simple/logger_example.py", line 25, in <module>
    initialize_logging(mpi=True, filename='mpi.log')
  File "/Users/elizabeth/opt/miniconda3/envs/vmec2000_conda/lib/python3.8/site-packages/simsopt-0.6.1.dev129+gbdf9974-py3.8-macosx-10.9-x86_64.egg/simsopt/util/logging.py", line 45, in initialize_logging
    logging.config.dictConfig(config_dict)
  File "/Users/elizabeth/opt/miniconda3/envs/vmec2000_conda/lib/python3.8/logging/config.py", line 808, in dictConfig
    dictConfigClass(config).configure()
  File "/Users/elizabeth/opt/miniconda3/envs/vmec2000_conda/lib/python3.8/logging/config.py", line 570, in configure
    raise ValueError('Unable to configure handler '
ValueError: Unable to configure handler 'mpi_file_handler'

mbkumar:
@ejpaul, I'll take a look into it.

Issue #159, opened 2021-11-05
@requires decorator confuses sphinx


landreman:
When the docs have a link to a class, using e.g.

    :obj:`simsopt.util.mpi.MpiPartition`

the link does not work if the class has the `@requires` decorator. I noticed this problem in particular for the links to `MpiPartition` in
https://simsopt.readthedocs.io/en/latest/concepts.html#mpi-partitions-and-worker-groups
If the `@requires` decorator is removed from `MpiPartition`, the links in the docs work again.

landreman:
The same problem occurs if the `SimsoptRequires` decorator is used instead of `requires`.

PR #184, opened 2021-12-23
Logging fix

mbkumar:
This PR fixes bugs encountered in importing `initialize_logging` from `simsopt.util.logging` reported in #178. 

`functools.wraps` works only for functions, so using `SimsoptRequires`, which uses `functools.wraps` internally, on classes won't work as expected. Hence, `SimsoptRequires` is removed in MHD modules.


landreman:
If you are removing `SimsoptRequires` from classes, does it make sense to also remove `monty.dev.requires` (which also uses `functools.wraps`, see http://guide.materialsvirtuallab.org/monty/_modules/monty/dev.html#requires) from `simsopt.util.mpi.MpiPartition`? This is causing issue #159.

mbkumar:
Earlier monty.dev.requires was not using functools.wraps. In recent
versions, they might have changed it. This change obsoletes the needs for
SimsoptRequires.

Anyhow, let's not use either of them to decorate classes because
functools.wraps is not working for classes.

Bharat Medasani

Engineer

mbkumar:
Removed `monty.dev.requires` from `simsopt.util.mpi.MpiPartition` to address  #159.

Issue #186, opened 2022-01-02
graph_optimizable: when a parent is appended, children do not update


landreman:
Here is a behavior that I think is not intended. Suppose we have two Optimizable objects, a parent and child. When a third "grandparent" object is appended as a parent to the parent, the child object presently does not update its state vector to include the new dofs of the grandparent. Here is a minimal example, using some mock classes for simplicity:
```python
from simsopt._core.graph_optimizable import Optimizable

class LeastSquaresProblem(Optimizable):
    def __init__(self, obj):
        super().__init__(depends_on=[obj])

class Vmec(Optimizable):
    def __init__(self):
        super().__init__(x0=[1, 2, 3])

class PressureProfile(Optimizable):
    def __init__(self):
        super().__init__(x0=[10, 20])

# vmec is the parent, prob is the child
vmec = Vmec()
prob = LeastSquaresProblem(vmec)

print('Before adding grandparent:')
print('  vmec.dof_size:', vmec.dof_size, ' vmec.x:', vmec.x)
print('  prob.dof_size:', prob.dof_size, ' prob.x:', prob.x)

pressure = PressureProfile()
vmec.append_parent(pressure)
# pressure is the grandparent

print('After adding grandparent:')
print('  vmec.dof_size:', vmec.dof_size, ' vmec.x:', vmec.x)
print('  prob.dof_size:', prob.dof_size, ' prob.x:', prob.x)
```
The output is:
~~~~
Before adding grandparent:
  vmec.dof_size: 3  vmec.x: [1. 2. 3.]
  prob.dof_size: 3  prob.x: [1. 2. 3.]
After adding grandparent:
  vmec.dof_size: 5  vmec.x: [10. 20.  1.  2.  3.]
  prob.dof_size: 3  prob.x: [1. 2. 3.]
~~~~
The output shows that `vmec` did update its state vector to include the new `pressure` dofs `[10., 20.]`, but `prob` did not update its dofs. We probably want `prob` to update its state vector too, right?

PR #187, opened 2022-01-03
Update ancestors for children after a parent is added/removed

mbkumar:
This PR address the issue #186.

Changes introduced:
Ancestor gathering moved to _update_full_dof_size_indices method. This method is called whenever a parent is added/removed recursively for all children. So the ancestors lists in the children are kept up-to-date.


landreman (approved):
Great, I confirmed this fixes #186 on my end. Go ahead and merge when you're ready.

PR #188, opened 2022-01-10
MPIObjective and more geometry objectives

florianwechsung:


src/simsopt/geo/curveobjectives.py
```diff
@@ -242,3 +249,102 @@ def dJ(self):
...
+        return self.curve.dincremental_arclength_by_dcoeff_vjp(
+            self.thisgrad(self.curve.incremental_arclength()))
+
+    return_fn_map = {'J': J, 'dJ': dJ}
+
+
+@jit
+def curve_msc_pure(kappa, gammadash):
+    """
+    This function is used in a Python+Jax implementation of the curve arclength variation.
```
landreman, line 324: arclength variation -> mean squared curvature?

src/simsopt/geo/curveobjectives.py
```diff
@@ -242,3 +249,102 @@ def dJ(self):
...
+    """
+    arc_length = jnp.linalg.norm(gammadash, axis=1)
+    return jnp.mean(kappa**2 * arc_length)/jnp.mean(arc_length)
+
+
+class MeanSquaredCurvature(Optimizable):
+
+    def __init__(self, curve):
+        r"""
+        Compute the mean of the squared curvature of a curve.
```
landreman, line 334: Can we include an explicit formula for `J` here in the docstring, e.g. `(1/L) \int d\ell \kappa^2` where `L` is the curve length and `d\ell` is the incremental arclenth?

tests/objectives/test_utilities.py
```diff
@@ -0,0 +1,49 @@
...
+        deriv = np.sum(dJ * h)
+        err = 1e6
+        for i in range(5, 15):
+            eps = 0.5**i
+            curve.x = curve_dofs + eps * h
+            Jh = J.J()
+            deriv_est = (Jh-J0)/eps
+            err_new = np.linalg.norm(deriv_est-deriv)
+            print("err_new %s" % (err_new))
+            # assert err_new < 0.55 * err or err_new < 1e-13
```
landreman, line 42: Uncomment this line so there is an assert?

landreman (commented):
Would be good to add tests for the MPI bits. From `.github/workflows/ci.yml` line 195-197, I believe tests are run with multiple procs if `unittest discover -k "mpi"` discovers them.

It would be nice to have an example showing all the possible coil objectives somewhere. Should we add another example like `examples/2_Intermediate/stage_two_optimization.py` but more complex, with mean curvature, max curvature, and arclength terms included? Or should we put all those objectives into the existing `stage_two_optimization.py`?

src/simsopt/geo/curveobjectives.py
```diff
@@ -120,10 +113,9 @@ class LpCurveTorsion(Optimizable):
 
     """
 
-    def __init__(self, curve, p):
+    def __init__(self, curve, p, threshold=0.):
```
landreman, line 114: Docstring then would have a `-\tau_0` term, right?

tests/objectives/test_utilities.py
florianwechsung: I like this notation since if `\ell = |\Gamma'|`, then this formula matches the implementation
```
jnp.mean(kappa**2 * arc_length)/jnp.mean(arc_length)
```
closely.

florianwechsung: Actually, comparing to the other objectives, i guess we are not consistent then. I'll change it

landreman (approved):

PR #189, opened 2022-01-16
Use mpich for docker container

mbkumar:
This PR modifies Dockerfile.ubuntu as a preliminary step to get the docker imagefile work with shifter at NERSC. In place of openmpi, mpich is used to take advantage of mpich's ABI compatibility.

Specifically, the docker container build is modified to build mpich and scalapack manually in place of libopenmpi-dev, openmpi-bin, and libscalapack-dev installed with apt-get. libboost-all-dev is also not installed to prevent mpi libraries installed as dependencies by apt-get.

### DIFF
--- a/ci/Dockerfile.ubuntu
+++ b/ci/Dockerfile.ubuntu
@@ -2,10 +2,38 @@
 # ===================================
 FROM ubuntu:20.04 as intermediate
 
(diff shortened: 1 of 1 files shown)


mbkumar:
Testing on my laptop shows that the MPI jobs inside the docker container run with only 1 core similar to the behavior seen on NERSC.

mbkumar:
After rebuilding the VMEC with Matt's changes inside the Docker container, VMEC runs fine with multiples MPI processes.

landreman (approved):

PR #194, opened 2022-01-30
Real-space surface dofs

landreman:
This PR introduces a class `SurfaceRZPseudospectral` which is used to replace the Fourier-space dofs of `SurfaceRZFourier` with real-space dofs, corresponding to the `(r,z)` coordinates of the surface on grid points.  The advantage of having the dofs in real-space is that they are all of the same magnitude,  so it is easier to know what reasonable box constraints are. This class may therefore be useful for stage-1 optimization using algorithms that require box constraints. 

Typical usage:

    vmec = Vmec("input.your_filename_here")
    vmec.boundary = SurfaceRZPseudospectral.from_RZFourier(vmec.boundary)


mbkumar (approved):
Looks good to me.

PR #197, opened 2022-02-04
Write vmec input files

landreman:
This pull request adds a routine to write vmec input files. Now at the end of a stage-1 optimization, the vmec input file for the optimized configuration is available. Previously just the output file was available.


mbkumar (commented):
I have two comments

1. Transition to formatted strings (f-string).
2. Instead of using write_nml to get the nml string or write to the file based on filename argument value, use two functions
    ```python
    def get_nml(self):
          # Produce the nml string
          # return the nml string
 
    def write_nml(self, filename):
           nml_str = self.get_nml()
           with open(filename) as f:
                    f.write(nml_str) 
    ``` 
    This is the preferred way in many codes I have seen.

mbkumar (approved):

mbkumar:
@landreman , 

Some tests are failing. Could you please look into this?

landreman:
@mbkumar all tests pass now in the CI.

mbkumar (approved):

PR #208, opened 2022-03-20
Deprecated optimizable and dofs modules deleted 

mbkumar:
This PR deletes the deprecated optimizable and dofs modules as well as the associated code to define objective functions and solver classes. 

At present, the replacement has ``graph_`` prefix. In future, ``graph_`` prefix will be removed from the names of the replacement modules.


landreman (approved):

landreman:
Looks good, thanks.

PR #209, opened 2022-03-21
Virtual casing

landreman:
This PR provides an interface to the `virtual_casing` package. This will be used to do coil optimization for finite-beta plasmas, replacing the BNORM code in stellopt. The new functions in simsopt adds lots of functionality to `virtual_casing`: providing an interface to Vmec, more tests, ability to save and load results, resample the results to the resolution desired for the stage-2 optimization, plot results, and suggest resolution parameters that minimize grid anisotropy on the surface. The `virtual_casing` module has also been added to the simsopt containers.

Typical usage:
~~~~py
    from simsopt.mhd.virtual_casing import VirtualCasing
    VirtualCasing.from_vmec('wout_li383.nc', nphi=300)
~~~~


examples/2_Intermediate/B_internal_normal.py
```diff
@@ -0,0 +1,32 @@
...
+This example requires that the python virtual_casing module is installed.
+"""
+
+filename = os.path.join(os.path.dirname(__file__), '..', '..',
+                        'tests', 'test_files', 'wout_li383_low_res_reference.nc')
+
+# Only the phi resolution needs to be specified. The theta resolution
+# is computed automatically to minimize anisotropy of the grid.
+vc = VirtualCasing.from_vmec(filename, nphi=180)
+
```
mbkumar, line 24: This example would be confusing to the users because it appears that nothing is happening after instantiating the VirtualCasing class. My suggestion would to be print some of the important quantities and add comments stating how those quantities are used.

src/simsopt/mhd/virtual_casing.py
```diff
@@ -0,0 +1,355 @@
...
+        vc.unit_normal = unit_normal
+        vc.B_internal = Binternal3d
+        vc.B_internal_normal = Binternal_normal
+
+        if filename is not None:
+            if filename == 'auto':
+                directory, basefile = os.path.split(vmec.output_file)
+                filename = os.path.join(directory, 'vcasing' + basefile[4:])
+                logger.debug(f'New filename: {filename}')
+            vc.save(filename)
```
mbkumar, line 227: I'd not call VirtualCasing.save automatically and instead ask the users to explicitly make a call to save the data.

src/simsopt/mhd/virtual_casing.py
```diff
@@ -0,0 +1,355 @@
...
+        return vc
+
+    def save(self, filename="vcasing.nc"):
+        """
+        Save the results of a virtual casing calculation in a NetCDF file.
+
+        Args:
+            filename: Name of the file to create.
+        """
+        f = netcdf_file(filename, 'w')
```
mbkumar, line 238: Please use context manager.

```python
with netcdf_file(filename, 'w') as f:
     ...
```

src/simsopt/mhd/virtual_casing.py
```diff
@@ -0,0 +1,355 @@
...
+        # Check order:
+        index = 0
+        for jxyz in range(3):
+            for jphi in range(nphi):
+                for jtheta in range(ntheta):
+                    np.testing.assert_allclose(Binternal1d[index], Binternal3d[jphi, jtheta, jxyz])
+                    index += 1
+        """
+
+        Btotal_normal = np.sum(B3d * unit_normal, axis=2)
```
mbkumar, line 208: ``Btotal_normal`` is not used anywhere

src/simsopt/mhd/virtual_casing.py
```diff
@@ -0,0 +1,355 @@
...
+    def load(cls, filename):
+        """
+        Load in the results of a previous virtual casing calculation,
+        previously saved in NetCDF format.
+
+        Args:
+            filename: Name of the file to load.
+        """
+        vc = cls()
+        f = netcdf_file(filename, mmap=False)
```
mbkumar, line 301: File is not closed. Again using context manager helps.

src/simsopt/mhd/virtual_casing.py
```diff
@@ -0,0 +1,355 @@
...
+        variables = ['gamma', 'unit_normal', 'B_total', 'B_internal']
+        for variable in variables:
+            oldvar = eval('self.' + variable)
+            newvar = np.zeros((nphi, ntheta, 3))
+            for j in range(3):
+                newvar[:, :, j] = resample_2D(oldvar[:, :, j], nphi, ntheta)
+            newvc.__setattr__(variable, newvar)
+        return newvc
+
+    def plot(self, show=True):
```
mbkumar, line 338: Saving the fig is not enabled. Also, the approach used in the plot method is archaic. Instead, OO-approach should be used to make it more convenient for end-users.

Please refer to https://matplotlib.org/stable/tutorials/introductory/lifecycle.html

mbkumar: My suggestion is to make plot return fig, axes objects so users can further customize the plots. Then they can decide whether to save or show them.

mbkumar (changes requested):

src/simsopt/mhd/virtual_casing.py
mbkumar: I fixed this.

src/simsopt/mhd/virtual_casing.py
mbkumar: Fixed this.

florianwechsung:
Discussion in Princeton:
- interpolation should take angle shift into account
- vmec object should return sum of currents
- only optimize for ncoil-1 currents, set the last one to I_total - \sum_remaining I_i

src/simsopt/mhd/virtual_casing.py
landreman: Good catch. That line has been removed.

jimlobsien:
Here is a list of what I think should be implemented in simsopt concerning the virtual casing principle:

- Component of the internal currents on the boundary (work in progress)
- Volumetric internal currents on a mesh surrounding the plasma boundary

Use-cases for the later (approximating finite beta equilibrium):

-  Poincare plot of coils plus internal currents 
-  Derivation of a current distribution on a current carrying surface 
-  Derivation of properties of the approximated finite beta magnetic field
-  Coil optimization with these properties

florianwechsung:
I think the functionality is all here now. Remaining todo is to fix the docstrings in the virtual_casing class and in the w7x example

florianwechsung:
Docs updated. @landreman @mbkumar I think this is ready for review.

examples/2_Intermediate/B_internal_normal.py
landreman: A bit more is done now in this example.

src/simsopt/mhd/virtual_casing.py
landreman: I'd prefer to keep saving as the default. The reason is that 99% of the time, the use case is to run virtual casing once, save the result, and load in the result many times for stage-2 coil optimizations with different parameters. If a user ever wants to not save a file, they can set `filename=None` when calling `VirtualCasing.from_vmec()`.

src/simsopt/mhd/virtual_casing.py
landreman: This has been addressed.

landreman:
Thanks for all your work on this, @florianwechsung . I'm happy for it to be merged now.

mbkumar (approved):
Looks good to me.

PR #210, opened 2022-03-26
added plot DAG test

andrewgiuliani:
This PR adds a unit test for the plot DAG functionality.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,7 +87,7 @@ jobs:
 
     - name: Install python dependencies
       run: |
-        sudo apt-get install graphviz
+        sudo apt-get install graphviz graphviz-dev
         pip install wheel numpy f90nml scikit-build cmake qsc sympy pyevtk matplotlib ninja plotly networkx pygraphviz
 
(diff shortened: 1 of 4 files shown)


landreman:
Some objects like curves and surfaces have their own plot() functions that will override this plot() function, and more objects may have plot() functions in the future. Would it make sense to rename this function plot_graph()?

andrewgiuliani:
I agree @landreman , I'll make this change.  By the way, see attached as to what the DAG plots actually look like for the three unit tests that I've included -- I've stopped plotting the nodes and only plotted the node name.  @mbkumar mentioned resizing the nodes to be proportional to the node name length, but I was not satisfied with how it looked visually.

test1
test2
test3

florianwechsung:
Can you also return the network at the end of the function? That way it can then directly be used with the network2tikz library without the user having to modify your function.

andrewgiuliani:
Sure, `plot_graph` now returns the networkx graph as well as a dictionary of positions indexed by node name.  These two things can then be given to the network2tikz library to export the DAG to a latex file.

florianwechsung (approved):

PR #211, opened 2022-04-04
Curve-Curve distance improvements and introduction of Curve-Surface distance

florianwechsung:
Summary

- introduced `CurveSurfaceDistance` objective to enforce minimum distance constraints between the coils and a surface
- renamed `MinimumDistance` to `CurveCurveDistance`
- exploit symmetries to avoid computing redundant distances
- added C++ routines that return which pairs of curves are within a certain threshold of each other and which curves are within a threshold of a surface.

The idea behind the last point is:
- Assume we want to compare pointcloud A and B.
- We create a uniform grid of size `threshold`.
- Loop over points in cloud A, mark all cells that have a point in it (blue in the image).
- Loop over points in cloud B, mark all cells that have a point in it (green) and also all neighbouring cells around it (red).
- Check whether the intersection `(green ∪ red) ∩ blue` between the two sets is non-empty.
  - if empty: the two point clouds are *definitely* more than `threshold` apart
  - if non-empty: they *may* be points in the pointclouds closer than `threshold` to each other

image

The above is a necessary condition for two point clouds to have points that are within `threshold`. We then do the exact check afterwards (O(n^2) cost) for all pairs that are still candidates after the above step.

### DIFF
--- a/src/simsoptpp/python_distance.cpp
+++ b/src/simsoptpp/python_distance.cpp
@@ -1,0 +1,177 @@
+#include "pybind11/pybind11.h"
+#include "pybind11/stl.h"
+namespace py = pybind11;
+#include "xtensor-python/pyarray.hpp"     // Numpy bindings
+typedef xt::pyarray<double> PyArray;
+using std::vector;
+using std::tuple;
+using std::set;
+using std::floor;
+
+
+template<class T>
+bool empty_intersection(const set<T>& x, const set<T>& y)
+{
+    auto i = x.begin();
+    auto j = y.begin();
+    while (i != x.end() && j != y.end())
+    {
+        if (*i == *j)
+            return false;
+        else if (*i < *j)
+            ++i;
+        else
+            ++j;
+    }
+    return true;
+}
+
+bool two_points_too_close_exist(PyArray& XA, PyArray& XB, double threshold_squared){
+    for (int k = 0; k < XA.shape(0); ++k) {
+        for (int l = 0; l < XB.shape(0); ++l) {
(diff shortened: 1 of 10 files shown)


florianwechsung:
Now also includes the ability to exploit underlying symmetry in the curves to reduce computations

andrewgiuliani:
neat! I'll take a look.  Do you have any timing comparisons between this optimized algorithm and the previous kdtree algorithm?

florianwechsung:
For 100 iterations of standard stage II:

without coil-coil-distance: 2.2s
with old coil-coil-distance: 5.6s
with new coil-coil-distance: 3.3s

src/simsoptpp/python_distance.cpp
```diff
@@ -0,0 +1,177 @@
...
+        else
+            ++j;
+    }
+    return true;
+}
+
+bool two_points_too_close_exist(PyArray& XA, PyArray& XB, double threshold_squared){
+    for (int k = 0; k < XA.shape(0); ++k) {
+        for (int l = 0; l < XB.shape(0); ++l) {
+            double dist = std::pow(XA(k, 0) - XB(l, 0), 2) + std::pow(XA(k, 1) - XB(l, 1), 2) + std::pow(XA(k, 2) - XB(l, 2), 2);
```
andrewgiuliani, line 32: might save some cycles by hard coding std::pow(XA(k, 0) - XB(l, 0), 2) as (XA(k, 0) - XB(l, 0))*(XA(k, 0) - XB(l, 0))

andrewgiuliani (approved):
Looks good, neat algorithm

PR #216, opened 2022-04-18
Remove assert in tracing.cpp

florianwechsung:
I noticed that one of the tests triggered this assert, when running the test suite on a Debug build of simsoptpp. It never failed on CI because asserts are not active when building in Release mode.

I checked with @ejpaul and she confirmed that this can go.


landreman:
Discussed in the Monday meeting: make the simple CI compile in debug mode, while the extensive CI compiles in release mode. This would cover these asserts without adding to the testing matrix. @florianwechsung

florianwechsung:
In addition to the two asserts, I also changed the build/test system so that the Simple CI is now run with the code compiled in Debug mode. The fact that after [7436e02](https://github.com/hiddenSymmetries/simsopt/pull/216/commits/7436e02f84006f245b6d63055ba3ba89b300d1b3) the Simple CI failed due to an assert in tracing.cpp indicated that this was successful. 

Added @landreman as reviewer to confirm the changes to the CI.

landreman (approved):

PR #218, opened 2022-04-21
Use slots for DOFs

mbkumar:
Reduces the memory footprint of DOFs objects


florianwechsung (approved):
Good idea!

landreman:
I'm curious if this makes a measurable difference. Dofs are usually no more than a few hundred floats.

PR #224, opened 2022-04-25
mpi barrier to avoid wout not found error

landreman:
In stage one optimization, if there are multiple mpi processes per worker group, there can be a problem where proc 0 deletes a wout file before other procs have read it. In this PR an MPI barrier is added to avoid this condition.


daringli (approved):
With master merged into it, produces identical wout files for a PRL-based Fourier-stepping test case (cobra:/u/sbul/umd/matt_2022-02-02) when compared to the codebase in February.

landreman:
Thanks!

PR #227, opened 2022-04-28
Improvements to RegularGridInterpolant3D and C++ documentation #103

### LINKED ISSUE #103: More granular testing and documentation of RegularGridInterpolant

florianwechsung:
- Introduces some general remarks regarding the C++ part of SIMSOPT in `docs/source/cpp.rst` 
- Adds the ability to skip some cells in the mesh to the `RegularGridInterpolant3D` class
- some clean up of the C++ and some more docs of `RegularGridInterpolant3D`

This will close #103


florianwechsung:
To illustrate the `skip` function: we previously would have to pick the bounds of the interpolation domain so that the entire domain of interest is contained. This meant that potentially we were querying the field in regions that we're not going to be interested in after all, see e.g. here

image

(note that the field here satisfies stellarator symmetry and 3-fold rotational symmetry).

With the new capability, we can reduce this and skip many cells, as shown here:

image

src/simsopt/field/magneticfieldclasses.py
```diff
@@ -517,7 +517,7 @@ class InterpolatedField(sopp.InterpolatedField, MagneticField):
     This resulting interpolant can then be evaluated very quickly.
     """
 
-    def __init__(self, field, degree, rrange, phirange, zrange, extrapolate=True, nfp=1, stellsym=False):
+    def __init__(self, field, degree, rrange, phirange, zrange, extrapolate=True, nfp=1, stellsym=False, skip=None):
```
landreman, line 520: How about adding `skip` to the docstring that follows.

examples/1_Simple/tracing_fieldline.py
```diff
@@ -56,14 +56,14 @@
 ntor = 5
 stellsym = True
 nfp = 3
-phis = np.linspace(0, 1, nfp*2*ntor+1, endpoint=False)
-thetas = np.linspace(0, 1, 2*mpol+1, endpoint=False)
+phis = np.linspace(0, 1, 64, endpoint=True)
+thetas = np.linspace(0, 1, 24, endpoint=True)
```
landreman, line 59: Usually we have `endpoint=False` for the quadpoints. Could you add a comment here to explain why we need `endpoint=True` in this scenario?

examples/1_Simple/tracing_fieldline.py
```diff
@@ -86,16 +86,45 @@ def trace_fieldlines(bfield, label):
...
 
-n = 16
+# Bounds for the interpolated magnetic field chosen so that the surface is
+# entirely contained in it
+n = 20
 rs = np.linalg.norm(s.gamma()[:, :, 0:2], axis=2)
 zs = s.gamma()[:, :, 2]
 rrange = (np.min(rs), np.max(rs), n)
 phirange = (0, 2*np.pi/3, n*2)
 zrange = (0, np.max(zs), n//2)
```
landreman, line 95: Shouldn't the interpolant cover negative z values too?

examples/1_Simple/tracing_fieldline.py
landreman: And should `endpoint=True` be used also in the analogous line of `tracing_particles.py`?

examples/1_Simple/tracing_fieldline.py
landreman: The same comment applies also to the analogous line of `tracing_particles.py`.

src/simsopt/field/magneticfieldclasses.py
florianwechsung: good point, done

examples/1_Simple/tracing_fieldline.py
florianwechsung: good catch, this was a left over from me debugging something. undid this.

examples/1_Simple/tracing_fieldline.py
florianwechsung: no, we're exploiting stellarator symmetry in the interpolant, so negative `z` values are mapped to `-z`

florianwechsung:
Thank you for your comments @landreman , I addressed them in the latest commit.

landreman (approved):

PR #230, opened 2022-05-05
Add a Weight object that encapsules a scalar

florianwechsung:
Often we use a weight in front of a penalty, then solve and optimization problem, and then we want to adjust the weight and solve the problem again.

With this PR one can now write

```
J1 = SomeObjective()
J2 = SomePenalty()
alpha = Weight(1e1)

J = J1 + alpha * J2

minimize(J,...)
alpha *= 100
minimize(J, ...)
```


landreman:
How about if at least one of the stage-2 optimization examples is tweaked to demonstrate this new `Weight` object?

mbkumar (dismissed):

mbkumar (approved):

PR #233, opened 2022-05-18
Tq/mgrid

tmqian:
Hello! I have written a module that writes mgrid files based on data provided by the `magneticfields` class. The write function `to_mgrid()` lives in `magneticfields.py`, but its implementation is in a separate module `fields/mgrid.py`. I've written a test script in `tests/fields/test_mgrid.py`. But it is not yet a "unit test" because it does not import the `unittest` module or use the same class structure that I find in other tests. I could use some guidance in how to implement that properly. A final resting place for the mgrid unit test may actually be inside the `test_magneticfields.py` file. I have put a dummy function with comments in line 680 of that file.


src/simsopt/field/mgrid.py
```diff
@@ -0,0 +1,229 @@
+import numpy as np
+#import netCDF4 as nc
+from scipy.io import netcdf as nc
+import sys
+
+### File I/O ###
```
mbkumar, line 6: Lines 6-11 could be deleted

src/simsopt/field/mgrid.py
```diff
@@ -0,0 +1,229 @@
...
+        self.zmax = zmax
+
+        self.n_ext_cur = 0
+        self.cur_labels = []
+
+        self.br_arr = [] 
+        self.bz_arr = [] 
+        self.bp_arr = [] 
+
+        print('Initialized mgrid file: (nr,nphi,nz,nfp) = ({}, {}, {}, {})'.format(nr, nphi, nz, nfp))
```
mbkumar, line 38: You can use formatted string such as `f"Initializized mgrid file: nr = {nr}"`, instead of the `format` string method.

src/simsopt/field/mgrid.py
```diff
@@ -0,0 +1,229 @@
...
+# updated 15 November 2021
+
+
+class MGRID():
+
+    def __init__(self, fname='temp', binary=False,\
+                 nr=51, nz=51, nphi=24, nfp=2,\
+                 rmin=0.20, rmax=0.40, zmin=-0.10, zmax=0.10,\
+                 nextcur=0):
+
```
mbkumar, line 20: Please add docstrings for the class and the methods!
Instead of fname use filename as argument name.

src/simsopt/field/mgrid.py
```diff
@@ -0,0 +1,229 @@
...
+    def export_phi(self):
+        phi = np.linspace(0, 2*np.pi/self.nfp, self.nphi)
+        return phi
+
+    def export_grid_spacing(self):
+        return self.nr, self.nz, self.nphi
+
+
+### for writing coil groups
```
mbkumar, line 128: For comments, a single # should suffice.

src/simsopt/field/mgrid.py
```diff
@@ -0,0 +1,229 @@
...
+
+        self.nextcur = int(f.variables['nextcur'].getValue())
+        coil_data = f.variables['coil_group'][:]
+        self.coil_names = [unpack(coil_data[j]) for j in range(self.nextcur)] 
+
+        self.rmin = f.variables['rmin'].getValue()
+        self.rmax = f.variables['rmax'].getValue()
+        self.zmin = f.variables['zmin'].getValue()
+        self.zmax = f.variables['zmax'].getValue()
+        self.nr = f.variables['ir']    .getValue()
```
mbkumar, line 164: Please delete the white space before .getValue()

src/simsopt/field/mgrid.py
```diff
@@ -0,0 +1,229 @@
...
+
+### File I/O ###
+# This class is prepared by Tony Qian (tqian@pppl.gov)
+# the original file included reading options for netCDF and binary
+# these have been removed and the code only writes netCDF
+#
+# updated 15 November 2021
+
+
+class MGRID():
```
mbkumar, line 14: Please use MGrid (camel case) instead of MGRID.

tests/field/test_mgrid.py
```diff
@@ -0,0 +1,65 @@
...
+import numpy as np
+from simsopt.field.biotsavart import BiotSavart
+from simsopt.field.coil import Current, coils_via_symmetries, ScaledCurrent
+from simsopt.geo.curve import curves_to_vtk
+from simsopt.geo.curvexyzfourier import CurveXYZFourier
+from simsopt.geo.surfacerzfourier import SurfaceRZFourier
+
+
+'''
+    This script tests the MGRID module in simsopt/field/magneticfield.py
```
mbkumar, line 13: Please remove the indent here.

tests/field/test_mgrid.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env python
+
```
mbkumar, line 2: This is better a as regression/integrated test. For unit tests, each method of the class needs to have a separate test.

mbkumar (changes requested):
Hello Tony,

Thank you for your contribution. I added some style and test-related comments. I'll take another look at the code after the comments are addressed.

Bharat

florianwechsung:
Thank you for getting started on this. Before we can merge this, a few things need to be addressed.

- Please use the conventions that are used throughout SIMSOPT in order to document classes or functions, instead of the comments that you currently use. See [here](https://github.com/hiddenSymmetries/simsopt/blob/105722d897a79869cf914b6748278c73a8b66d4d/src/simsopt/geo/curveperturbed.py#L10) for an example.
- There is no need for authorship lines as contained at the top of src/simsopt/field/mgrid.py. Git keeps track of this automatically, so there is less risk of dates and author lists getting out of date.
- We need a test and an example. Please see the existing tests to get started.

In my understanding, the main point of MGRID files is to then run free boundary VMEC. How are you doing that at the moment? Do you have further code that extends the existing vmec class in SIMSOPT? Or do you manually run free boundary vmec from the terminal?

src/simsopt/field/mgrid.py
tmqian: Thanks for the tip!

landreman:
This PR is ready for review. I added a unit test and example in which free-boundary vmec is run. Although @tmqian had been experiencing low-level netcdf issues, it appears to be working fine in the CI and on my macbook.

src/simsopt/field/mgrid.py
```diff
@@ -0,0 +1,290 @@
...
+
+
+def _unpack(binary_array):
+    '''
+    Decrypt binary char array into a string.
+    This function is used for reading coil group names.
+    '''
+    return "".join(np.char.decode(binary_array)).strip()
+
+
```
mbkumar, line 22: This class is a good candidate for the descriptor protocol. I'll push the changes

src/simsopt/field/mgrid.py
```diff
@@ -0,0 +1,290 @@
...
+        The field data is represented as a single "current group". For
+        free-boundary vmec, the "extcur" array should have a single nonzero
+        element, set to 1.0.
+
+        Args:
+            filename: output file name
+        '''
+
+        ds = netcdf_file(filename, 'w', mmap=False)
```
mbkumar, line 122: Context manager is helpful in case code crashes for some reason.

src/simsopt/field/mgrid.py
```diff
@@ -0,0 +1,290 @@
...
+    @classmethod 
+    def from_file(cls, filename):
+        '''
+        This method reads MGrid data from file.
+
+        Args:
+            filename: mgrid netCDF input file name
+        '''
+
+        f = netcdf_file(filename, 'r', mmap=False)
```
mbkumar, line 188: Again go for context manager

src/simsopt/field/mgrid.py
```diff
@@ -0,0 +1,290 @@
...
+        mgrid.br = br
+        mgrid.bp = bp
+        mgrid.bz = bz
+
+        mgrid.bvec = np.transpose([br, bp, bz])
+
+        return mgrid
+
+    def plot(self, jphi=0, bscale=0, show=True):
+        '''
```
mbkumar, line 253: If `show=False`, nothing useful happens. It should return a plot that could be saved.

tests/field/test_magneticfields.py
```diff
@@ -1005,6 +1007,45 @@ def test_to_vtk(self):
...
+        curves, currents, ma = get_ncsx_data()
+        nfp = 3
+        coils = coils_via_symmetries(curves, currents, nfp, True)
+        bs = BiotSavart(coils)
+        with tempfile.TemporaryDirectory() as tmpdir:
+            filename = Path(tmpdir) / "mgrid.bfield.nc"
+            bs.to_mgrid(filename, nfp=nfp)
+
+            # Compare the B data in the file to a separate evaluation here
+            f = netcdf_file(filename, mmap=False)
```
mbkumar, line 1020: Better to use context manager as a good coding style

tests/field/test_mgrid.py
```diff
@@ -0,0 +1,121 @@
...
+            f = netcdf_file(filename, mmap=False)
+            np.testing.assert_allclose(f.variables['zmin'][()], -0.5)
+            assert f.variables['nextcur'][()] == 1
+            assert f.variables['br_001'][:].shape == (6, 11, 11)
+            assert f.variables['mgrid_mode'][:][0].decode('ascii') == 'N'
+
+            byte_string = f.variables['coil_group'][:][0]
+            message = "".join([x.decode('ascii') for x in byte_string])
+            assert message == '________simsopt_coils_________'
```
mbkumar, line 66: Either use `f.close()` or the context manager (preferably the second option)

mbkumar (changes requested):
Hello Matt,

I have some minor comments. I'll add the descriptor protocol to the MGrid class.

landreman:
Hmm, some tests in the CI are failing intermittently, and from the error messages it looks like there is a memory leak in vmec, where it reads in the mgrid file during a free-boundary calculation:
~~~~
...
[fv-az461-187:10458] [ 3] /opt/hostedtoolcache/Python/3.9.17/x64/lib/python3.9/site-packages/vmec/_vmec.cpython-39-x86_64-linux-gnu.so(__mpi_sharmem_MOD_mpialloc_2d_dbl+0x88)[0x7f689147a738]
[fv-az461-187:10458] [ 4] /opt/hostedtoolcache/Python/3.9.17/x64/lib/python3.9/site-packages/vmec/_vmec.cpython-39-x86_64-linux-gnu.so(__mgrid_mod_MOD_read_mgrid+0x353d)[0x7f689146289d]
....
~~~~

mbkumar:
There is not much memory in the Github runner. Can you reduce the problem size for the free boundary calculation and retry?

src/simsopt/field/mgrid.py
landreman: Done.

src/simsopt/field/mgrid.py
landreman: Done.

tests/field/test_magneticfields.py
landreman: Done.

src/simsopt/field/mgrid.py
landreman: Done.

mbkumar (approved):

PR #241, opened 2022-06-25
Improvements to documentation

landreman:
This PR includes some improvements to the documentation:

- Fixed some links to the API docs that were not working in the docs after merging #239.
- Simplified more of the sub-submodule imports to submodule imports in the docs.
- In the coil optimization tutorial, the minimal example was simplified, and a new section was added to list the many available regularization terms.
- `stage_two_optimization.py` was split into a simpler minimal version and a version with all the regularization terms.

### DIFF
--- a/src/simsopt/field/__init__.py
+++ b/src/simsopt/field/__init__.py
@@ -1,8 +1,9 @@
 from .biotsavart import *
 from .coil import *
+from .magneticfield import *
 from .magneticfieldclasses import *
 from .boozermagneticfield import *
 from .tracing import *
 
-__all__ = (biotsavart.__all__ + coil.__all__ + magneticfieldclasses.__all__ +
+__all__ = (biotsavart.__all__ + coil.__all__ + magneticfield.__all__ + magneticfieldclasses.__all__ +
            boozermagneticfield.__all__ + tracing.__all__)
--- a/src/simsopt/geo/curve.py
+++ b/src/simsopt/geo/curve.py
@@ -13,7 +13,7 @@ from .._core.derivative import Derivative
 from .jit import jit
 from .plotting import fix_matplotlib_3d
 
-__all__ = ['RotatedCurve', 'curves_to_vtk', 'create_equally_spaced_curves']
+__all__ = ['Curve', 'RotatedCurve', 'curves_to_vtk', 'create_equally_spaced_curves']
 
 
 @jit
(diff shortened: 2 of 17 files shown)


src/simsopt/field/__init__.py
```diff
@@ -1,8 +1,9 @@
 from .biotsavart import *
 from .coil import *
+from .magneticfield import *
 from .magneticfieldclasses import *
```
mbkumar, line 4: I believe classes from magneticfield submodule need no exposure to public.

src/simsopt/geo/curve.py
```diff
@@ -13,7 +13,7 @@
 from .jit import jit
 from .plotting import fix_matplotlib_3d
 
-__all__ = ['RotatedCurve', 'curves_to_vtk', 'create_equally_spaced_curves']
+__all__ = ['Curve', 'RotatedCurve', 'curves_to_vtk', 'create_equally_spaced_curves']
```
mbkumar, line 16: Curve on its own is not usable. So public exposure is not needed.

src/simsopt/geo/surface.py
```diff
@@ -13,6 +13,8 @@
 from .._core.dev import SimsoptRequires
 from .plotting import fix_matplotlib_3d
 
+__all__ = ['Surface', 'signed_distance_from_surface', 'SurfaceClassifier', 'SurfaceScaled', 'best_nphi_over_ntheta']
+
```
mbkumar, line 17: Again my suggestion is not to expose Surface and SurfaceScaled to end users.

mbkumar (changes requested):
Hello Matt,

Thanks for the fixes to the docs. Some of the classes such as Curve, Surface, ScaledSurface need not be exposed to the public. We can revert some of the changes in this PR.

landreman:
I agree that the base classes `Curve`, `Surface`, and `MagneticField` aren't user-facing so we shouldn't need to import them into `simsopt.geo` and `simsopt.field` in principle. However, when I tried this approach, I experienced two problems. First, then `Curve`, `Surface`, and `MagneticField` don't appear anywhere in the API docs, not even in the more detailed API docs "for developers". This is a problem because then some important user-facing methods like `.plot()` and `.to_vtk()` aren't in the API docs anywhere. Second, if `Curve`, `Surface`, and `MagneticField` aren't imported into `simsopt.geo` and `simsopt.field`, then it doesn't seem possible to link to these classes in the `.rst` docs files. For instance, in this approach, then where `:obj:'simsopt.geo.curve.Curve'` appears in one of the `.rst` files, the hyperlink to the developer API docs does not work. Do you see some other way around these problems?

mbkumar:
Ok.  Makes sense. In that case your changes are correct.

mbkumar (approved):

PR #243, opened 2022-07-04
2X faster vmec_fieldlines

rahulgaur104:
Hi everyone,

In the `vmec_diagnostics.py` script, to find theta_vmec for a given theta_pest, one needs a non-linear root-finding algorithm. I have added a faster root finder to the vmec_fieldlines function. This function, written by Alan Goodman, speeds up the root-finding procedure by a factor of eight and the vmec_fieldlines function by at least a factor of two.


rahulgaur104:
I think a few of these tests compare the values of gbdrift and cvdrift with the geometry module in Stella which is why they fail.

src/simsopt/mhd/vmec_diagnostics.py
```diff
@@ -971,13 +979,13 @@ def vmec_fieldlines(vs, s, alpha, theta1d=None, phi1d=None, phi_center=0, plot=F
         for js in range(ns):
             theta_pest[js, :, :] = theta1d[None, :]
             phi[js, :, :] = phi_center + (theta1d[None, :] - alpha[:, None]) / iota[js]
-
+    
     def residual(theta_v, phi0, theta_p_target, jradius):
```
landreman, line 983: Because `fzero_residuals_function` is introduced, this earlier `residual` function can be removed.

src/simsopt/mhd/vmec_diagnostics.py
```diff
@@ -986,18 +994,147 @@ def residual(theta_v, phi0, theta_p_target, jradius):
...
+            theta_vmec_mins = theta_guess - 0.3*np.ones(theta_guess.shape)
+            theta_vmec_maxs = theta_guess + 0.3*np.ones(theta_guess.shape)
+            theta_test, converged, nconv = get_roots(theta_vmec_mins, theta_vmec_maxs, theta_guess, np.reshape(phi[js, jalpha, :], (-1, 1)), xm, xn, lmns[js], lmnc[js], 'False')
+            theta_vmec[js, jalpha] = np.reshape(theta_test, (-1, ))
+    
+            if converged == False:
+                print("* Error! Conversion from theta_pest to theta_vmec failed to converge")
+                print("===========================================================================")
+                print("")
+                err = True
```
landreman, line 1136: `err` is not used

src/simsopt/mhd/vmec_diagnostics.py
```diff
@@ -986,18 +994,147 @@ def residual(theta_v, phi0, theta_p_target, jradius):
         return theta_p_target - theta_p
         """
         return theta_p_target - (theta_v + np.sum(lmns[jradius, :] * np.sin(xm * theta_v - xn * phi0)))
-
-    # Solve for theta_vmec corresponding to theta_pest:
+    
+    # A helper function that returns 0 when theta_vmec_trys are close to the
+    # proper toroidal angle in magnetic coordinates
```
landreman, line 999: I don't think this description is accurate. The function returns 0 when `theta_vmec_trys` correspond to points along a field line, right?

src/simsopt/mhd/vmec_diagnostics.py
```diff
@@ -986,18 +994,147 @@ def residual(theta_v, phi0, theta_p_target, jradius):
...
+    #    for jalpha in range(nalpha):
+    #        for jl in range(nl):
+    #            theta_guess = theta_pest[js, jalpha, jl]
+    #            solution = root_scalar(residual,args=(phi[js, jalpha, jl], theta_pest[js, jalpha, jl], js),
+    #                                   bracket=(theta_guess - 1.0, theta_guess + 1.0))
+    #            theta_vmec[js, jalpha, jl] = solution.root
+    ## Solve for theta_vmec corresponding to theta_pest (new routine)
+    ## Does the same calculation as the commented code above but faster
     theta_vmec = np.zeros((ns, nalpha, nl))
     for js in range(ns):
```
landreman, line 1124: If you can get rid of the loop over l, can you get rid of the loops over s and alpha too?

landreman (changes requested):

landreman:
Also, `fzero_resisduals_function` could return derivatives with respect to `theta_vmec`, so `scipy.optimize.newton` could then use derivative-based root-finding. This would reduce the number of function evaluations, which would probably speed things up.

src/simsopt/mhd/vmec_diagnostics.py
```diff
@@ -986,18 +994,147 @@ def residual(theta_v, phi0, theta_p_target, jradius):
...
+    # vmec_fieldlines new residual routine (written by Alan Goodman)
+    def fzero_residuals_function(theta_vmec_trys, theta_pest_targets, phis, xm, xn, lmns, lmnc, lasym):
+        angle = xm * theta_vmec_trys - xn * phis
+        fzero_residual = theta_pest_targets - (theta_vmec_trys + np.reshape(np.sum(lmns * np.sin(angle), axis=1), (-1, 1)))
+        return fzero_residual
+    
+    # A helper function that numerically finds values that minimize fzero_residuals_function
+    # This is the non-linear root finder that turns out to be ~8X faster than scipy root_scalar
+    # written by Alan Goodman
+    def get_roots(a0, b0, theta_pest_targets, phis, xm, xn, lmns, lmnc, lasym):
```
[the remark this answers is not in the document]
rahulgaur104: I assumed that scipy.optimize.newton is slower than Alan's routine. Let me compare the speed of the two root finders.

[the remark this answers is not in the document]
rahulgaur104: I just tested the `scipy.optimize.newton` and it's slightly faster than Alan's routine. I thought he had compared his code with the existing root finders. Here are the profiling results:
```
 Time taken to specify all arrays = 0.0372 s
 Time taken by root_scalar = 0.5243 s
 Time taken by Alan's routine = 0.0649 s
 Time taken by vectorized Newton = 0.0467 s
 Difference in theta_arrays = 2.50E-11
 Time taken by Fourier sum = 0.1972 s
 Time taken by dual relation, gradient, and all the vector calculations = 0.0010 s
 Time taken by geometric coeff calculations = 0.0013 s
```
I think I should close this PR and create a new one with the vectorized Newton solver. It would be simpler to implement and test. 

I can also separate the sign issue from this PR to prevent any confusion.

src/simsopt/mhd/vmec_diagnostics.py
rahulgaur104: Yes, but the overall speedup will not be significant. For example, speeding up vectorized Newton by a factor of two would correspond to < 10% overall speedup.
```
Time taken to specify all arrays = 0.0372 s 
Time taken by vectorized Newton = 0.0467 s
Time taken by Fourier sum = 0.1972 s
```
Maybe for a future version.

src/simsopt/mhd/vmec_diagnostics.py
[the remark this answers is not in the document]
rahulgaur104: I have removed Alan's code with `scipy.optimize.newton`
I am marking all your comments as resolved now.

landreman (approved):

PR #244, opened 2022-07-04
PR: Make xsimd an optional dependency

rogeriojorge:
This pull request is born from an effort to install SIMSOPT in supercomputing clusters based on the Power9 IBM microprocessors. One of the challenges here is the fact that xsimd is a required dependency for SIMSOPT but it does not support PowerPC. In here, we make xsimd an optional dependency, therefore alleviating this constraint.


mbkumar:
Hello @rogeriojorge ,

Can I close this PR?

rogeriojorge:
Yes

PR #252, opened 2022-08-07
Avoid netcdf warning

landreman:
With this PR, `scipy.io.netcdf_file` is used instead of `scipy.io.netcdf.netcdf_file`, since the latter is being deprecated in `scipy`, so `scipy` has started printing a warning about it.


rogeriojorge:
Looks good to me.

rogeriojorge (approved):

PR #253, opened 2022-08-07
Fixed one vmec_fieldlines test that was failing

landreman:
On my new laptop one of the tests was failing since 0.0 != 1e-16. This is fixed in this pull request.


daringli (approved):
Minor improvement. Not much to say except that I approve.

PR #260, opened 2022-08-28
Surface classes refactoring to confirm with Graph JSON framework

mbkumar:
In this PR, I refactored the initialization of surface classes so that it works with the newly introduced graph JSON framework. Another advantage is a lot of if-else conditions tested in the initialization are also eliminated, leading to a more pythonic way of class inits for surface classes. The existing way of initializing surface classes is also preserved to a large extent. 

Bumped the python versions in the CI.

### DIFF
--- a/docs/source/geo.rst
+++ b/docs/source/geo.rst
@@ -106,18 +106,27 @@ to :obj:`~simsopt.geo.SurfaceRZFourier` when running the code.
 The points :math:`\phi_j` and :math:`\theta_j` are used for evaluating
 the position vector and its derivatives, for computing integrals, and
 for plotting, and there are several available methods to specify these
-points.  For :math:`\theta_j`, you typically specify a keyword
-argument ``ntheta`` to the constructor when instantiating a surface
-class. This results in a grid of ``ntheta`` uniformly spaced points
-between 0 and 1, with no endpoint at 1. Alternatively, you can specify
+points. For :math:`\theta_j`,  you can specify
 a list or array of points to the ``quadpoints_theta`` keyword argument
-when instantiating a surface class, specifying the :math:`\theta_j`
-directly.  If both ``ntheta`` and ``quadpoints_theta`` are specified,
-an exception will be raised.  For the :math:`\phi` coordinate, you
+when instantiating surface classes.
+Similarly, you can pass a list or array to the
+``quadpoints_phi`` keyword argument of the constructor for any Surface
+subclass to specify the :math:`\phi_j` points directly.  
+
+Often, it is desirable to specify the number of grid points instead of  
+supplying to the Surface classes. For such cases, use the class method
+provided :meth:`~simsopt.geo.surface.Surface.with_grid_range`. Replace
+``Surface`` class with the desired subclass when calling the ``with_grid_range`` 
+class method.
+For :math:`\theta_j`, you typically specify a keyword
+argument ``ntheta`` to the class method when instantiating a surface
+class. This results in a grid of ``ntheta`` uniformly spaced points
+between 0 and 1, with no endpoint at 1. 
+For the :math:`\phi` coordinate, you
 sometimes want points up to 1 (the full torus), sometimes up to
 :math:`1/n_{fp}` (one field period), and sometimes up to :math:`1/(2
 n_{fp})` (half a field period). These three cases can be selected by
-setting the ``range`` keyword argument of the surface subclasses to
+setting the ``range`` keyword argument of the class method to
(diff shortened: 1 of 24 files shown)


src/simsopt/geo/surface.py
```diff
@@ -1,4 +1,5 @@
 import abc
+from enum import Enum
```
landreman, line 2: I don't see `Enum` used anywhere.

landreman (commented):
All the other `@classmethod`s in simsopt are named `from_...()` instead of `with_...()`, so maybe it makes sense to do the same here for consistency. Some options: `from_range()`, `from_n()`, `from_npoints()`, `from_nquadpoints()`, `from_nphi_ntheta()`?

There are now a bunch of errors and warnings when building the docs. I think these are caused by the removed blank lines in `surface.py`. Make sure these errors and warnings are resolved.

docs/source/geo.rst
```diff
@@ -106,18 +106,27 @@ to :obj:`~simsopt.geo.SurfaceRZFourier` when running the code.
...
-when instantiating a surface class, specifying the :math:`\theta_j`
-directly.  If both ``ntheta`` and ``quadpoints_theta`` are specified,
-an exception will be raised.  For the :math:`\phi` coordinate, you
+when instantiating surface classes.
+Similarly, you can pass a list or array to the
+``quadpoints_phi`` keyword argument of the constructor for any Surface
+subclass to specify the :math:`\phi_j` points directly.  
+
+Often, it is desirable to specify the number of grid points instead of  
+supplying to the Surface classes. For such cases, use the class method
```
landreman, line 117: Missing words: "...instead of supplying quadpoints_phi and quadpoints_theta to the Surface..."

src/simsopt/geo/surface.py
```diff
@@ -32,17 +33,23 @@ class Surface(Optimizable):
...
     def __init__(self, **kwargs):
-        Optimizable.__init__(self, **kwargs)
+        super().__init__(**kwargs)
 
-    def get_quadpoints(quadpoints_phi=None,
-                       quadpoints_theta=None,
-                       range=RANGE_FULL_TORUS,
-                       nphi=None,
+    @classmethod
+    def with_grid_range(cls, nphi=61, ntheta=62, range="full torus", nfp=1,
```
landreman, line 39: docstring needed.

landreman (commented):
The docstrings for the `__init__` methods of each Surface subclass (`SurfaceRZFourier`, `SurfaceHenneberg`, etc) have text that needs to be updated for consistency. Specifically they presently say "For more information about the arguments nphi, ntheta, range, quadpoints_phi, and quadpoints_theta, see the general documentation on Surfaces." This should be updated to mention the new `@classmethod`. Also each surface class has `range`, `ntheta`, and `nphi` listed in the `Args:` section, which will no longer be correct.

src/simsopt/geo/surface.py
mbkumar: It is the remnant of playing with Enum class for Surface.RANGE_FULL_TORUS... Will delete it.

src/simsopt/geo/surface.py
```diff
@@ -61,50 +68,51 @@ def get_quadpoints(quadpoints_phi=None,
...
+            quadpoints_phi += 0.5 * dphi
 
-        return list(quadpoints_phi), list(quadpoints_theta)
+        return list(quadpoints_phi)
 
     def plot(self, engine="matplotlib", ax=None, show=True, close=False, axis_equal=True,
              plot_normal=False, plot_derivative=False, wireframe=True, **kwargs):
         """
-        Plot the surface in 3D using matplotlib/mayavi/plotly. 
-
```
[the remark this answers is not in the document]
mbkumar: Fixed the warnings generated by sphinx.

src/simsopt/geo/surface.py
mbkumar: Deleted the enum import

landreman (approved):
Looks ready to merge. One minor thing I noticed: for the `Surface` functions `get_quadpoints()`, `get_theta_quadpoints()`, and `get_phi_quadpoints()`, the first argument is missing in the function signature when the docs are rendered. (All the arguments do appear correctly in "Parameters:" list; the problem is only inside the function signature.) This is presumably because these functions do not take `self` as the first argument. I don't know if there is a way to fix this or not. Fixing that could be in a separate PR.

image

mbkumar:
We may have to use staticmethod decorator.  I'll fix it in a separate PR.

PR #265, opened 2022-09-21
Updated SurfaceRZFourier.from_vmec_input() to be faster.

landreman:
For high Fourier resolution, e.g. mpol=ntor=50, the function `SurfaceRZFourier.from_vmec_input()` can take several seconds. This small change makes it substantially faster.

### DIFF
--- a/src/simsopt/geo/surfacerzfourier.py
+++ b/src/simsopt/geo/surfacerzfourier.py
@@ -297,31 +297,34 @@ class SurfaceRZFourier(sopp.SurfaceRZFourier, Surface):
                    **kwargs)
 
         # Transfer boundary shape data from the namelist to the surface object:
+        # In these loops, we set surf.rc/zs rather than call surf.set_rc() for speed.
         for jm in range(rc.shape[0]):
             m = jm + nml.start_index['rbc'][1]
             for jn in range(rc.shape[1]):
                 n = jn + nml.start_index['rbc'][0]
-                surf.set_rc(m, n, rc[jm, jn])
+                surf.rc[m, n + ntor_boundary] = rc[jm, jn]
 
         for jm in range(zs.shape[0]):
             m = jm + nml.start_index['zbs'][1]
             for jn in range(zs.shape[1]):
                 n = jn + nml.start_index['zbs'][0]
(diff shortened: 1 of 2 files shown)


src/simsopt/geo/surfacerzfourier.py
```diff
@@ -297,31 +297,34 @@ def from_vmec_input(cls,
...
 
             for jm in range(zc.shape[0]):
                 m = jm + nml.start_index['zbc'][1]
                 for jn in range(zc.shape[1]):
                     n = jn + nml.start_index['zbc'][0]
-                    surf.set_zc(m, n, zc[jm, jn])
+                    surf.zc[m, n + ntor_boundary] = zc[jm, jn]
 
+        # Sync the dofs:
+        surf.local_full_x = surf.get_dofs()
```
andrewgiuliani, line 327: Looks good Matt -- although I find it unintuitive that `get_dofs()` must be called.  I have been tripped by this before and done something like `surf.zc[1]=5`, without the requisite followup `get_dofs` command and been confused why `surf.x` wouldn't update automatically.

It's probably not a high priority, but it might be nice to make surf.{zc,zs,rc,rs} properties, with their own setters that update `surf.local_full_x` simulatenously.  It seems like this can even be done with indexing

https://stackoverflow.com/questions/63214416/how-do-i-define-a-setter-for-a-list-with-an-index-or-slicing

andrewgiuliani (approved):
Looks good with a minor comment.

src/simsopt/geo/surfacerzfourier.py
andrewgiuliani: EDIT: seems like the linked solution on stackoverflow won't work here -- maybe @mbkumar has another idea.

landreman: Agreed, this is a fundamental issue with the present system in which the graph framework keeps a copy of dofs that are usually managed by Optimizable classes, so the 2 copies can get out of sync.

mbkumar:
Thanks Andrew for the stackoverflow link. I agree that the current method
is not intuitive. I will investigate the approach shown in the
stackoverflow.

Bharat

PR #270, opened 2022-10-13
Curves and Coils saved to FOCUS and MAKEGRID format

zhucaoxiang:
The `Curve` and `Current` objects can now be saved to the FOCUS format (assemble of Fourier coefficients) and the MAKEGRID format (discrete points). 

Two separate functions `curves_to_focus` and `coils_to_makegrid` are implemented. An example is as follows.
```python
from simsopt.field import coils_to_makegrid, coils_to_focus
from simsopt.configs import get_ncsx_data

nfp = 3
stellsym = True
curves, currents, ma = get_ncsx_data()

coils_to_focus('test.focus', curves, currents, nfp=3, stellsym=True)
coils_to_makegrid('coils.test', curves, currents, nfp=3, stellsym=True)
```

Some comments:
1. FOCUS can read both formats. Use `case_init=0` to read `test.focus` and `case_init = -1 ` to read `coils.test`.  
2. The current group in MAKEGIRD format can be improved if `coils_via_symmetries` not flip current but flip x,y,z order.
3. `Ifree` and `Lfree` in `curves_to_focus` can to use to flag whether the coil current or geometry is free.


landreman (commented):
Thanks for this contribution @zhucaoxiang. Can you add a unit test for each of the two functions? It could be sufficient to just write the files without checking the contents.

landreman:
Also I would think it makes sense to move `simsopt.geo.curves_to_focus()` to `simsopt.field.coils_to_focus()`, since this routine involves currents and not just curves. What do you think?

zhucaoxiang:
@landreman Thanks for the suggestion. I will move the function and add some unit test functions. I guess I will add to `tests/field/test_coil.py`?

florianwechsung:
Thank you for adding this Caoxiang. Looking at ˋcoils_to_focusˋ it seems that that function implicitly assumes that you are dealing with a ˋCurveXYZFourierˋ. If that’s the case, can you please add a type check?

zhucaoxiang:
@florianwechsung Thanks for the remind. @mbkumar also mentioned the `CurveRZFourier` class. Should I added an `if` statement to support other classes?

florianwechsung:
@zhucaoxiang Does FOCUS support the discretisation that we use in SurfaceRZFourier? I thought FOCUS only did Fourier expansion in XYZ? If FOCUS doesn't support it, I would just `assert isinstance(curve, CurveXYZFourier)` and let the user take care of converting between curve types. The reason being that I think the conversion is not lossless, so it should be up to the user to do it and decide how much accuracy is required.

zhucaoxiang:
@florianwechsung You are right. FOCUS or MAKEGRID use cartesian coordinates. So if users want to use these functions, they can convert to `CurveXYZFourier` first.

src/simsopt/field/coil.py
```diff
@@ -185,3 +186,108 @@ def coils_via_symmetries(curves, currents, nfp, stellsym):
...
+        filename: Name of the file to write.
+        curves: A python list of Curve objects.
+        currents: Coil current of each curve.
+        groups: Coil current group. Coils in the same group will be assembled together. Defaults to None.
+        nfp: The number of field periodicity. Defaults to 1.
+        stellsym: Whether or not following stellarator symmetry. Defaults to False.
+    """
+
+    assert len(curves) == len(currents)
+    assert isinstance(curves, CurveXYZFourier)
```
florianwechsung, line 207: I don't think we need this here. You're just exporting XYZ coordinates, that works for every choice of discretisation.

src/simsopt/field/coil.py
```diff
@@ -185,3 +186,108 @@ def coils_via_symmetries(curves, currents, nfp, stellsym):
...
+        currents: Coil current of each curve.
+        nfp: The number of field periodicity. Defaults to 1.      
+        stellsym: Whether or not following stellarator symmetry. Defaults to False.
+        Ifree: Flag specifying whether the coil current is free. Defaults to False.
+        Lfree: Flag specifying whether the coil geometry is free. Defaults to False.
+    """
+    from simsopt.geo import CurveLength
+
+    assert len(curves) == len(currents)
+    assert isinstance(curves, CurveXYZFourier)
```
florianwechsung, line 256: Think you should be asserting individually for each element in `curves` here, since `curves` is a list according to the docstring.

zhucaoxiang: You are right. The assertation has been moved inside the `for` loop.

zhucaoxiang:
The tutorial for free-boundary VMEC can be found at https://princetonuniversity.github.io/STELLOPT/VMEC%20Free%20Boundary%20Run. I just fixed the format. @rogeriojorge

zhucaoxiang:
Now, `nfp` is written into the file. I was wrong to fix `nfp=1`. The value will determine the `phi` range in the MGRID file. @rogeriojorge

rogeriojorge:
Thank you @zhucaoxiang 
Could an example on how to output the coils from a biot_savart.json SIMSOPT file look like this?
If it's ok, can I ask you to run it for the NCSX coils and benchmark it with the NCSX free boundary run that is on that tutorial? I still can't get VMEC to converge using this script. Perhaps I should change PHIEDGE?

`#!/usr/bin/env python3
import os
import numpy as np
from simsopt import load
from subprocess import run
from simsopt.mhd import Vmec
from simsopt.field.coil import coils_to_makegrid

biot_savart_file = "biot_savart.json"
mgrid_file = 'mgrid_opt_coils.nc'
coils_file = 'coils.opt_coils'
xgrid_input = 'input_xgrid.dat'
vmec_file = 'wout.nc'
xgrid_log = 'log_xgrid.opt_coils'
mgrid_executable = '~/bin/xgrid'
vmec_executable = '~/bin/xvmec2000'

vmec_final = Vmec(vmec_file)
nfp = vmec_final.indata.nfp
surface = vmec_final.boundary
bs = load(biot_savart_file)
ncoils = len(bs.coils)

r0 = np.sqrt(surface.gamma()[:, :, 0] ** 2 + surface.gamma()[:, :, 1] ** 2)
z0 = surface.gamma()[:, :, 2]
nzeta = 41
nr = 43
nz = 45
rmin=0.9*np.min(r0)
rmax=1.1*np.max(r0)
zmin=1.1*np.min(z0)
zmax=1.1*np.max(z0)

curves = [c.curve for c in bs.coils]
currents = [c.current for c in bs.coils]

coils_to_makegrid(coils_file, curves, currents, nfp=nfp, stellsym=True)

r0 = np.sqrt(surface.gamma()[:, :, 0] ** 2 + surface.gamma()[:, :, 1] ** 2)
z0 = surface.gamma()[:, :, 2]
with open(xgrid_input, 'w') as f:
    f.write('opt_coils\n')
    f.write('S\n')
    f.write('y\n')
    f.write(f'{rmin}\n')
    f.write(f'{rmax}\n')
    f.write(f'{zmin}\n')
    f.write(f'{zmax}\n')
    f.write(f'{nzeta}\n')
    f.write(f'{nr}\n')
    f.write(f'{nz}\n')

run_string = f"{mgrid_executable} < {xgrid_input} > {xgrid_log}"
run(run_string, shell=True, check=True)

vmec_final.indata.lfreeb = True
vmec_final.indata.mgrid_file = mgrid_file
vmec_final.indata.extcur[0:ncoils] = [c.current.get_value() for c in bs.coils]
vmec_final.indata.nvacskip = 6
vmec_final.indata.nzeta = nzeta
vmec_final.indata.phiedge = -np.abs(vmec_final.indata.phiedge)

vmec_final.indata.ns_array[:4]    = [   9,    29,    49,   101]
vmec_final.indata.niter_array[:4] = [4000,  6000,  6000,  8000]
vmec_final.indata.ftol_array[:4]  = [1e-5,  1e-6, 1e-12, 1e-15]
vmec_final.write_input(os.path.join(dir,'input.final_freeb'))
run_string = f"{vmec_executable} {os.path.join(dir,'input.final_freeb')}"
run(run_string, shell=True, check=True)`

zhucaoxiang:
@rogeriojorge I guess you need a new function to convert `biot-savart.json` to MAKEGRID. I have not used `biot-savart.json`. Can you send me an example file? (or show me where I can find the description & example file).

If it is smooth to convert `biot-savart.json` into `CurveFourierXYZ`, then we don't have to write a new one. As to your issue with NCSX, it is likely that the NCSX data used in SIMSOPT is not the same as in STELLOPT. SIMSOPT one is vacuum while STELLOPT one is finite-beta.

zhucaoxiang:
@rogeriojorge After reading the example, I don't think you need a new function. What you are doing above looks fine to me. The following line will save the coils into a MAKEGRID file.
```
curves = [c.curve for c in bs.coils]
currents = [c.current for c in bs.coils]

coils_to_makegrid(coils_file, curves, currents, nfp=nfp, stellsym=True)
```

I don't have the files you are using. Let me try to figure out the problem.
1. The box defined by [rmin, rmax, zmin, zmax] should be large enough to cover the plasma. I didn't get the lines you wrote.
```
nzeta = 41
nr = 43
nz = 45
rmin=0.9np.min(r0)
rmax=1.1np.max(r0)
zmin=1.1np.min(z0)
zmax=1.1np.max(z0)
```

2.  You should use the same `phiedge` for both fixed- and free-boundary VMEC.

rogeriojorge:
> 1. The box defined by [rmin, rmax, zmin, zmax] should be large enough to cover the plasma. I didn't get the lines you wrote.
> rmin=0.9np.min(r0)
> rmax=1.1np.max(r0)
> zmin=1.1np.min(z0)
> zmax=1.1np.max(z0)
I calculate the maximum and minimum R/Z of the plasma volume and use it (+10% tolerance) as the box size
> 2. You should use the same `phiedge` for both fixed- and free-boundary VMEC.
I agree.

zhucaoxiang:
Anything else I need to do with this PR? If someone thinks that we need a comprehensive example, I can try to work on it. But I need some help finding a coil set in the SIMSOPT format paired with a fixed-boundary VMEC equilibrium.

landreman:
I think the PR is ready to merge now. If anyone has issues or enhancements to the functions, they can be addressed later.

landreman (approved):

zhucaoxiang:
Great, Matt, @landreman thank you!

PR #274, opened 2022-11-04
Surface wrapper class with new quadrature points

mbkumar:
The surface wrapper class uses the DOFs in the parent surface class with the new quadrature points. Circumvents the need to have two surface classes share DOFs.


mbkumar:
Hello Andrew & Matt,

I need your feedback on two things before making the PR official.

1. Tests: Do I need more tests and if so, what should I test?
2. Derivatives: I haven't looked at derivatives. Should I modify the vjp methods as well?

mbkumar:
@andrewgiuliani.

This PR is ready for review

src/simsoptpp/surface_newquadpts.cpp
```diff
@@ -0,0 +1,90 @@
...
+//    we want to avoid lots of calls to sin and cos. Instead, we use the rules
+//          sin(a + b) = cos(b) sin(a) + cos(a) sin(b)
+//          cos(a + b) = cos(a) cos(b) - sin(a) sin(b)
+//    to write
+//          sin(m*theta-(n+1)*nfp*phi) = cos(-nfp*phi) * sin(m*theta-n*nfp*phi) + cos(m*theta-n*nfp*phi) * sin(-nfp*phi)
+//          cos(m*theta-(n+1)*nfp*phi) = cos(m*theta-n*nfp*phi) * cos(-nfp*phi) + sin(m*theta-n*nfp*phi) * sin(-nfp*phi)
+//    In our code we loop over n. So we start with n=-ntor, and then we always
+//    just increase the angle by -nfp*phi.
+
+#define ANGLE_RECOMPUTE 5
```
landreman, line 19: These lines are leftover from surfacerzfourier and can be removed, right?

tests/geo/test_surface_newquadpts.py
```diff
@@ -0,0 +1,81 @@
...
+        for surface_type in surface_types:
+            s = eval(surface_type + ".from_nphi_ntheta(ntheta=31, nphi=30, range='field period', nfp=1)")
+            s1 = eval(surface_type + ".from_nphi_ntheta(ntheta=62, nphi=61, range='full torus', nfp=1)")
+            quadpoints_phi, quadpoints_theta = Surface.get_quadpoints(
+                nphi=61, ntheta=62, range='full torus', nfp=1)
+            s_newquadpts = SurfaceNewQuadPoints(
+                s, quadpoints_phi=quadpoints_phi,
+                quadpoints_theta=quadpoints_theta)
+
+            self.assertTrue(np.allclose(s1.gamma(), s_newquadpts.gamma()))
```
landreman, line 54: Not at all critical, but I would think `np.testing.assert_allclose()` is preferable to `self.assertTrue(np.allclose())` so the error messages when a test fails are more informative.

landreman (commented):
Looks good to me.

tests/geo/test_surface_newquadpts.py
mbkumar: Made the suggested change

src/simsoptpp/surface_newquadpts.cpp
mbkumar: Yes. Removed them.

mbkumar:
Hello Florian,

This is the easier route to accomplish the same objective. When DOFS are
shared across different Optimizable objects,  the tracking of dependencies
gets complicated. We can have a discussion when you are available.

Bharat

florianwechsung:
I see now. Are you planning to do the same thing for curves?

Can you outline what the problem is with the approach that we had discussed? It seems much more elegant to just modify the DOFs and Optimizable class, instead of every single Curve, Surface, etc object for which we may want this sort of DOF sharing.

And yes, the `vjp` methods need modifying as well.

andrewgiuliani:
Sharing `DOFs` between surfaces seems like a cleaner solution that does not require the creation of new wrapper classes like this one.  Also it avoids having to modify the `vjp` methods and create wrapper classes for `Curve` objects too...

@mbkumar can you give a few more details why the bookkeeping of dependencies becomes too complicated?

mbkumar:
OK. I was thinking about how to do it over the break. I think I have the
implementation mapped out in my brain. Let me work on it this week.

Bharat Medasani

Engineer

mbkumar:
Closing this PR because the issue was addressed in #276.

PR #276, opened 2022-11-29
Shared DOFs 

mbkumar:
I implemented a DOFs sharing mechanism between different Optimizable subclasses. To get it to work, I had to call a couple of Optimizable methods from DOFs class. Since this is a circular dependency between the two classes, I was resistant to think about it, but this is the only simple solution that I can think of. 

Please give me your thoughts. If you are happy with this, we can discard PR #274.

Edit1: Updated `Surface` and `Curve` subclasses for sharing DOFs. JSON serialization is also updated to work with shared DOFs. Bugs in deserialization in many of the from_dict methods are eliminated. 
Edit2: Updated global dof vector aggregation for shared DOF cases.


landreman:
For the new tests in `TestOptimizableSharedDOFs`, it would be useful to have similar tests in which dofs are shared between two surfaces, and between two curves. Those cases are a bit more complicated than `OptClassSharedDOFs` because in surfaces and curves, the dofs are duplicated in C++ and python, and everything needs to stay in sync.

andrewgiuliani:
does serialization still work when two Optimizables share dofs? if yes, then this should be tested.  If not, is it possible to add this functionality?

mbkumar:
> For the new tests in `TestOptimizableSharedDOFs`, it would be useful to have similar tests in which dofs are shared between two surfaces, and between two curves. Those cases are a bit more complicated than `OptClassSharedDOFs` because in surfaces and curves, the dofs are duplicated in C++ and python, and everything needs to stay in sync.

Yes. Having such tests helps. We may have to modify the API of those classes to make them share DOFs.

@andrewgiuliani, Do you want me to work on it.

mbkumar:
> does serialization still work when two Optimizables share dofs? if yes, then this should be tested. If not, is it possible to add this functionality?

That's a good question. I think we will end up distinct DOFs after deserialization. I have to extend the serialization framework to DOFs class as well. Up to now, only Optimizable class is serialized and the DOFs class is not considered in serialization.

mbkumar:
> For the new tests in `TestOptimizableSharedDOFs`, it would be useful to have similar tests in which dofs are shared between two surfaces, and between two curves. Those cases are a bit more complicated than `OptClassSharedDOFs` because in surfaces and curves, the dofs are duplicated in C++ and python, and everything needs to stay in sync.

On this point, if the classes maintain their own DOFs on C++ side, their C++ APIs need to be modified to accept external DOFs. Currently, the simsopt keeps in sync the python and C++ DOFs for the same object.

mbkumar:
Updated the API's of `Optimizable`, `Surface` and `Curve` classes. The new API makes it feasible to share DOFs for the `Surface` and `Curve` subclasses on the python side. One side benefit is reimplementation of `from_dict` classmethod is not required anymore for the `Curve` and `Surface` subclasses.

Tested the serialization of shared DOFs for `SurfaceRZFourier`. If wanted, I can write similar tests for other `Surface` subclasses as well as `Curve` subclasses.

@andrewgiuliani, Could you please explore the idea of having "points" argument in the `__init__` for MagnetificField classes?

src/simsopt/_core/optimizable.py
```diff
@@ -119,9 +119,32 @@ def __init__(self,
...
+        self._upper_bounds = upper_bounds
         self._names = list(names)
+        self._dep_opts = []
+        self._hash = id(self) % 10**32
+        self.name = str(id(self))   # For serialization
+
+    def __hash__(self):
+        return self._hash
+
+    def add_opt(self, opt):
```
andrewgiuliani, line 132: @mbkumar could you add documentation for these functions?

tests/core/test_optimizable.py
```diff
@@ -1276,5 +1276,90 @@ def test_load_save(self):
...
+
+class TestOptimizableSharedDOFs(unittest.TestCase):
+    """
+    Test the DOFs sharing for the Optimizable classes
+    """
+
+    def test_adder_dofs_shared_change_vals(self):
+        adder_orig = OptClassSharedDOFs(x0=[1, 2, 3], names=["x", "y", "z"],
+                                        fixed=[False, False, True])
+        adder_shared_dofs = OptClassSharedDOFs(dofs=adder_orig._dofs)
```
andrewgiuliani, line 1295: is accessing `adder_orig._dofs` directly recommended here? is the convention that variables starting with underscore are "internal"?

src/simsopt/_core/optimizable.py
```diff
@@ -680,7 +723,11 @@ def get_return_fn_list(self) -> List[List[Callable]]:
         Optimizable objects
 
         Returns:
-            List of methods that return a value when the current Optimizable
+            List of methods that return a value w            x0: Initial state (or initial values of DOFs)
```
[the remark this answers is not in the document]
mbkumar: Yes. Thanks for spotting it.

tests/core/test_optimizable.py
mbkumar: I updated the code to remove the underscore. The DOFs object is made available because we are making use of it outside the parent Optimizable object.

tests/core/test_optimizable.py
```diff
@@ -1324,6 +1328,13 @@ def test_adder_dofs_shared_fix_unfix(self):
 
         adder_orig.fix_all()
         self.assertTrue(len(adder_shared_dofs.x) == 0)
+    
+    def test_derivative(self):
+        adder_orig = OptClassSharedDOFs(x0=[1, 2, 3], names=["x", "y", "z"],
+                                        fixed=[False, False, True])
+        adder_shared_dofs = OptClassSharedDOFs(dofs=adder_orig.dofs)
+        sum_obj = adder_orig + adder_shared_dofs
+        self.assertEqual(adder_orig.dJ()*2, sum_obj.dJ())
```
andrewgiuliani, line 1337: @mbkumar the `Optimizable` framework is not combining derivatives correctly for `Optimizable`s that share `DOF`s, see above example.

tests/geo/test_surface_rzfourier.py
```diff
@@ -54,6 +55,24 @@ def test_init(self):
...
+        s.rc[1, 0] = 0.4
+        s.zs[1, 0] = 0.2
+        s.local_full_x = s.get_dofs()
+
+        quadpoints_phi, quadpoints_theta = Surface.get_quadpoints(
+            ntheta=31, nphi=30, range='field period')
+        s2 = SurfaceRZFourier(quadpoints_phi=quadpoints_phi,
+                              quadpoints_theta=quadpoints_theta,
+                              dofs=s.dofs)
+        self.assertIs(s.dofs, s2.dofs)
```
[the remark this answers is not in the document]
andrewgiuliani: maybe we could use a different naming than `dofs`, such as `dofs_obj`.  The difference between `s.x`, `s.dofs`, and `s.get_dofs()` would be/is confusing.

tests/core/test_optimizable.py
mbkumar: This is not the deficiency of the Optimizable framework. The Derivative class uses the Optimizable objects as keys. Since each instance is a different object, the keys are different. So Derivative class is giving a 4 element vector for sum_obj.dJ(). 

IMO, the Derivative class needs to be updated to account for DOF sharing.

andrewgiuliani: I have attempted a fix in the `__call__` dunder

mbkumar: @andrewgiuliani, @landreman, @florianwechsung 

I think this issue is more general than just derivatives. What if the final objective contains terms from two or more Optimizable objects that share a single DOFs object. The global state vector contains duplicates of the dofs from the shared DOFs object. We need to have a more rigorous framework for assembling the global state vector.
My question is will this situation arise or is the sharing of DOFs limited to simple cases?

mbkumar: I updated the global dof vector assembly to account for shared DOFs.

andrewgiuliani:
@mbkumar here's another failing unit test.  Since the `Derivative` dictionary is keyed by `Optimizable`, 

`sum_obj.dJ(partials=True)(adder_orig) `

does not return the proper partial derivative.

mbkumar:
> @mbkumar here's another failing unit test. Since the `Derivative` dictionary is keyed by `Optimizable`,
> 
> `sum_obj.dJ(partials=True)(adder_orig) `
> 
> does not return the proper partial derivative.

@andrewgiuliani, 
The Derivative class implementation still assumes that Optimizable objects do not share DOFs. Its implementation needs to be updated to account for the shared DOFs. The key type may have to be changed from Optimizable to DOFs in the internal dictionary and use a map of the Optimizable -> DOFs for user-facing functions. I am not sure how to keep track of the graph dependencies if we make that switch. Let me think about this a bit more.

mbkumar:
@andrewgiuliani 
The Derivative class now passes all the tests. However, CI is failing for unknown reasons on Github, but passing on my machine.

mbkumar:
> CI is failing for unknown reasons on Github, but passing on my machine.

I fixed the JSON related issue.

andrewgiuliani:
> @andrewgiuliani Could you please take a look at the failing test in Extensive CI #1089? I ran it on my machine and the failing test generated a Deprecated warning. The offending part is Line 488 in magneticfieldclasses.py

sure I'll take a look

src/simsopt/_core/optimizable.py
```diff
@@ -928,30 +1012,27 @@ def full_x(self) -> RealArray:
...
-        self._set_new_x()
-
-    def _set_local_x(self, x: RealArray) -> None:
         if self.local_dof_size != len(x):
             raise ValueError
-        self._dofs.x = x
+        self._dofs.free_x = x
         if self.local_dof_setter is not None:
             self.local_dof_setter(self, list(self.local_full_x))
+        self.set_recompute_flag()
```
andrewgiuliani, line 1035: @mbkumar I've pushed some more failing unit tests.  Seems to me that we need to set the recompute flag on `self` but also on the `Optimizables` that share `dofs` with `self`

src/simsopt/_core/optimizable.py
```diff
@@ -1030,8 +1030,11 @@ def local_x(self, x: RealArray) -> None:
         if self.local_dof_size != len(x):
             raise ValueError
         self._dofs.free_x = x
-        if self.local_dof_setter is not None:
-            self.local_dof_setter(self, list(self.local_full_x))
+
```
andrewgiuliani, line 1033: @mbkumar we forgot to update the dofs on the cpp side, so here's an attempted fix

mbkumar:
I fixed the issue with eliminate_magnetic_islands.py. The problem is that one extra function has to be called in the SurfaceRZFourier resolution-changing method. I added a new function `replace_dofs` Optimizable to replace the DOFs object properly and used that in SurfaceRZFourier class.

I also updated a few functions in DOFs and Optimizable classes to account for shared DOFs.

The code is ready to be reviewed and merged.

src/simsopt/geo/surfaceobjectives.py
```diff
@@ -16,7 +16,18 @@ class Area(Optimizable):
...
+        if range is None:
+            if in_surface.stellsym:
+                range = Surface.RANGE_HALF_PERIOD
+            else:
+                range = Surface.RANGE_FIELD_PERIOD
+        if nphi is None:
+            nphi = len(in_surface.quadpoints_phi)
+        if ntheta is None:
+            ntheta = len(in_surface.quadpoints_theta)
+        surface = in_surface.__class__.from_nphi_ntheta(nphi=nphi, ntheta=ntheta, range=range, nfp=in_surface.nfp, stellsym=in_surface.stellsym, \
```
florianwechsung, line 29: Do we really need to force the creation of a new surface object here? It seems that this is only necessary when 
`range is not None or nphi is not None or ntheta is not None`. Same for the objectives below.

florianwechsung:
Thank you @andrewgiuliani and @mbkumar . This looks good to me, I would just like to raise one think.

In the derivative object, we use `Optimizable` objects as keys. In particular, we use those optimizables that are 'highest' in the dependency chain. In the past, dofs were tied uniquely to an optimizable object, so that made intuitive sense (to me). However, now we may have two optimizables using the same DOFs, but the derivative object would have one entry associated with each optimizable. It seems more canonical to me to use the DOFs keys, as these are ultimatively what we want to take the derivative with respect to. What do you think about this?

mbkumar:
> 
> In the derivative object, we use `Optimizable` objects as keys. In particular, we use those optimizables that are 'highest' in the dependency chain. In the past, dofs were tied uniquely to an optimizable object, so that made intuitive sense (to me). However, now we may have two optimizables using the same DOFs, but the derivative object would have one entry associated with each optimizable. It seems more canonical to me to use the DOFs keys, as these are ultimatively what we want to take the derivative with respect to. What do you think about this?

Hello Florian,

Thank you for your comments. Yes, I agree with your observation about the Derivative class. Doing so, however, would take some time. I am thinking of implementing the changes in a separate PR. Can the other two PRs (#229 and #258) move forward if this PR is merged?

landreman (commented):
There are several attributes that are renamed to include "free": `free_upper_bounds`, `free_lower_bounds`, `free_names`, `free_x`. This breaks the pattern in https://simsopt.readthedocs.io/en/latest/optimizable.html#function-reference where the prefix `full_` means "both fixed and free", otherwise there is no prefix and "free" is implied. Is there a way to keep a consistent naming pattern?

mbkumar:
> There are several attributes that are renamed to include "free": `free_upper_bounds`, `free_lower_bounds`, `free_names`, `free_x`. This breaks the pattern in https://simsopt.readthedocs.io/en/latest/optimizable.html#function-reference where the prefix `full_` means "both fixed and free", otherwise there is no prefix and "free" is implied. Is there a way to keep a consistent naming pattern?

The renaming happened within the `DOFs` class, which is not fully user exposed in the sense that users are not expected to call any of the methods within the DOFs except for class initialization in rare cases. So we have a lee-way to rename the methods, properties, or attributes. Now the reason for renaming is to get the serialization of the `DOFs` class working. Serialization depends on having internal attributes that match the arguments passed to `DOFs.__init__`. Earlier corresponding names were used with implied free status. But for serialization, we have to store the full arrays and not just free degrees of freedom. So after renaming, standalone names mean full arrays. Arrays masked with "free" status are prepended with `free_`.

florianwechsung:
> Thank you for your comments. Yes, I agree with your observation about the Derivative class. Doing so, however, would take some time. I am thinking of implementing the changes in a separate PR. Can the other two PRs (#229 and #258) move forward if this PR is merged?

I think so, yes. Can you address https://github.com/hiddenSymmetries/simsopt/pull/276#discussion_r1063031216 before you merge though?

src/simsopt/geo/surfaceobjectives.py
andrewgiuliani: @florianwechsung done
@landreman this was a simple change to test out the new shared dofs functionality with the geometry objects

andrewgiuliani:
> I think so, yes. Can you address [#276 (comment)](https://github.com/hiddenSymmetries/simsopt/pull/276#discussion_r1063031216) before you merge though?

done

src/simsopt/geo/surfaceobjectives.py
```diff
@@ -104,17 +114,25 @@ class ToroidalFlux(Optimizable):
...
-        surface = in_surface.__class__.from_nphi_ntheta(nphi=nphi, ntheta=ntheta, range=range, nfp=in_surface.nfp, stellsym=in_surface.stellsym,\
-                                                        mpol=in_surface.mpol, ntor=in_surface.ntor, dofs=in_surface.dofs)
-        self.surface = surface
+
+        if range is not None or nphi is not None or ntheta is not None:
+            surface = in_surface.__class__.from_nphi_ntheta(nphi=nphi, ntheta=ntheta, range=range, nfp=in_surface.nfp, stellsym=in_surface.stellsym, \
+                                                            mpol=in_surface.mpol, ntor=in_surface.ntor, dofs=in_surface.dofs)
+            self.surface = surface
+        else:
+            self.surface = in_surface
```
landreman, line 131: In the 3 surface objectives, I don't think this logic makes sense. `range`, `nphi`, and `ntheta` are all ensured to be different from `None` before line 125. So lines 130-131 never execute, right? Do you want to move lines 116-124 inside the `if` statement on line 126? (Same for `Volume` and `Area`.)

PR #280, opened 2022-12-19
Enabling vmec input writing if keep_all_files=True.

ejpaul:
This PR changes a single line in vmec.py to enable the vmec input files to be written if keep_all_files = True.


landreman:
@mbkumar: The tests are failing with that error "Object of type Array is not JSON serializable" which I think you fixed in #276. To add this fix to #280 also, are you able to make that fix in master and then merge master into #280?

andrewgiuliani:
@landreman there was another failing test in Extensive CI #1089, which I have fixed in the most recent commit.  However, this doesn't address your other concern.

landreman (approved):

PR #282, opened 2023-01-10
Linking number

AlexWiedman:
Added a function to calculate a linking number for curves. This is helpful for coil optimizations to make sure the curve objects are not linked. Does not support JaxCurveXYZFourier.


src/simsoptpp/python.cpp
```diff
@@ -1,12 +1,14 @@
 #include "pybind11/pybind11.h"
 #include "pybind11/stl.h"
 #include "pybind11/functional.h"
+#include <iostream>
```
landreman, line 4: This change can be reverted.

src/simsoptpp/python.cpp
```diff
@@ -1,12 +1,14 @@
...
 #define FORCE_IMPORT_ARRAY
 #include "xtensor-python/pyarray.hpp"     // Numpy bindings
 #include <Eigen/Core>
 typedef xt::pyarray<double> PyArray;
 #include <math.h>
 #include <chrono>
+#include <stdio.h>
```
landreman, line 11: This change can be reverted.

tests/geo/test_curve_objectives.py
```diff
@@ -358,6 +358,40 @@ def test_curve_surface_distance(self):
             assert err_new < 0.3 * err
             err = err_new
 
+    def test_linking_number(self):
+
+        curves1 = create_equally_spaced_curves(2, 1, stellsym=True, R0=1, R1=0.5, order=5, numquadpoints=128)
+
+        from simsopt.geo.curvexyzfourier import CurveXYZFourier
```
landreman, line 365: It's best to keep imports at the top of the file, and I think this class is imported there already anyway.

src/simsopt/geo/curveobjectives.py
```diff
@@ -476,3 +476,53 @@ def dJ(self):
...
+    def __init__(self, curves):
+        Optimizable.__init__(self, depends_on=curves)
+        self.curves = curves
+        r"""
+        Compute the Linking number of a set of curves (whether the curves 
+        are interlocked or not).
+
+        The value is 1 if the are interlocked, 0 if not.
+        
+        .. math::
```
landreman, line 492: There are some issues with the docstring math in the rendered version. To view the rendered docs, go to the `docs` directory and run `make html`. Then open `docs/build/html/index.html` in a browser. I think `\cross` should be `\times`, and the quotes after `:math:` should be slanted (the key next to !).

landreman (commented):
Users should be able to import `LinkingNumber` directly from `simsopt.geo`. To do this, we should add `LinkingNumber` to `__all__` at the top of `curveobjectives.py`.

PR #284, opened 2023-01-13
Updated workflows

mbkumar:
Mac wheel building is updated to use `cibuildwheel`. `cibuildwheel` version is upgraded to the latest. Support for arm64 arch added for macOS.


landreman:
Does it make sense to also add wheels for macos-12?

mbkumar:
The wheels are the same for macos11 and macos12.

Bharat Medasani

Engineer

landreman (approved):

PR #285, opened 2023-01-27
Adds netcdf4 support

daringli:
scipy.io.netcdf_file cannot read netcdf4 files. It may thus be desirable to allow different netcdf libraries to be used.
This branch adds support for netCDF4 if the user has it installed, using the existing scipy.io.netcdf_file import as a fall-back.
This is achieved using a thin abstraction layer between simsopt and the netcdf library. Since scipy.io netcdf_file implements a subset of the methods of netCDF4's Dataset object, simsopt is already compatible with both, and no wrapper class is needed. The abstract class in util/netcdf.py merely documents this shared interface.

This may solve the problems Tony had with not being able to read certain mgrid files.
As further motivation, gyrokinetic code stella now only supports writing to netcdf4 files, since it uses neasy-f for its netcdf output (https://github.com/PlasmaFAIR/neasy-f). If neasy-f becomes widespread, equilibrium codes may follow suit, and wout files in netcdf4 format could start appearing in the wild.

### DIFF
--- a/src/simsopt/util/netcdf.py
+++ b/src/simsopt/util/netcdf.py
@@ -17,6 +17,8 @@ class NetCDF(Enum):
     netCDF4 = 0
     scipy = 1
 
+logger = logging.getLogger(__name__)
+    
 try:
     from netCDF4 import Dataset
 except ImportError as e:
--- a/src/simsopt/util/netcdf.py
+++ b/src/simsopt/util/netcdf.py
@@ -10,7 +10,6 @@ _all__ = ['netcdf_file']
 
 import logging
 from enum import Enum
-from abc import ABC, abstractmethod
 
 
 class NetCDF(Enum):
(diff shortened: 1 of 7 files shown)


src/simsopt/util/netcdf.py
```diff
@@ -0,0 +1,90 @@
...
+        from scipy.io import netcdf_file as scipy_netcdf_file
+    except ImportError as e:
+        # this error is fatal
+        logger.debug(str(e))
+        raise e
+    else:
+        netcdf = NetCDF.scipy
+else:
+    netcdf = NetCDF.netCDF4
+
```
mbkumar, line 38: We are not using `AbstractNetCDF` to define any concrete classes. It is better to put the class definition in the docstrings of `netcdf_file` function, so users can know how to use the returned netcdf object.

daringli:
Thank you for reviewing this request.
I have addressed your other comments, although given that vmec doesn't use netcdf4, I think the value of adding an additional (although optional) dependency might be slim, compared to the burden of setting up additional testing environments for the CI.
I also ran into problems related to  simsopt/src/simsopt/util/__init__.py importing MPI, both through .logger and .mpi. As a result, more components of simsopt were not runnable on the login nodes of MPG cobra. I have fixed the issue, but the fix potentially breaks backwards compatibility for users that import * from simsopt.util.

Until a clear use-case emerges, I think the downsides of this pull request outweigh the benefits.

mbkumar:
@daringli

I am planning to close this PR. If you still want the changes to be merged, please let me know.

daringli:
Please feel free to close it.

> @daringli
> 
> I am planning to close this PR. If you still want the changes to be merged, please let me know.

PR #289, opened 2023-02-16
Jb/curvecws

joaopedrobiu6:
This PR enables simsopt to create coil curves on a specified coil winding surface. For this purpose the following changes are implemented.

1. New class CurveCWS allows to create a coil curve given the dofs of a surface.
2. CurveCWS inherits from the Curve class.
3. At this moment the only supported surface for the coil curves are of the SurfaceRZFourier type.

To finalize this PR:
1. dgamma_by_dcoeff_impl needs to be implemented.
2. Need to add unit tests.

(João Biu, student of Rogério Jorge)


landreman:
Thanks for working on this feature!

It looks like the present implementation is hard-wired to use a winding surface represented like in `SurfaceRZFourier`, but duplicating some of the machinery for computing the position vector on the surface instead of using the functions like gamma already in `SurfaceRZFourier`. Do avoid code duplication and make the new class applicable to any surface type, how about instead having `CurveCWS` point to a general `Surface` object. The position vector along the curve could be computed by first having `CurveCWS` evaluate the map from `t` to `(theta, phi)`, then calling the surface's functions on the resulting points. I believe one could potentially use the `gamma_lin()` function for `Surfaces`, which allows you to evaluate the position vector on sets of (theta, phi) points that are not tensor product grids. For derivatives of the position vector, this might require adding some other functions to Surfaces akin to `gamma_lin()`. What do you think?

In the future there might be other classes like `CurveCWS` that use a spline instead of Fourier representation for the map from `t` to `(theta, phi)`. Therefore it might make sense to add `Fourier` to the name of the class, e.g. `CurveCWSFourier`.

PR #290, opened 2023-02-23
Bug fix for padding in RegularGridInterpolant3D

mbkumar:
There is a bug in simd padding in RegularGridInterpolant3D. When the size is the multiple of simdcount, spurious padding bytes are added.


mbkumar:
Minor bug fix in padding for simd code

landreman (approved):
Good catch.

PR #293, opened 2023-03-02
Minor fixes in some tests

mbkumar:
This PR adds some minor fixes to tests to make them optional and definitive.


landreman:
Should we add `@unittest.skipIf(virtual_casing is None)` at [test_virtual_casing line 56](https://github.com/hiddenSymmetries/simsopt/blob/master/tests/mhd/test_virtual_casing.py#L56) too? I believe the tests in that file are divided in 2 classes because only `test_different_initializations()` needs to actually have vmec installed, but all the tests in the `test_virtual_casing` file need to have the `virtual_casing` module installed.

mbkumar:
Yes.  Good catch.

mbkumar:
Updated mhd/test_virtual_casing.py to correctly capture the skip conditions.

landreman (approved):

PR #300, opened 2023-04-16
Bug in coil examples and docs: missing "max" argument

landreman:
Yikes, discussing with Antoine Cerfon just now, we realized there was an important error with all the simsopt stage-2 coil optimization examples and documentation. All the `QuadraticPenalty` terms were missing the `"max"` argument. The default for `QuadraticPenalty` was `"min"`, meaning that coil length and mean-squared-curvature were penalized only if they were below the target rather than above. I don't know how this error made it in. This problem is fixed in the examples and tutorial in this PR.

In this PR the default for `QuadraticPenalty` is also changed from `"min"` to `"identity"`. We only ever use `"max"` or `"identity"` in the examples, never `"min"`. Also `"identity"` seems like a safer and more intuitive default.

### DIFF
--- a/docs/source/example_coils.rst
+++ b/docs/source/example_coils.rst
@@ -214,13 +214,19 @@ used and the corresponding terms::
   Jls = [CurveLength(c) for c in base_curves]
 
   # Form the total objective function.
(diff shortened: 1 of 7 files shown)


rogeriojorge:
The changes seem good to me.

PR #304, opened 2023-04-20
Spec dofs

abaillod:
I finally found the time to document and improve my code. In this branch I implemented some additional features for optimizing SPEC equilibria:

- Can now use any input profile as dofs, including the toroidal flux profile (i.e. the position of SPEC interfaces). See for example the test in `tests/mhd/spec_test/SpecTests.test_optimize_net_toroidal_current()`
- Can now use the Vns (i.e. B.n on the computational boundary) as dofs, to perform quasi-free-boundary optimization. These harmonics are stored in a `NormalField` object, implemented in `src/simsopt/field/normal_field.py`
- `Spec` now stores the latest converged equilibrium, and uses it as initial guess for the next iteration. Interfaces are stored as `SurfaceRZFourier` objects.

These new features have been used to obtain the results presented in [this publication](https://doi.org/10.1063/5.0080809). I am still working on the coupling of SPEC with `booz_xform` and the Redl formulas for the computation of the bootstrap current. These changes will come in a future PR.

Please let me know if you have any question or if anything needs to be improved!


src/simsopt/mhd/spec.py
```diff
@@ -192,6 +722,12 @@ def set_dofs(self, x):
         self.inputlist.phiedge = x[0]
         self.inputlist.curtor = x[1]
 
+        # Update phiedge in profiles if necessary. This is important when taking
+        # profiles derivatives.
+        for ii, p in enumerate(self.profiles):
+            if not p is None:
+                self.profiles[ii].phiedge = x[0]
```
landreman, line 729: See the issue around line 272 - I think `self.profiles` becomes out of sync if any profiles are set to new profiles. I think the following change here would resolve the issue:
~~~~py
        profiles = [
            self._volume_current_profile,
            self._interface_current_profile,
            self._pressure_profile,
            self._iota_profile,
            self._oita_profile,
            self._mu_profile,
            self._pflux_profile,
            self._tflux_profile,
            self._helicity_profile
	]
        for p in profiles:
            if p is not None:
                p.phiedge = x[0]
~~~~

src/simsopt/mhd/spec.py
```diff
@@ -303,28 +974,74 @@ def run(self):
...
+        # Deal with unconverged equilibria - these are excluded by 
+        # the optimizer, and the objective function is set to a large number
+        if self.results.output.ForceErr > self.tolerance:
+            raise ObjectiveFailure(
+                'SPEC could not find force balance'
+            )
+
+        # Save geometry as initial guess for next iterations
+        if update_guess:
+            try:
```
landreman, line 1008: It's best to keep `try` blocks as small as possible, especially when all exceptions are caught instead of just a specific exception, as is the case here. A lot happens in lines 1009-1028, so if an exception occurs, we don't know exactly what caused it. Can you move code outside of the `try` block so it encloses only the specific line that you expect to throw an exception?

src/simsopt/mhd/spec.py
```diff
@@ -303,28 +974,74 @@ def run(self):
...
+                            if not si.istellsym:
+                                new_guess[lvol].set_rs(mm, nnorm, self.results.output.Rbs[lvol+1, ii])
+                                new_guess[lvol].set_zc(mm, nnorm, self.results.output.Zbc[lvol+1, ii])
+
+                    axis = {}
+                    axis['rac'] = self.results.output.Rbc[0, 0:si.ntor+1]
+                    axis['zas'] = self.results.output.Zbs[0, 0:si.ntor+1]
+                    self.axis = copy.copy(axis)
+
+            except:
```
landreman, line 1030: Also it would be good to catch only a specific type of Exception here.

tests/mhd/test_spec.py
```diff
@@ -190,6 +331,10 @@ def test_integrated_stellopt_scenarios_1dof_Garabedian(self):
             surf = SurfaceGarabedian.from_RZFourier(equil.boundary)
             equil.boundary = surf
 
+            # Need to set axis to zero
+            equil.axis['rac'][:] = 0
+            equil.axis['zas'][:] = 0
```
landreman, line 335: It seems unsatisfying that this needs to be done manually. Can this be set automatically as a default?

src/simsopt/field/normal_field.py
```diff
@@ -0,0 +1,281 @@
...
+        if m < 0 or m > self.mpol:
+            raise ValueError('m out of bound')
+        if n < -self.ntor or n > self.ntor:
+            raise ValueError('n out of bound')
+        if m == 0 and n < 0:
+            raise ValueError('n has to be positive if m==0')
+
+    def _make_names(self):
+        """
+        Form a list of names of the ``rc``, ``zs``, ``rs``, or ``zc``
```
landreman, line 193: r and z -> v

src/simsopt/field/normal_field.py
```diff
@@ -0,0 +1,281 @@
...
+        Change the values of `mpol` and `ntor`. Any new Fourier amplitudes
+        will have a magnitude of zero.  Any previous nonzero Fourier
+        amplitudes that are not within the new range will be
+        discarded.
+        """
+        old_mpol = self.mpol
+        old_ntor = self.ntor
+        old_vns = self.vns
+        if not self.stellsym:
+            old_zc = self.vnc
```
landreman, line 230: z -> v

abaillod:
@landreman, I think I implemented all your suggestions. For some reasons, there are however some tests that are not successful. I will have a look at this tomorrow

abaillod:
@landreman This is ready to be reviewed. I don't know why the docker container / docker check is failing. It seems a password is missing?

landreman (commented):
Looks good, just left a couple questions. The remaining issue is why the spec integrated tests are failing in the CI.

tests/mhd/test_spec.py
```diff
@@ -292,6 +434,11 @@ def test_residue(self):
 
         # Initialize SPEC from an input file
         spec = Spec(filename)
+
+        # set axis to zero
+        spec.axis['rac'][:] = 0
+        spec.axis['zas'][:] = 0
+
```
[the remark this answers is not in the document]
abaillod: In the old version of `spec.py`, the coordinate axis given as input to SPEC, i.e. `rac` and `zas`, were ignored. Instead, these arrays were filled with zeros and SPEC constructed its own guess for the coordinate axis.

I modified `spec.py` to read the initial guess in SPEC input file. See [these lines](https://github.com/abaillod/simsopt/blob/4b4ba1ba98ae631f67dccbc9e6b958d19d62aa4e/src/simsopt/mhd/spec.py#L210-L220) and [these lines](https://github.com/abaillod/simsopt/blob/4b4ba1ba98ae631f67dccbc9e6b958d19d62aa4e/src/simsopt/mhd/spec.py#L792-L798).

When using the initial guess for the axis in `QH-residues.sp`, the `residue1` differs by about 0.5% from the value in the test. Another fix would be to modify `tests/test_files/QH-residue.sp` and remove the axis from it. A similar fix had to be made to `examples/2_Intermediate/inputs/QH-residues.sp` so that the example `examples/2_Intermediate/eliminate_magnetic_islands.py` still works. Let me know what you think is best!

src/simsopt/field/normal_field.py
```diff
@@ -0,0 +1,293 @@
...
+        if self.stellsym:
+            self.ndof = self.ntor + self.mpol * (2 * self.ntor + 1)
+        else:
+            self.ndof = 2 * (self.ntor + self.mpol * (2 * self.ntor + 1)) + 1
+
+        # Pack in a single array
+        if self.stellsym:
+            nvn = self.ntor + self.mpol*(2*self.ntor+1)
+        else:
+            nvn = 2*(self.ntor + self.mpol*(2*self.ntor+1)) + 1
```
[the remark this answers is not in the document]
abaillod: True - these lines have now been removed.

tests/mhd/test_spec.py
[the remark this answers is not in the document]
landreman: > When using the initial guess for the axis in `QH-residues.sp`, the `residue1` differs by about 0.5% from the value in the test. Another fix would be to modify `tests/test_files/QH-residue.sp` and remove the axis from it. A similar fix had to be made to `examples/2_Intermediate/inputs/QH-residues.sp` so that the example `examples/2_Intermediate/eliminate_magnetic_islands.py` still works. Let me know what you think is best!

Got it. I'm indifferent then between 3 options: (1) including those lines to set the axis to 0, (2) editing `QH-residue.sp` so the axis is 0 there, or (3) loosening the tolerance of the test so it passes without setting the axis to 0. For option (1), it might be useful to state in a comment why those lines are there.

src/simsopt/mhd/spec.py
```diff
@@ -215,9 +739,14 @@ def init(self, filename: str):
         spec.preset()
         logger.debug("Done with init")
 
-    def run(self):
+    def run(self, update_guess: bool = True):
```
landreman, line 742: To what extent does the default `update_guess=True` confuse finite differencing and make optimization depend on the number of MPI processes? If the initial guesses for the axis and internal surfaces are not the same for every function evaluation, then the objective function depends slightly on the history of previous objective function evaluations on that process. That history in turn depends on how many MPI processes are used. For finite difference derivatives, where tiny differences in the objective function are particularly important, this could be an issue.

src/simsopt/field/normal_field.py
```diff
@@ -0,0 +1,289 @@
...
+                    ii = self.get_index_in_dofs(mm, nn, even=True)
+                    dofs[ii] = vnc[mm, input_ntor+nn]
+
+        Optimizable.__init__(
+            self,
+            x0=dofs,
+            names=self._make_names())
+
+    @classmethod
+    def init_from_spec(cls, filename):
```
mbkumar, line 85: Please change the `init_from_spec` to `from_spec`.

src/simsopt/mhd/profiles.py
```diff
@@ -55,6 +56,88 @@ def plot(self, ax=None, show=True, n=100):
...
+    def f(self, lvol: int):
+        """
+        Return the value of the profile in volume lvol
+
+        Args:
+            lvol: int, list or np.array of int, between 0 and Mvol
+        """
+
+        # If input is a integer, make an np.array
+        if isinstance(lvol, int) or isinstance(lvol, float):
```
mbkumar, line 86: Please use `numbers.Number` for the check at 86.

src/simsopt/mhd/profiles.py
```diff
@@ -55,6 +56,88 @@ def plot(self, ax=None, show=True, n=100):
...
+
+            dp/ds = \sum_l [[p]]_l \psi_{edge} \delta(\psi_t-\psi_{t,l})
+
+        with p the profile, and the sum is on the interfaces.
+
+        Args:
+            lvol: int, list or np.array of int, between 1 and Mvol-1. 
+        """
+        # If input is a integer, make an np.array
+        if isinstance(lvol, int) or isinstance(lvol, float):
```
mbkumar, line 120: Please use numbers.Number at 120.

mbkumar (changes requested):
Hello Antoine,

Thanks for the PR with the code improvements and additional features. My comments are minor, but please address them.

src/simsopt/mhd/spec.py
abaillod: Good point - I haven't thought about this.

My first approach would be to set `update_guess=False` whenever a worker calls `spec.run()` to evaluate a finite-difference derivative, and only call with `update_guess=True` when a step is taken by the optimizer. I guess this means we would have to modify `least_square_mpi_solve._f_proc0` and `least_square_mpi_solve._constrained_mpi_workers_task`?

I don't think this is ideal - I would prefer to only modify `spec.py`. Any idea?

tests/mhd/test_profiles.py
```diff
@@ -135,13 +173,30 @@ def test_pressure(self):
...
-            np.testing.assert_allclose(pressure(s), ne(s) * Te(s) + nD(s) * TD(s) + nT(s) * TT(s),
-                                       atol=atol)
-            np.testing.assert_allclose(pressure.dfds(s),
-                                       ne.dfds(s) * Te(s) + ne(s) * Te.dfds(s) \
-                                       + nD.dfds(s) * TD(s) + nD(s) * TD.dfds(s) \
-                                       + nT.dfds(s) * TT(s) + nT(s) * TT.dfds(s),
-                                       atol=atol)
+            np.testing.assert_allclose(
+                pressure(s),
+                ne(s) *
```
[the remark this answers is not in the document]
abaillod: Ok

src/simsopt/mhd/spec.py
landreman: > When evaluating finite differences, only small deviations dx of each degree of freedom x are considered around x_0. The equilibrium obtained by SPEC at x_0+dx is thus very close to the one at x_0, and can also be used as an initial guess. The way I understand it is that the objective function depends on SPEC equilibrium, but _does not_ depend on SPEC initial guess. 

Ideally, if the equations are solved in spec to very tight tolerance, the objective should not depend on the initial guess. But for finite tolerances, the initial guess may affect the result a bit, and it only takes a small change to the objective to cause an O(1) change to a derivative computed by finite differencing. As an indication that we might want to worry about this, consider the issue discussed in this PR of the residue changing by 0.5% when the coordinate axis shape changed. Also note that when there are multiple worker groups, when an MPI process in group > 0 starts evaluating the objective at x_j + dx for a Jacobian, the initial condition will come from a run from the previous Jacobian evaluation, at x_{j-1}, which may not be super close to x_j.

It's fine with me to have this default `update_guess = True`, since it's working for you. Just wanted to flag this in case it affects your calculations.

abaillod: You are right, we do not necessarily solve SPEC equations to machine precision, which might cause issues when evaluating finite differences. As I see it, we have two choices:

(1) Leave it as it is, with the risk that the different initial guesses might affect the Jacobian evaluation. 

(2) Enforce the initial guess to be the same for each finite-difference evaluation. The only (simple) way to do it is to modify the calls to the objective function from the solver and give an additional argument that contains the information on whether or not this call is for a finite-difference evaluation.

For example, [in finite_difference.py](https://github.com/hiddenSymmetries/simsopt/blob/c3f89c83a3013a2ea9ce7cbfaa66f6bbed0d3caf/src/simsopt/_core/finite_difference.py#LL256C42-L256C42),
```python
options = dict('finite_diff'=True)
out = np.asarray(self.fn( options ))
```
and then, in [spec.py](https://github.com/hiddenSymmetries/simsopt/blob/c3f89c83a3013a2ea9ce7cbfaa66f6bbed0d3caf/src/simsopt/mhd/spec.py#L218),
```python
def run(self, options):
    if options['finite_diff']:
        update_guess=False
    else:
        update_guess=True
```

The downside is that all objective functions must accept an `options` argument... which might require some changes in different files.

While option (2) is probably better, I think leaving it for another PR is best. What do you think?

landreman: I agree that this can be addressed later in a separate pull request.

landreman:
I think this is ready to be merged as soon as the failing integrated tests are fixed. The error message is different in https://github.com/hiddenSymmetries/simsopt/actions/runs/4916745286/jobs/8780906806?pr=304 vs https://github.com/hiddenSymmetries/simsopt/actions/runs/4916745286/jobs/8780906678?pr=304 but it appears the failing example is `stellarator_benchmarks/1DOF_circularCrossSection_varyAxis_targetIota_spec.py` in both cases. @abaillod are you able to fix this?

abaillod:
> I think this is ready to be merged as soon as the failing integrated tests are fixed. The error message is different in https://github.com/hiddenSymmetries/simsopt/actions/runs/4916745286/jobs/8780906806?pr=304 vs https://github.com/hiddenSymmetries/simsopt/actions/runs/4916745286/jobs/8780906678?pr=304 but it appears the failing example is `stellarator_benchmarks/1DOF_circularCrossSection_varyAxis_targetIota_spec.py` in both cases. @abaillod are you able to fix this?

Ok, this is now fixed. The problem was again related to SPEC coordinate axis; I fixed it by adding [these lines](https://github.com/abaillod/simsopt/blob/09518ea34afc4a5294fb15f4e694539696a1afa5/examples/stellarator_benchmarks/1DOF_circularCrossSection_varyR0_targetVolume_spec.py#L55-L60) and by removing the axis from [there](https://github.com/abaillod/simsopt/blob/09518ea34afc4a5294fb15f4e694539696a1afa5/examples/stellarator_benchmarks/inputs/1DOF_Garabedian.sp#L34).

landreman (approved):

PR #310, opened 2023-05-09
Method for loading makegrid files

MigMadeira:
Although simsopt allows exporting coils to a MAKEGRID input format it does not allow loading MAKEGRID inputs. This would be a useful tool. With this in mind I adapted the method `load_curves_from_file` from `CurveXYZFourier` so that it also allows the use of "coils." files and added the method  `coils_via_file` to `Coil`  that imports a makegrid file. I have already developped some unit tests to verify the methods. 

My methods are functional and converge for high orders. However, my accuracy is lower than I had hoped for, specially when calculating the field close to the coils (check `test_coils_via_file` in `test_coil.py`). 

I was hoping you could help me achieve a better approximation.


src/simsopt/util/coil_util.py
```diff
@@ -0,0 +1,72 @@
+def importCurves(file):
```
mbkumar, line 1: How is this different from the load_curves_from_file method? Can we distinguish them in an easy way?

mbkumar (commented):
Few minor comments.

src/simsopt/util/coil_util.py
MigMadeira: I added these by mistake on my last commit. I will remove them right away. I used them while I was still testing load_curves_from_file.

src/simsopt/util/coil_util.py
```diff
@@ -0,0 +1,72 @@
...
+                coilPos[coilN].append(floatVals)
+                coilN = coilN+1
+                coilPos.append([])
+            except:
+                break
+
+    return coilPos[:-1]
+
+
+def importCoils_and_current(file):
```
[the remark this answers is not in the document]
MigMadeira: same as above

mbkumar:
@MigMadeira 
Is this ready for re-review?

MigMadeira:
> @MigMadeira Is this ready for re-review?

yes

src/simsopt/field/coil.py
```diff
@@ -186,6 +186,39 @@ def coils_via_symmetries(curves, currents, nfp, stellsym):
     return coils
 
 
+def load_coils_from_file(filename, order, ppp=20, maxiter=1000, tol=1e-8):
```
mbkumar, line 189: Can you use `load_coils_from_makegrid_file`?  We may have different file formats and don't want to confuse users.

src/simsopt/geo/curvexyzfourier.py
```diff
@@ -72,16 +74,77 @@ def set_dofs(self, dofs):
         sopp.CurveXYZFourier.set_dofs(self, dofs)
 
     @staticmethod
-    def load_curves_from_file(filename, order=None, ppp=20, delimiter=','):
+    def load_curves_from_file(filename: str, order: int, ppp=20, delimiter=',', maxiter=2000, tol=1.48e-8):
```
mbkumar, line 77: What is the change introduced in this function? If it is to support makegrid format, use a different function and leave the existing functionality alone.

src/simsopt/geo/curvexyzfourier.py
```diff
@@ -72,16 +74,77 @@ def set_dofs(self, dofs):
...
+            ppp: point-per-period: number of quadrature points per period.
+            delimiter: in the case of a file containing Fourier coefficients the delimiter between coefficients.
+            maxiter: maximum number of iterations for the integration method.
+            tol: minimum tolerance for the convergence of the integration method. 
+        Returns:
+            A list of objects CurveXYZFourier with the coefficients given by the file.
+
         """
-        coil_data = np.loadtxt(filename, delimiter=delimiter)
+        if "coils." in str(filename):
```
mbkumar, line 100: `"coils."` is very arbitrary. Someone could change the filename. Do you mean `".coils"`?

src/simsopt/util/fourier_interpolation.py
```diff
@@ -52,3 +53,55 @@ def fourier_interpolation(fk, x):
 
     # Evaluate interpolant as matrix-vector products
     return np.dot(D, w * fk) / np.dot(D, w)
+
+
+def sin_coeff(f, n, L=pi, maxiter=2000, tol=1.48e-8):
```
mbkumar, line 58: Why can't we use the scipy.fft or fftw3 library here? What is the need for separate functions?

mbkumar (changes requested):

src/simsopt/geo/curvexyzfourier.py
MigMadeira: To keep the number of functions to a minimum I adapted the already existing function "load_curves_from_file" to also support the makegrid format instead of creating a new one, lets say "load_coils_from_makegrid_file". Would it be better to have two separate functions?

src/simsopt/util/fourier_interpolation.py
MigMadeira: Well it's not that we can't, but the way I have done it is to the best of my knowledge the most straight forward way.  scipy.fft is implemented in the complex exponential form and also adds the intricacy of manipulating the data before performing the fft so that the right fundamental frequency is picked. Nonetheless, I will try to implement it with the fft and get back to you.

mbkumar:
The changes look good. Please fix the failing tests

mbkumar (approved):
Thanks!

mbkumar (approved):
Thanks!

landreman:
In the makegrid input format, for each coil, the first data point is repeated at the end (on the line that lists the coil group.) This repeated point had been included in the data being fft-ed, which was causing the dofs to change a bit if you save and then load. I refactored to fix this, and added some assertions that when you save and then load, the dofs are unchanged to high precision. Once the CI passes, this should be good to go.

mbkumar (approved):

PR #317, opened 2023-05-26
Refactor single stage

landreman:
In #301, we've discussed how it is awkward that `finite_difference.py` had to be modified to manipulate the `non_dofs` attribute, which most Optimizable objects do not have. This PR shows one way things could be refactored to avoid this, so `non_dofs` is never mentioned. Here the changes to `finite_difference.py` and `optimizable.py` in #301 are reverted to master, and instead the `MPIFiniteDifference` class is extended to allow extra data besides `.x` to be broadcast among the worker groups. The example `single_stage_optimization_finite_beta.py` is refactored correspondingly. This example gives identical values for all quantities at the end of the optimization and uses fewer calls to `VirtualCasing.from_vmec()` in this branch compared to #301.

I'm not wedded to the approach here, but wanted to put it on the table. @rogeriojorge @mbkumar what do you think?

### DIFF
--- a/src/simsopt/_core/optimizable.py
+++ b/src/simsopt/_core/optimizable.py
@@ -936,9 +936,7 @@ class Optimizable(ABC_Callable, Hashable, GSONable, metaclass=OptimizableMeta):
         # TODO: node after fixing/unfixing any DOF
         dof_indices = [0]
         free_dof_size = 0
-        dof_objs = set()
         for opt in self._unique_dof_opts:
-            dof_objs.add(opt.dofs)
             size = opt.local_dof_size
             free_dof_size += size
             dof_indices.append(free_dof_size)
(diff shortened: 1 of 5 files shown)


tests/core/test_finite_difference.py
```diff
@@ -13,6 +13,9 @@
     from simsopt.util.mpi import MpiPartition
     from simsopt._core.finite_difference import MPIFiniteDifference
 
+#from simsopt.util.mpi import log
+#log()
+
```
rogeriojorge, line 18: Should these comments stay?

rogeriojorge (approved):
This looks much cleaner than the previous approach. Besides the left out comments, this looks very good to me.

rogeriojorge:
@mbkumar are you ok with merging this into rj/single_stage_PR and then master?

src/simsopt/_core/optimizable.py
```diff
@@ -1639,7 +1639,6 @@ class TempOptimizable(Optimizable):
         def __init__(self, func, *args, dof_indicators=None, **kwargs):
 
             self.func = func
-            args = np.ravel(args)  # The user may pass a tuple or list
```
mbkumar, line 1642: Why is this deleted?

mbkumar (approved):
This implementation is tidy and looks good.

src/simsopt/_core/optimizable.py
rogeriojorge: I think that line was present only in the rj/single_stage_PR branch, and not in the master branch. Therefore, its removal makes ml/single_stage_PR even simpler.

rogeriojorge (approved):

PR #325, opened 2023-06-15
New magnetic field class - mirror machine

rogeriojorge:
This PR introduces the mirror machine model from https://arxiv.org/abs/2305.06372
It expands the set of magnetic field models available to SIMSOPT to include a magnetic mirror model for fusion applications. The default input parameters follow the ones used to model WHAM (https://wippl.wisc.edu/wisconsin-hts-axisymmetric-mirror/)

### DIFF
--- a/docs/source/fields.rst
+++ b/docs/source/fields.rst
@@ -170,6 +170,32 @@ Physics Communications 43 (1986) 157—167
 It is an analytical magnetic field representation that allows the
 explicit calculation of the width of the magnetic field islands.
 
+MirrorModel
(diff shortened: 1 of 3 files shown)


rogeriojorge:
Ready for review. However, for some reason, I don't obtain a symmetric grad B tensor. Still checking out why.

rogeriojorge:
Actually the field is not curl-free so the gradB tensor is expected to not be symmetric

mbkumar (dismissed):
Looks good to me.

landreman (approved):

PR #326, opened 2023-06-16
Descriptor protocol

mbkumar:
This PR introduces descriptor protocol for constrained class fields, so one doesn't have to test (mostly) for the constraints on the arguments in the ``__init__`` method.


mbkumar:
@landreman 
This is ready for review

landreman (approved):

PR #332, opened 2023-06-26
fix #278: tracing.hpp does not find boost

### LINKED ISSUE #278: Building on Arch Linux in Anaconda
Hi,

I had some trouble building `simsopt` on Arch Linux in an Anaconda environment, which might be interesting for you.
A problem occured when building `simsoptpp`. The `boost` include in `tracing.hpp` was not found, since the `Boost_INCLUDE_DIRS` were not forwarded by CMake/pybind11 to the compiler.
Here is the call that should have built the `tracing.cpp.o` object:
```bash
/home/jonathan/anaconda3/envs/simsopt/bin/x86_64-conda-linux-gnu-c++ -Dsimsoptpp_EXPORTS -I/data/jonathan/work/code/simsopt/thirdparty/xtensor/include -I/data/jonathan/work/code/simsopt/thirdparty/xtensor-python/include -I/data/jonathan/work/code/simsopt/thirdparty/xsimd/include -I/data/jonathan/work/code/simsopt/thirdparty/xtl/include -I/data/jonathan/work/code/simsopt/thirdparty/eigen -I/home/jonathan/.local/lib/python3.10/site-packages/numpy/core/include -I/data/jonathan/work/code/simsopt/src/simsoptpp -I/data/jonathan/work/code/simsopt/thirdparty/fmt/include -isystem /home/jonathan/anaconda3/envs/simsopt/include/python3.10 -isystem /data/jonathan/work/code/simsopt/thirdparty/pybind11/include -O3 -march=native -mfma -ffp-contract=fast -O3 -DNDEBUG -fPIC -fvisibility=hidden -flto -fno-fat-lto-objects -fopenmp -std=gnu++14 -MD -MT CMakeFiles/simsoptpp.dir/src/simsoptpp/tracing.cpp.o -MF CMakeFiles/simsoptpp.dir/src/simsoptpp/tracing.cpp.o.d -o CMakeFiles/simsoptpp.dir/src/simsoptpp/tracing.cpp.o -c /data/jonathan/work/code/simsopt/src/simsoptpp/tracing.cpp
```

Boost was found (indicated in the output above the cited call) at `/home/jonathan/anaconda3/envs/simsopt/include`, which is however not a system-wide include dir (hence the requirement for an explict mention of it on the compiler call), but also does not show up in the compiler call.

I was able to fix this by specifying (in `CMakeLists.txt`)
`set_target_properties(${PROJECT_NAME} PROPERTIES COMPILE_FLAGS "-I${Boost_INCLUDE_DIRS}")`
instead of
`target_include_directories(${PROJECT_NAME} PRIVATE ${Boost_INCLUDE_DIRS})`.

To my understanding, those two options should have almost in all cases exactly the same effects.
However, I seem to have managed to stumble into a weird edge case scenario where the effect of those two calls is not identical.

Any ideas?

jons-pf:
This PR fixes #278 .

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -136,7 +136,7 @@ target_link_libraries(${PROJECT_NAME} PRIVATE fmt::fmt-header-only)
 if(NOT Boost_FOUND)
     add_dependencies(${PROJECT_NAME} ${boost_target})
 endif()
-target_include_directories(${PROJECT_NAME} PRIVATE ${Boost_INCLUDE_DIRS})
+set_target_properties(${PROJECT_NAME} PROPERTIES COMPILE_FLAGS "-I${Boost_INCLUDE_DIRS}")
 # target_link_libraries(${PROJECT_NAME} PRIVATE pybind11::module Boost::headers)
 
 if(OpenMP_CXX_FOUND)


mbkumar (approved):

jons-pf:
Thanks for merging this! :-)

PR #333, opened 2023-06-26
additional VMEC input file parameters

jons-pf:
- free-boundary parameters included in VMEC input file
- ntheta, nzeta resolution parameters included in VMEC input file


jons-pf:
also this PR seems to be blocked by #334

landreman (approved):

PR #337, opened 2023-06-26
Add support for python 3.11

mbkumar:


landreman:
Should some of the CI tests run with python 3.11 then? Presently they run on (3.8, 3.9, 3.10). If we modified the extensive CI to run on (3.8, 3.10, 3.11), that would effectively still cover 3.9 fairly well.

mbkumar:
I'd wait one more year for running CI on 3.11. I consider 3.11 to be
experimental and the support for simsopt on 3.11 is still in the beta
phase.
We followed the same logic for 3.10. The wheels and conda packages are
released much earlier before we started testing for 3.10 compatibility on
CI.

Bharat Medasani

Engineer

landreman (approved):

PR #338, opened 2023-06-27
CI for docs

mbkumar:
This CI is to catch failures when building the docs.

### DIFF
--- a/.github/workflows/docs_test.yml
+++ b/.github/workflows/docs_test.yml
@@ -19,6 +19,9 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
 
+    - name: which pip after python setup
+      run: python -m pip install --upgrade pip
+
     - name: Install sphinx and its dependencies for building docs
       run: |
         pip install sphinx sphinx-rtd-theme sphinxcontrib-napoleon sphinx-autodoc-napoleon-typehints
--- a/.github/workflows/docs_test.yml
+++ b/.github/workflows/docs_test.yml
@@ -12,11 +12,6 @@ jobs:
         python-version: [3.9]
 
     steps:
-    # - name: apt-get stuff needed for libstell and vmec
-    #   run: |
-    #     sudo apt-get update
-    #     sudo apt-get install -y build-essential gfortran openmpi-bin libopenmpi-dev libnetcdf-dev libnetcdff-dev liblapack-dev libscalapack-mpi-dev libhdf5-dev libhdf5-serial-dev git m4 libfftw3-dev libboost-all-dev libopenblas-dev
-
     - uses: actions/checkout@v3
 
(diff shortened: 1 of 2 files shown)


.github/workflows/docs_test.yml
```diff
@@ -0,0 +1,54 @@
...
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    - name: which pip after python setup
+      run: python -m pip install --upgrade pip
+
+    - name: Install sphinx and its dependencies for building docs
+      run: |
+        pip install sphinx sphinx-rtd-theme sphinxcontrib-napoleon sphinx-autodoc-napoleon-typehints
```
landreman, line 32: Would it make sense to use `pip install -r docs/requirements.txt` here, to provide some test coverage for the `docs/requirements.txt` file?

.github/workflows/docs_test.yml
```diff
@@ -0,0 +1,54 @@
...
+    - name: Install python dependencies for simsopt
+      run: |
+        sudo apt-get install graphviz graphviz-dev
+        pip install wheel numpy scipy f90nml h5py scikit-build cmake qsc sympy pyevtk matplotlib ninja plotly networkx pygraphviz
+
+    - name: Install booz_xform
+      run: pip install -v git+https://github.com/hiddenSymmetries/booz_xform
+
+    - name: Install virtual_casing
+      run: pip install -v git+https://github.com/hiddenSymmetries/virtual-casing
```
landreman, line 43: In this workflow, is installing booz_xform and virtual_casing necessary?

.github/workflows/docs_test.yml
```diff
@@ -0,0 +1,54 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [3.9]
+
+    steps:
+    - name: apt-get stuff needed for libstell and vmec
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y build-essential gfortran openmpi-bin libopenmpi-dev libnetcdf-dev libnetcdff-dev liblapack-dev libscalapack-mpi-dev libhdf5-dev libhdf5-serial-dev git m4 libfftw3-dev libboost-all-dev libopenblas-dev
```
landreman, line 18: Since vmec and spec are not installed I think we can drop `gfortran libscalapack-mpi-dev m4 libfftw3-dev`. If we don't install booz_xform and virtual_casing then I think we can also drop `libnetcdf-dev libnetcdff-dev liblapack-dev  libhdf5-dev libhdf5-serial-dev libopenblas-dev`. Since MPI isn't used, perhaps we can also drop `openmpi-bin libopenmpi-dev`

mbkumar:
@landreman 
Updated as per the comments raised

landreman (approved):

PR #341, opened 2023-06-29
Flip orientation in create_equally_spaced_curves() to fix free-boundary

landreman:
This PR addresses an issue we noticed when doing free-boundary vmec calculations following stage-2 coil optimization. The function `create_equally_spaced_curves()` has a sign convention associated with which way the curve parameter increases. In other words, does a positive current correspond to current going up or down through the plasma's donut hole? We need to be particularly careful about this sign in finite-beta stage-2 optimizations (see `examples/2_Intermediate/stage_two_optimization_finite_beta.py`), where the sign of the coil currents is set by `Vmec.external_current()`, and Bnormal from virtual casing must have the proper sign relation to B from the coils. In the master branch, the signs are not correct, such that when you run free-boundary vmec from the optimized coils, there are two symptoms. First, VMEC complains that B_toroidal has the wrong direction:
~~~~
  R * BTOR(vac) =  -8.14E+01 R * BTOR(plasma) =   8.14E+01

 PHIEDGE HAS WRONG SIGN IN VACUUM SUBROUTINE
~~~~
Second, if you try to fix this problem by flipping the sign of the `EXTCUR` values or `PHIEDGE` and `CURTOR`, the free-boundary equilibrium either doesn't converge or doesn't match the target. Here's an example for the W7-X finite-beta case in `stage_two_optimization_finite_beta.py`:
image
(red = fixed-boundary target, green = free-boundary solution). The mismatch arises because the relative sign of Bcoils and Bplasma was not correct during the coil optimization, so it can't be fixed by sign flips afterwards. After the sign flip in this PR, the free-boundary solution matches the target:
image
Differences are even more pronounced for QH/QA cases since they have more plasma current.

I thought Florian and I had sorted out these signs back around April 2022 (I have some notes dated 20220417-03 discussing the signs), but it appears the fix never made it into master? Perhaps someone else remembers what happened better than I. Anyhow, with this sign flip in this PR, things work as of today.

This PR includes a new test to cover this issue.

This sign flip in `create_equally_spaced_curves()` will reverse the direction of B in many scripts, so there may be side-effects of this change in your applications. If you have any concerns, please comment.

### DIFF
--- a/src/simsopt/geo/curve.py
+++ b/src/simsopt/geo/curve.py
@@ -875,7 +875,10 @@ def create_equally_spaced_curves(ncurves, nfp, stellsym, R0=1.0, R1=0.5, order=6
         curve.set("xc(1)", cos(angle)*R1)
         curve.set("yc(0)", sin(angle)*R0)
         curve.set("yc(1)", sin(angle)*R1)
-        curve.set("zs(1)", R1)
+        # The the next line, the minus sign is for consistency with
+        # Vmec.external_current(), so the coils create a toroidal field of the
+        # proper sign and free-boundary equilibrium works following stage-2 optimization.
+        curve.set("zs(1)", -R1)
         curve.x = curve.x  # need to do this to transfer data to C++
         curves.append(curve)
     return curves
--- a/src/simsopt/mhd/vmec.py
+++ b/src/simsopt/mhd/vmec.py
@@ -574,7 +574,7 @@ class Vmec(Optimizable):
             nml += f"MGRID_FILE = '{vi.mgrid_file.decode('utf-8')}'\n"
             nml += 'EXTCUR = ' + array_to_namelist(vi.extcur)
             nml += '\n'
-        
+
         nml += '\n! ---- Resolution parameters ----\n'
         nml += f'MPOL = {vi.mpol}\n'
         nml += f'NTOR = {vi.ntor}\n'
--- a/tests/mhd/test_vmec.py
+++ b/tests/mhd/test_vmec.py
@@ -18,7 +18,8 @@ except ImportError:
 
 from simsopt._core.optimizable import make_optimizable
 from simsopt.objectives.least_squares import LeastSquaresProblem
-from simsopt.geo.surfacerzfourier import SurfaceRZFourier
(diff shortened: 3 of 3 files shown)


zhucaoxiang:
I remember that once I asked about the signs when the Virtual Casing code was merged. I shared the note (https://princetonuniversity.github.io/FOCUS/notes/Coil_design_codes_benchmark.html) that I wrote several years ago. In the appendix, I observed the sign issue with NESCOIL and REGCOIL when doing free-boundary calculations. I think the issue was that BNORM is calculating Bn components using normal vectors pointing inwards, while people thought BNORM was using exterior normal vectors.  So when using BNORM results, we should flip the signs, otherwise, we will get coils that produce the opposite `phiedge`.

The note was written several years ago and might be unclear. I was trying to figure out whether I was correct by comparing the Virtual Casing code, but I didn't proceed with the task myself and it was sitting there. I am glad that Matt found this issue.

One question I  have is that in my example (check the appendix A), I can get a very good match if simply flipped the sign of `extcur`, but this is not the case in Matt's observation.

> An mgrid file is generated by these modular coils and then used in VMEC for free-boundary running. Somehow, the sign of phiedge in VMEC has to be flipped. Except that, the free-boundary VMEC results are consistent with fixed-boundary results, with surprisingly good precision.

rogeriojorge:
Would it be possible to add a simple VMEC free-boundary run in a testing routine? This would not only make it more future-proof, but it would also be a place with a free-boundary Vmec example that works.

landreman:
> Would it be possible to add a simple VMEC free-boundary run in a testing routine? 

For test coverage on this issue that uses free-boundary vmec, we would need to generate an mgrid file during the test. There are 2 options for doing this. (1) Download and build xgrid from stellopt in the simsopt CI, or (2) finish #233 . Probably 2 is the way to go. I would suggest merging the present PR and adding a free-boundary test later when #233 is completed, since #233 might take some time to resolve.

rogeriojorge:
Ok, I have tested on my end and it seems to provide good results with free boundary VMEC.

rogeriojorge (approved):

PR #342, opened 2023-07-01
Cylindrical magnets grid fix

akaptano:
This branch fixes an issue with the grid generation for permanent magnets, particularly for uniform cylindrical grids, pointed out by @MigMadeira. Basically, the magnets are initialized on a uniform grid (cartesian, cylindrical, or simple toroidal), and then magnets that do not lie between an inner and outer toroidal surface are eliminated by a ray-tracing routine. Some of the magnets near pi and 2pi  were incorrectly eliminated or incorrectly kept around. I think this is because of not being careful that the inner, outer, and plasma surfaces all had the same toroidal angle range between [0, 2pi], and that the ray tracing knew that 0 and 360 were the same angle. 

Part of this issue is that, for historical reasons (bad coding), I had a separate ray tracing routine for both the cartesian and cylindrical scenarios. Now, the code is reduced to a single ray tracing routine for all the coordinate systems, the unit tests for this method are improved, and the bug eliminated. The code is also now compatible with scenarios in which the resolution of the inner and outer toroidal surfaces is different than the plasma surface.

### DIFF
--- a/src/simsopt/geo/permanent_magnet_grid.py
+++ b/src/simsopt/geo/permanent_magnet_grid.py
@@ -72,65 +72,86 @@ class PermanentMagnetGrid:
         some important grid variables for later.
         """
         # Get (X, Y, Z) coordinates of the two boundaries
-        xyz_inner = self.inner_toroidal_surface.gamma()
-        self.xyz_inner = xyz_inner.reshape(-1, 3)
-        xyz_outer = self.outer_toroidal_surface.gamma()
-        self.xyz_outer = xyz_outer.reshape(-1, 3)
-        self.x_outer = xyz_outer[:, 0]
-        self.y_outer = xyz_outer[:, 1]
-        self.z_outer = xyz_outer[:, 2]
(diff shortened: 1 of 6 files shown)


rogeriojorge:
@MigMadeira can you run your scripts on this branch and check if the problem is fixed on your end?

MigMadeira:
> @MigMadeira can you run your scripts on this branch and check if the problem is fixed on your end?

It is also fixed on my end. Thanks for the bug fix @akaptano!

akaptano:
small note on this -- the ray tracing does give small errors if the resolution is too low. For the full torus using the tokamak equilibrium, 32 points in the toroidal direction leads to errors in the ~0.01 range, so in other words, magnets may be retained that are ~0.01 outside the outer toroidal surface (that in principle should have been removed). However, this error is small and decreases steadily with the resolution. I account for this in the unit tests.

rogeriojorge (approved):

PR #343, opened 2023-07-14
tidy up examples

landreman:
Lots of the examples had a test to see if they were being run inside the CI. Here, this function is brought into `simsopt.util` as a new variable `in_github_actions` so it doesn't need to be implemented in every example. Also, many of the MPI examples printed messages on every process, rather than just on process 0, or had their own function defined to print only on proc 0. This is fixed/streamlined now using a new function `simsopt.util.proc0_print()` which is like `print()` but only on proc 0, which all the relevant examples now use. Also some imports are tidied up and documentation has been added in some examples.


mbkumar (approved):
Looks good.

rogeriojorge:
Agreed. The proc0 print function is really handy. Every script of mine had a redefined print function checking for the mpi processor that can now be streamlined.

rogeriojorge (approved):

PR #344, opened 2023-07-18
[FIX] theta_vmec computation in vmec_fieldlines

jons-pf:
... no idea why this still works as-is (`js` is not defined in the scope), but Python somehow magically seems to find the correct value. Anyway, I think one should use the (intended?) `jradius` here.


landreman (approved):
Good catch, thanks!

PR #347, opened 2023-08-14
Enable initial guess solutions for GPMO

kchammond:
This pull request adds a feature to the Greedy Permanent Magnet Optimization (GPMO) function allowing users to provide an initial guess of the solution to the optimizer. To date, the GPMO algorithms have all been initialized to an empty magnet array. In this PR, the user can supply a non-empty initial solution via an optional keyword argument. Presently, this feature is only supported for the `'ArbVec_backtracking'` algorithm within the GPMO function. The default behavior is to initialize to an empty magnet array; thus, this version should not affect the results of existing optimization scripts that do not set the new keyword argument.

This feature gives users the ability to perform multi-stage GPMO optimizations, with each stage fine-tuning the result of the previous stage with adjusted optimization parameters. It also provides the option for users to employ GPMO to improve on solutions obtained through other optimization methods.


rogeriojorge:
Adding @MigMadeira for extra review

landreman:
Looks reasonable to me, though I haven't studied the code logic in detail. Any comments @akaptano @MigMadeira ?

MigMadeira:
I am happy with the added features. They seem quite useful and make the backtracking algorithm easier to use while also allowing for more detailed analysis. To the best of my knowledge they follow the code philosophy up to this point, although @akaptano is better suited to comment on that. The PR also fixes a bug I had recently found.

landreman (approved):

landreman:
Thanks @MigMadeira @kchammond

PR #348, opened 2023-08-29
surface copying classmethod

smiet:
Added a classmethod to surfacerzfourier.py that copies an existing surface but allows the user to request a different `nphi`, `ntheta` and `grid_range`. 

Usecase: your existing equilibrium has half period, but you want to plot the full surface, so you use this method to generate a brother with higher resolution and full range.  

Currently only implemented surfaceRZFourier, if this is deemed useful, will generate for the other classes as well. 

[is a bit of duplication when this is added to each inherited class, but cleaner than adding a classmethod to the Surface class, as different arrays need to be populated  in each inherited class's constructor, which does not follow a neat pattern]

If useful, give me some time to add classmethod to other classes, otherwise please cancel request. If I'm duplicating functionality, my apoligies, Thanks!


andrewgiuliani:
This functionality is already implemented in the shared dofs feature.  If you want a new `SurfaceRZFourier` with the same dofs as another, but different set of quadrature points, see the following unit test:

https://github.com/hiddenSymmetries/simsopt/blob/c859dfb33e9a193a064e7221b4100b83a8950575/tests/geo/test_surface_rzfourier.py#L60

smiet:
@andrewgiuliani  I intended this method to provide different functionality from the example; the dofs are not shared, but a full copy is made. 

This would allow the user also for example make copies of the surface object during optimization, and plot a series of surfaces, in the same plot, or next to each other, and to 'save' the initial or intermediate state for comparison with the final result. 

Having it as a class method is also a nice convenience for the user, who does not have to look into the class definition. 
If it is only me who likes this way of working, let me know and I will close the pull request.

landreman (commented):
Generally I think it does seem useful to include this kind of copy function, for the use cases @smiet  mentioned, which are a bit different from the use cases for shared dofs.

In order to generalize this routine to other `Surface` types, maybe there is a clever way to do it by serializing the original surface and then deserializing?

Presently the following test fails, specifying only `ntheta`:
~~~~py
s2 = SurfaceRZFourier.from_other_surface(s, ntheta=100)
self.assertEqual(s2.quadpoints_theta.size, 100)
~~~~
Result:
~~~~
AssertionError: 62 != 100
~~~~
Do we want that to work? Or should a user need to also specify `nphi` and `range` in order to change only `ntheta`?

If you didn't already see this, running `./run_autopep` from the root directory will fix any linting errors.

andrewgiuliani:
If you want to get this working on other surface types, maybe a more general approach would work since the user is allowed to change `mpol`, `ntor`, `quadpoints`, `stellsym`, etc... One possibility that is agnostic of the surface type is to evaluate the XYZ positions on the old surface and complete a least squares projection of those points onto the new surface basis functions using `least_squares_fit`. 

https://github.com/hiddenSymmetries/simsopt/blob/f8f9d59e19bc3c83cace47eebc4425c179e289f2/src/simsoptpp/surface.cpp#L40

You'd just need to make sure there are enough quadrature points in the projection so that the normal equations aren't singular.

I only suggest this approach because it seems tedious to hard code a `change_resolution` functions for `SurfaceXYZFourier` and `SurfaceXYZTensorFourier` since you'd have to go into the weeds of the DOF ordering.

mbkumar:
Instead of having a class method, why don't you have a new method called `copy`? The intent will be clear. With the copy method, you can generate an instance of the same class as that of the original object. And the expectation would also be the same.

smiet:
@mbkumar this is the way, and this is why we have code review. I will implement the change, update the pull request and re-factor the code I have written using my convoluded method in the coming week(s)

PR #353, opened 2023-09-20
Extend virtual casing to full torus

rogeriojorge:
This PR allows the user to plot the squared flux in the full torus range when using the virtual-casing package instead of only plotting half or a single field period.


rogeriojorge:
Upon closer inspection, for optimized configurations the stellarator symmetry is spoiled. It warrants a fix.
Screenshot 2023-09-21 at 01 36 35

landreman (approved):

PR #355, opened 2023-09-22
PR: Added Interface term to single stage finitebeta

rogeriojorge:
The single stage optimization, up to now, has minimized the squared flux corrected for the target magnetic field.
According to texts on free boundary [add reference], an extra term is needed that minimizes the squared of the magnetic field just outside the plasma boundary and the square of magnetic field just inside the boundary.
Added that term here to the optimization.

### DIFF
--- a/examples/3_Advanced/single_stage_optimization_finite_beta.py
+++ b/examples/3_Advanced/single_stage_optimization_finite_beta.py
@@ -175,10 +175,10 @@ def fun_J(prob, coils_prob):
 
     bs.set_points(vc.trgt_surf.gamma().reshape((-1, 3)))
     Bvmec = np.transpose(np.array(B_cartesian(vmec, quadpoints_phi=vc.trgt_surf.quadpoints_phi,
-                                            quadpoints_theta=vc.trgt_surf.quadpoints_theta)),(1,2,0))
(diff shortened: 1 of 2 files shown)


rogeriojorge:
@f0uriest Could you help with the reference here so that we can compare the terms added here with the ones in the literature?

f0uriest:
I haven't seen any particularly clear explanation in the literature, the least bad is probably Hirshman's 1986 paper on free boundary VMEC (doi:10.1016/0010-4655(86)90058-5).

That was using NESTOR to compute the total B outside the plasma due to the coils + plasma field. The equations implemented here are similar but using the virtual casing code to get B_plasma.

One possible extra complication is the presence of a sheet current on the plasma boundary. This is needed when pressure at the edge is nonzero, but is also in general allowed even when edge pressure is zero (at least I haven't been able to prove it's not there). Physically this is just because the plasma itself is a conductor immersed in a magnetic field, so you can get a sheet current that causes a jump in the tangential component of B. In practice it seems like this is usually negligible (for all of the equilibria I've tested in free boundary DESC the field due to the sheet current ends up being < 1e-8x smaller than the field from other sources)

mishapadidar:
@rogeriojorge is this PR still in progress or should we close?

PR #358, opened 2023-10-03
Planar

AlexWiedman:
Created a curve class for planar coils with a fourier series in polar coordinates, a quaternion based rotation vector, and a center vector in cartesian coordinates. This allows for planar coils to be constructed and used in optimization.


landreman (commented):
Hi @AlexWiedman, thanks for working on this!!

src/simsopt/geo/curveplanarfourier.py
```diff
@@ -0,0 +1,66 @@
...
+    where :math:`\theta` is the counterclockwise rotation about a unit axis
+    :math:`(\hat{x},\hat{y},\hat{z})`.
+
+    The quaternion is normalized for calculations to prevent scaling. The dofs
+    themselves are not normalized. This results in a redundancy in the
+    optimization, where several different sets of dofs may correspond to the
+    same normalized quaternion. Normalizing the dofs directly would create a
+    dependence between the quaternion dofs, which may cause issues during
+    optimization.
+
```
andrewgiuliani, line 33: I'm not familiar with quaternions -- is there a computational advantage using them over a simple solid body rotation + translation?  You would have 3 dofs for the rotation and 3 dofs for the translation.  The formulas in curveplanarfourier.cpp on lines 43-45 are opaque to me.

src/simsoptpp/curveplanarfourier.cpp
```diff
@@ -0,0 +1,1633 @@
...
+    for (int k = 0; k < numquadpoints; ++k) {
+        double phi = 2 * M_PI * quadpoints[k];
+        for (int i = 1; i < order+1; ++i) {
+            data(k, 0) += rs[i-1] * ( (i) * cos(i*phi) * cos(phi) - sin(i*phi) * sin(phi));
+            data(k, 1) += rs[i-1] * ( (i) * cos(i*phi) * sin(phi) + sin(i*phi) * cos(phi));
+        }
+    }
+    data *= (2*M_PI);
+    for (int m = 0; m < numquadpoints; ++m) {
+        Array i = xt::zeros<double>({1});
```
andrewgiuliani, line 80: can i, j, k simply be stored as doubles?  Why do they need to be arrays?

src/simsoptpp/curveplanarfourier.cpp
```diff
@@ -0,0 +1,1633 @@
...
+            q_norm[i] = q[i] / std::sqrt(s); 
+    }
+    else {
+        q_norm[0] = 1;
+    }
+
+    for (int m = 0; m < numquadpoints; ++m) {
+        double phi = 2 * M_PI * quadpoints[m];
+        int counter = 0;
+        Array i = xt::zeros<double>({1});
```
andrewgiuliani, line 216: same question here, can i, j, k just be doubles?

src/simsoptpp/curveplanarfourier.cpp
```diff
@@ -0,0 +1,1633 @@
...
+    }
+    for (int m = 0; m < numquadpoints; ++m) {
+        double i;
+        double j;
+        double k;
+        i = data(m, 0);
+        j = data(m, 1);
+        k = data(m, 2);
+
+        data(m, 0) = (i - 2 * (q_norm[2] * q_norm[2] + q_norm[3] * q_norm[3]) * i + 2 * (q_norm[1] * q_norm[2] - q_norm[3] * q_norm[0]) * j + 2 * (q_norm[1] * q_norm[3] + q_norm[2] * q_norm[0]) * k) + center[0];
```
andrewgiuliani, line 43: lines 43, 44, 45 are opaque to me.  Do they complete the solid body rotation? can you comment the code here to explain, and maybe give a reference?

src/simsoptpp/curveplanarfourier.cpp
```diff
@@ -0,0 +1,1557 @@
...
+        sinphi = sin(phi);
+        cosphi = cos(phi);
+        data(k, 0) = rc[0] * cosphi;
+        data(k, 1) = rc[0] * sinphi;
+        for (int i = 1; i < order+1; ++i) {
+            siniphi = sin(i * phi);
+            cosiphi = cos(i * phi);
+            data(k, 0) += (rc[i] * cosiphi + rs[i-1] * siniphi) * cosphi;
+            data(k, 1) += (rc[i] * cosiphi + rs[i-1] * siniphi) * sinphi;
+        }
```
landreman, line 34: Look at how I refactored this loop to minimize the number of calls to `sin()` and `cos()`. These transcendental functions are slow, so it is good to call them as few times as possible, by storing the results in `cosphi`, `cosiphi`, etc. Please apply a similar refactoring to the other functions in this file to reduce the number of calls to `sin()` and `cos()`.

src/simsoptpp/curveplanarfourier.cpp
```diff
@@ -0,0 +1,1557 @@
...
+            counter++;
+        }
+    }
+}
+
+
+
+#include "xtensor-python/pyarray.hpp"     // Numpy bindings
+typedef xt::pyarray<double> Array;
+template class CurvePlanarFourier<Array>;
```
landreman, line 1557: Are these lines needed, or can they be deleted?

AlexWiedman: I'm not sure exactly what these lines do, but they are included in both of the other curve classes, so I thought it would be safest to include them.

PR #359, opened 2023-10-05
Make linting more rigorous.

mbkumar:
I'd like to expand the linting rule list to catch more issues. At the same time, I'd like to update the linting framework to the one based on ruff. In place of pep8 and flake8 tools, we will use ruff. 

To use ruff, run `pip install ruff` followed by `./run_ruff`.
The configuration for ruff is done in pyproject.toml. The linter workflow is updated to use ruff. If you are satisfied, I can commit the fixes by ruff and do one more push with all the changes to the source code.


landreman (dismissed):

mbkumar:
@landreman 
I pushed the changes where linter fixes related to unused imports and a few other minor things are applied. I have more in the pipeline, but they need more careful analysis before getting committed. So I want to get these out of the way.

mbkumar:
@landreman 
Ready for final review.

landreman (approved):

PR #361, opened 2023-10-06
ag/surface_dofs_fix

andrewgiuliani:
In this PR, I propose a fix for an issue that @ejpaul pointed out.  Python complains if you try to compute a derivative with BoozerSurfaces when a coil degree of freedom is fixed.  We didn't have any unit tests for this, but I have added one now.

I have proposed a not very elegant modification of the `__call__` function of the `Derivative` class.

The issue is that simsopt currently filters out the partial derivatives of the fixed dofs.  When the user requests it to return a `Derivative` object, rather than a numpy array of the derivative, this can cause issues.


landreman:
Here's another thought. Instead of having `call()` return derivatives with respect to only free dofs if `as_derivative==False` and return derivatives with respect to all dofs (fixed and free) if `as_derivative==True`, an alternative option could be to add a new keyword argument to `call()` that controls whether the returned data includes only fixed dofs or all dofs. Suppose this new keyword argument was named `free_only` and was `True` by default. Then users would have the option of both `call(as_derivative=True, free_only=True)` and `call(as_derivative=True, free_only=False)`. Would you foresee both of these cases as being potentially useful? I'm not saying we have to go this route, just wanted to put this option on the table.

It looks to me like `as_derivative=True` is used only in one place in all of simsopt, in `surface_objectives.py:boozer_surface_dexactresidual_dcoils_dcurrents_vjp()`. So not much will be affected by the choice of how `as_derivative=True` behaves.

andrewgiuliani:
> Can a test be added in `tests/core/test_derivative.py` to cover this issue using the little optimizable functions there without relying on `BoozerSurface`?

This is checked now in `test_as_derivative2`.  The issue was that two `Derivative` dictionaries are summed together on line 1088:
https://github.com/hiddenSymmetries/simsopt/blob/eb6b27fae8b7df43c8e5d6585de66b6ec551b23a/src/simsopt/geo/surfaceobjectives.py#L1086-L1088

The first `Derivative` dictionary had values with the full set derivatives wrt both fixed and free DOFs.  The second dictionary had values only wrt to the free DOFs obtained from the `as_derivative=True` keyword argument, hence the conflict.  `test_as_derivative2` simulates this by summing one `Derivative` obtained in the standard way, and another obtained using the `as_derivative=True` option.


> Here's another thought. Instead of having call() return derivatives with respect to only free dofs if as_derivative==False and return derivatives with respect to all dofs (fixed and free) if as_derivative==True, an alternative option could be to add a new keyword argument to call() that controls whether the returned data includes only fixed dofs or all dofs. Suppose this new keyword argument was named free_only and was True by default. Then users would have the option of both call(as_derivative=True, free_only=True) and call(as_derivative=True, free_only=False). Would you foresee both of these cases as being potentially useful? I'm not saying we have to go this route, just wanted to put this option on the table.

I decided against adding another keyword argument for `free_only=False` because this will result in two kinds of `Derivative` dictionaries, ones with derivatives wrt to all dofs and others with derivatives only wrt to free dofs.  These two kinds of `Derivatives` cannot be summed together without causing a numpy error.  This is highlighted by the unit test `test_as_derivative2`.  I think it's more intuitive for the user to always have dictionaries with keys having the derivative wrt to the full set of DOFs, this way `Derivatives` can always be summed together without issues.

landreman (approved):

Issue #371, opened 2023-10-29
Strainopt tests


mbkumar:
@ejpaul 

Could you please check the tests in test_strainopt.py? In particular, I am confused with the `subtest_binormal_curvature` and `subtest_torsion`. You are not asserting any conditions and the setting of  `errf_old` variable is not clear.

ejpaul:
Thanks, @mbkumar. I have created PR #373 to address this.

mbkumar:
Fixed in #373

PR #373, opened 2023-10-30
Fix to unittest.

ejpaul:
This PR is in response to  #371. Assert statements have now been included in the unit tests for the LPTorsionalStrainPenalty and LPBinormalCurvatureStrainPenalty.


mbkumar (approved):
@ejpaul 
Thanks for the quick fix

PR #377, opened 2023-11-07
Updated CircularCoil class to alllow for dofs

jloizu:
I have (with some help from Jason, thanks!) modified the class CircularCoil in order to define its attributes (r0, I, normal, etc.) as degrees of freedom so that the object can be used in optimization of circular coil fields. This should not prevent a nominal use of the class, namely without optimization.


andrewgiuliani:
we would need unit tests that cover this new functionality

src/simsopt/field/magneticfieldclasses.py
```diff
@@ -359,14 +362,44 @@ def __init__(self, r0=0.1, center=[0, 0, 0], I=5e5/np.pi, normal=[0, 0]):
...
         self.normal = normal
-        if len(normal) == 2:
-            theta = normal[0]
-            phi = normal[1]
-        else:
-            theta = np.arctan2(normal[1], normal[0])
-            phi = np.arctan2(np.sqrt(normal[0]**2+normal[1]**2), normal[2])
 
-        self.rotMatrix = np.array([
+        super().__init__(x0=self.get_dofs(), names=self._make_names(),external_dof_setter=CircularCoil.set_dofs_impl)
```
andrewgiuliani, line 366: Why is `external_dof_setter` needed here?  Typically, I've seen this used when there are dofs in `simsoptpp` that need to be set.  For example, when a Python class does not have dofs located in a cpp counterpart, we've simply had a `set_dofs` function Python-side like below:

https://github.com/hiddenSymmetries/simsopt/blob/4da52143991dd463d1a2e3bbd13a820dc73be338/src/simsopt/geo/surfacexyztensorfourier.py#L87

jloizu:
> we would need unit tests that cover this new functionality.
> 
> Also a broader question is: what about the other magnetic fields in this file (ToroidalField, PoloidalField, DipoleField, etc)? they don't have this added support. Not saying that we need to, but just something that I noticed.

About unit tests, sure, I can try to do that, but you mean creating new tests for this part or running already existing tests? 

About you broader question: yes I agree, we can think of extending this to other magnetic field objects – one step at a time :)

src/simsopt/field/magneticfieldclasses.py
jloizu: Honestly, I am not 100% sure, but if I don't use the external_dof_setter and simply use the set_dofs, then the dofs are defined and updated, but not the class attributes, something that was confusing to me. For example, after declaring a CircularCoil object OBJ, if an optimizer changes the dofs, then OBJ.x shows the correct values, but OBJ.get_dofs() does not show the correct updated values (rather the initial values). Not sure how else to make this work otherwise, but with the external_dof_setter, then everything works fine.

smoniewski:
I tried writing unit tests in #379. 

I also added the output to plot the circular coil. That made me remember that this field is tied to a specific coil, but that coil is not a geometry object. So this coil will not work for any of the regular curve objectives. Perhaps we should have a CircularCoil curve object that gets its field from CircularCoilField. However, I'm not sure that should be in the scope of this pull request.

andrewgiuliani:
> About unit tests, sure, I can try to do that, but you mean creating new tests for this part or running already existing tests?

In the code coverage report, the `set_dofs_impl` function is not covered by the unit tests (you can see this in the codecov/patch check).  It shouldn't be to difficult add a unit test for this, checking that `r0`, `center`, etc have the correct values.

smoniewski:
With no activity for a couple months, I think its about time to complete this pull request. I created a new issue for the discussion of further additions to magneticfieldclasses.py.

smoniewski (dismissed):

jloizu:
Thanks for reviving this, indeed I think we should go ahead with this and then if necessary carry out further additions/improvements to the class. I have been testing it for a while and it looks good to me!

Is it normal that the Linting CI tests fail? can one ignore that?

jloizu:
sorry but I had forgotten to add this minor quick fix that avoids errors when using load/save functions.

landreman (approved):

PR #379, opened 2023-11-13
Js/circularcoil dofs update

smoniewski:
Added for CircularCoil:

- unit tests for optimization
- export to_vtk

My first time trying to write tests, so I would appreciate any feedback.

### DIFF
--- a/src/simsopt/field/magneticfieldclasses.py
+++ b/src/simsopt/field/magneticfieldclasses.py
@@ -533,6 +533,44 @@ class CircularCoil(MagneticField):
         field.set_points_cart(xyz)
         return field
 
+    def gamma(self, points=64):
+        """Export points of the coil."""
+
+        angle_points = np.linspace(0, 2*np.pi, points+1)[:-1]
+
+        x = self.r0 * np.cos(angle_points)
+        y = self.r0 * np.sin(angle_points)
+        z = 0 * angle_points
+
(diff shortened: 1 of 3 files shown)


smiet:
Hi, 

Nice tests! I like how you test against first principles by optimizing the coil position to get a known magnetic field. 

Two suggestions: 
- break up `test_circularcoil_optimization` into smaller tests (`test_circular_coil_current_optimization` and `test_circular_coil_position_optimization` one for the normal optimization. 
- set the tolerance for the position optimization lower (line 87) as this test randomly fails. Random initialisations in testing are dangerous, so you could also consider using a fixed starting position or fixing the seed to the random generator. 

Setting `gtol` for the optimization very low doesn't help as it is already below machine precision (see warning in log). Better to 
`assert np.allclose(coil.x, np.zeros(3), atol=1e-6)`. Probably safe to do for the other `assert`ions as well.

smoniewski:
The position optimization was rather difficult to get within tolerance. It seems the point (0,0,0) was a problem, so now it finds the point (1,1,1) instead and seems to work. 

I copied the random initialization from the tests under geo, and the random position will be between [0,1]. I think that any position in that range shouldn't be a problem, but I can change that if necessary.

jloizu:
Great, thanks for creating these tests!

PR #381, opened 2023-11-14
fix parallel evaluation of derived optimizables

smiet:
fix the derived optimizables `Residue()` and `Quasisymmetry()` so that  `least_squares_mpi_solve` can be run in with more than 1 worker per group. 

code would call Residue.J(), the workers would skip and return nothing whilst the Leader would execute, and hang on expecting communications from its workers. 

Fixed and a rudimentary unit test added.


landreman:
This change to `spec.py` looks right.

If I understand correctly, the change in `boozer.py` does not end up having any effect. Is this your understanding too? Consider any worker process that previously would have hit line 310 (`return np.array([])`) before hitting 313 (`self.boozer.run()`). In the changed `boozer.py`, these worker processes now all call `self.boozer.run()`, but immediately exit that function on line 119, since the logic in line 117 is the same as what is now line 310. So there is no change in behavior. I'm not opposed to the change in `boozer.py`, just want to make sure we are on the same page.

Because there is no change to the behavior of `boozer.py`, the new test `test_Quasisymmetry_paralellization` passes even with the previous version of `boozer.py`. It's fine to add this test for the sake of extra coverage. But did you intend for the test to catch something that was failing before this PR?

Would it make sense to combine `test_derived_optimizables_mpi.py` and `test_integrated_vmec_mpi.py` into a single file? (Maybe `test_integrated_mpi.py` - it should have "mpi" in the filename so the CI knows to run it with multiple procs.)

smiet:
Hi Matt, 

Ah, I hadn't seen that! This means we should also adapt  the Boozer class to make workers participate in the VMEC evaluation. Apologies for this oversight, I just applied the changes I found worked for SPEC without looking deeper after they seemed to work. 

I did find that my changes had an effect: the unittestt that I created would fail for the old code, as would `examples/2_intermediate/QH_fixed_resolution_boozer.py` (with setting line 19 `mpi = MpiPartition(1)` and removing line 42 and 44.

Now I understand why I had to remove line 42: a plain evaluation of `Quasisymmetry.J()` would throw out the workers in line 117, and then let VMEC hang waiting for the workers that never started. 

why my change DID work for me is still a bit of a mystery. I think because the first target of the `least_squares_problem` was vmec.aspect (in example and in test), that would force everyone to participate in the equilibrium calculation each cycle, and then it wouldn't matter that some didn't reach the vmec.run() as it would also be skipped by proc0 who did. 

Good catch, and my latest push should fix this. 

I intentionally created a new file, as `test_integrated_vmec_mpi.py` seemed quite specific, and I didn't want to change a name, but I agree that combining the two into `test_integrated_mpi.py` would be a neater solution. 

Sorry I have to run home right now, so I cannot add that change, but I hope to get to it tomorrow or if you see the chance, by all means and thank you.

smiet:
Hi Matt, 

Sorry for the delay on this, I just merged the two tests into one file, which tests parallel simsopt-vmec integration and parallel evaluation of residues and quasisymmetry. 

This was the major point needed to run SPEC optimizations massively parallel, so it would be a nice addition to master when all tests pass sucessfully.

landreman (approved):

PR #383, opened 2023-11-20
New build system based on `scikit-build-core`

mbkumar:
I replaced the deprecated `setup.py` based build framework with a new build system based on `scikit-build-core`, which in turn uses `hatchling` as its backend. All the project meta-data is now in `pyproject.toml`. `setup.py` and `setup.cfg` are removed. `Pyproject.toml` can also contain options for various tools used in the project. At present, `ruff` configuration is written to it. In the future, I'll add more configurations to it, and remove the configuration files scattered around the project.


mbkumar:
I think this is ready for review.

mbkumar:
The tests are passing, but the coverage data collection is failing for some reason. I am not seeing the behavior exhibited by the runner on my Mac laptop.

.github/workflows/ci.yml
```diff
@@ -88,7 +88,7 @@ jobs:
     - name: Install python dependencies
       run: |
         sudo apt-get install graphviz graphviz-dev
-        pip install wheel numpy scipy f90nml h5py scikit-build cmake qsc sympy pyevtk matplotlib ninja plotly networkx pygraphviz
+        pip install wheel numpy scipy f90nml h5py scikit-build cmake qsc sympy pyevtk matplotlib ninja plotly networkx pygraphviz mpi4py  py_spec pyoculus h5py
```
landreman, line 91: h5py is listed twice here. 

I'm trying to remember why all these packages are manually `pip install`ed here. Wouldn't they automatically be installed when simsopt is `pip install`ed?

.github/workflows/extensive_ci.yml
```diff
@@ -165,15 +165,19 @@ jobs:
...
     - name: Install simsopt package
       if: contains(matrix.packages, 'none')
       run: pip install -v .
 
     - name: Install simsopt package
       if: contains(matrix.packages, 'vmec')
-      run: pip install -v .[MPI]
+      run: |
+        pip install -v .
+        pip install mpi4py
```
landreman, line 180: What wasn't working about the original line here? Doesn't [MPI] work for installing mpi4py?

.github/workflows/extensive_ci.yml
```diff
@@ -165,15 +165,19 @@ jobs:
 
     - name: Install simsopt package
       if: contains(matrix.packages, 'spec') || contains(matrix.packages, 'all')
-      run: pip install -v .[MPI,SPEC]
+      run: |
+        pip install -v .
+        pip install mpi4py py_spec pyoculus h5py
```
landreman, line 170: What wasn't working about the original line here? Doesn't `[MPI,SPEC]` work for installing `mpi4py py_spec pyoculus h5py`?

.github/workflows/ci.yml
mbkumar: I think it was the holdover from the initial days, when we were building wheels first and then installing from the wheels. Still that does not explain the reason. Let me clean it up.

.github/workflows/extensive_ci.yml
mbkumar: For some reason, mpi4py was not getting installed. The runner is acting very peculiar and I am not sure why.

.github/workflows/extensive_ci.yml
mbkumar: Yes. Not sure why.

.github/workflows/ci.yml
mbkumar: I checked it again. We are installing them because SPEC build requires them. SPEC installation setup is stagnant. I implemented a minimal setup when building the Python wrappers. No one updated it afterward.

mbkumar:
Fixed the coverage combine issue. Still not sure why it worked before and not now.

mbkumar:
@landreman This is ready for final review.

landreman (approved):

PR #390, opened 2024-02-04
Fix error with netcdf scalars in scipy v1.12

landreman:
[This change to scipy's netcdf module](https://github.com/scipy/scipy/commit/1bd1607fdbb5d762a7ff7bc851074cb38f829d1c) in scipy v1.12 causes simsopt's netcdf-writing functions to fail when setting scalar (i.e. 0D) values. I'm not sure why that change was made to scipy because `numpy.ndarray.itemset` does not appear to be deprecated. Anyhow, this pull request makes simsopt's netcdf routines work with scipy 1.12, while also working with earlier scipy versions. No changes to the required versions of numpy or scipy for simsopt are needed.

### DIFF
--- a/src/simsopt/field/mgrid.py
+++ b/src/simsopt/field/mgrid.py
@@ -151,16 +151,16 @@ class MGrid():
             var_raw_coil_cur = ds.createVariable('raw_coil_cur', 'f8', ('external_coils',))
 
             # assign values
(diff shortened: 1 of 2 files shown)


mbkumar (approved):
Thanks Matt! Your fix is better.

PR #392, opened 2024-02-10
Version bump for various Actions used in Workflows

mbkumar:
The recommended versions for many Actions used in the workflows have been changed. This PR brings the Actions versions to the latest ones.

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -26,7 +26,7 @@ jobs:
         shell: "bash -l {0}"
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           submodules: true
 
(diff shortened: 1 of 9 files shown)


landreman:
Hmm, in the last commit to the `infra` branch just now, I tried updating more of the github-actions package versions. Now github actions is giving errors "Conflict: an artifact with this name already exists on the workflow run." Does this mean there is a problem with some of the newer versions of the actions? Should we revert some of those changes I just made, or do something else to fix the error?

mbkumar:
There is a warning for upload-artifact and download-artifact actions.
```
Important

upload-artifact@v4+ is not currently supported on GHES yet. If you are on GHES, you must use [v3](https://github.com/actions/upload-artifact/releases/tag/v3).
```
Let me bump down the versions for these two.

landreman (approved):

Issue #357, opened 2023-10-01
SPEC test_set_profile_* sporadically failing


landreman:
The SPEC tests `test_set_profile_cumulative` and `test_set_profile_non_cumulative` are occasionally failing in the CI. Some examples:
https://github.com/hiddenSymmetries/simsopt/actions/runs/6364144339/job/17280224739
https://github.com/hiddenSymmetries/simsopt/actions/runs/6263108068/job/17006759326
Most of the time these tests both pass - usually it is just one of the many jobs in the "extensive CI" that fails. When they fail, the 2 tests seem to fail together, and the error message is
~~~~
======================================================================
ERROR: test_set_profile_cumulative (mhd.test_spec.SpecTests)
Set a SPEC profile of a cumulative quantity (volume current in this example)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/simsopt/simsopt/tests/mhd/test_spec.py", line 178, in test_set_profile_cumulative
    s = Spec(filename)
  File "/opt/hostedtoolcache/Python/3.9.17/x64/lib/python3.9/site-packages/simsopt/mhd/spec.py", line 181, in __init__
    self.initial_guess[lvol].set_rc(mm, nn, self.allglobal.allrzrz[0, lvol, imode])
IndexError: index 2 is out of bounds for axis 1 with size 2

======================================================================
ERROR: test_set_profile_non_cumulative (mhd.test_spec.SpecTests)
Set a SPEC profile of a non-cumulative quantity (surface current in this example)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/simsopt/simsopt/tests/mhd/test_spec.py", line 146, in test_set_profile_non_cumulative
    s = Spec(filename)
  File "/opt/hostedtoolcache/Python/3.9.17/x64/lib/python3.9/site-packages/simsopt/mhd/spec.py", line 181, in __init__
    self.initial_guess[lvol].set_rc(mm, nn, self.allglobal.allrzrz[0, lvol, imode])
IndexError: index 2 is out of bounds for axis 1 with size 2
~~~~
I'm not sure what would cause this error in a non-deterministic way.

These tests both begin with the following code:
~~~~py
        filename = os.path.join(TEST_DIR, 'RotatingEllipse_Nvol8.sp')

        with ScratchDir("."):
            s = Spec(filename)
~~~~
[...]

PR #396, opened 2024-02-27
Mgrid improvements

aaroncbader:
This PR updates the mgrid writing script to also include the arrays for the Magnetic Potential.  These values are needed in some codes (such as BMW).

There is a separate issue in the mgrid to allow for calculation of multiple coil sets separately with scaled currents, but that requires some more thought.


mbkumar:
In the MGrid class, the method `from_file` should read the A_vec data as well.

landreman:
I've never seen mgrid files that contain the vector potential before. Is there some version of xgrid other than the one in stellopt that writes mgrid files containing A? mgrid files are already quite large and storing A makes them twice as big, and vmec doesn't need A, so I'm not sure it makes sense to always include A.

aaroncbader:
> I've never seen mgrid files that contain the vector potential before. Is there some version of xgrid other than the one in stellopt that writes mgrid files containing A? mgrid files are already quite large and storing A makes them twice as big, and vmec doesn't need A, so I'm not sure it makes sense to always include A.

The oak ridge version of MakeGrid includes the vector potential.  It is necessary to run BMW.

aaroncbader:
Will make including potential optional and default to off

aaroncbader:
I would like to update the test that reads an mgrid file to include one with the potential fields enabled.  But I don't know how to regenerate the test mgrid file.

landreman:
It looks like that test file `mgrid.pnas-qa-test-lowres-standard.nc` was added by Tony Qian, so you could ask him. Or just use a different file for testing mgrid files that contain A.

aaroncbader:
Should I be worried about the failed tests?

mbkumar:
@aaroncbader The issue is not with your tests, but one of Spec tests is failing. I am not sure if that test is using the mgrid file that was modified, but I doubt it.

aaroncbader:
There looks to be an issue with loading mgrids. But it's not one that I introduced.  Simsopt is unable to load mgrids that it saves.  It looks like this problem predates my changes.  I'll see if I can fix it.

landreman:
> Should I be worried about the failed tests?

No need to worry - these failures are a known problem, #357, which is independent of this PR. @abaillod @smiet Any hope of fixing #357?

aaroncbader:
ok, I've fixed up the mgrid `from_file` function and added a test for a new mgrid with a potential value

landreman (approved):

PR #399, opened 2024-03-28
compute the rotational transform on a magnetic axis

andrewgiuliani:
This PR introduces functionality to compute the rotational transform on a magnetic axis, using an algorithm shared with me by @landreman and suggested by Stuart Hudson.  I've put this function in `util/magnetic_axis_helpers.py`.

Any suggestions of addition unit tests where the analytical rotational transform is known are welcome.


src/simsopt/util/magnetic_axis_helpers.py
```diff
@@ -0,0 +1,87 @@
...
+        d_Bphi_R_dZ = dBphi_dZ/R
+        
+        A11 = (dBR_dR - (BR / Bphi_R) * d_Bphi_R_dR) / Bphi_R
+        A21 = (dBZ_dR - (BZ / Bphi_R) * d_Bphi_R_dR) / Bphi_R
+        A12 = (dBR_dZ - (BR / Bphi_R) * d_Bphi_R_dZ) / Bphi_R
+        A22 = (dBZ_dZ - (BZ / Bphi_R) * d_Bphi_R_dZ) / Bphi_R
+        A = np.array([[A11,A12],[A21,A22]])
+        return 2*np.pi*np.array([A@x[:2],  A@x[2:]]).flatten()
+    
+    from scipy.integrate import solve_ivp
```
landreman, line 78: Any reason not to put this import at the top of the file?

src/simsopt/util/magnetic_axis_helpers.py
```diff
@@ -0,0 +1,87 @@
...
+
+def compute_on_axis_iota(axis, biotsavart):
+    """
+    Computes the rotational transform on the magnetic axis of a device using a method based on
+    equation (13) of Greene, Journal of Mathematical Physics 20, 1183 (1979); doi: 10.1063/1.524170.
+    This method was shared by Stuart Hudson and Matt Landreman.
+
+    Args:
+        axis: the magnetic axis of a device as a CurveRZFourier.
+        biotsavart: the magnetic field in which the axis lies.
```
landreman, line 12: This function works for any `MagneticField`, not just `BiotSavart`, correct? If so, how about change the variable name from `biotsavart` to something like `field`.

src/simsopt/util/magnetic_axis_helpers.py
```diff
@@ -0,0 +1,87 @@
+import numpy as np
+from simsopt.geo import CurveRZFourier
+
+def compute_on_axis_iota(axis, biotsavart):
+    """
+    Computes the rotational transform on the magnetic axis of a device using a method based on
+    equation (13) of Greene, Journal of Mathematical Physics 20, 1183 (1979); doi: 10.1063/1.524170.
+    This method was shared by Stuart Hudson and Matt Landreman.
+
```
landreman, line 9: This function does not actually check that the provided axis is an axis of the provided field. It would be good to mention this in the docstring.

landreman (commented):
What do you think about putting this function and test in `simsopt.field` rather than `simsopt.util`?

src/simsopt/util/magnetic_axis_helpers.py
andrewgiuliani: no reason, moved import to top

andrewgiuliani:
> What do you think about putting this function and test in `simsopt.field` rather than `simsopt.util`?

I've moved the `magnetic_axis_helpers.py` file to the simsopt.field directory.  This file would be a good place to include a function that actually computes a field's magnetic axis, in an upcoming PR.

landreman (approved):

PR #401, opened 2024-04-04
added surface self-intersection test as well as second derivatives of surface aspect ratio

andrewgiuliani:
This PR introduces a surface self-intersection test, and second derivatives of the surface aspect ratio with respect to the surface degrees of freedom.

The algorithm that checks for surface self-intersection first computes a surface cross section at a given cylindrical angle phi.  Then, approximating the cross section as a piecewise linear polyline, the Bentley-Ottmann algorithm is used to check if any line segments overlap one another.

The second derivatives of surface aspect ratio are useful if being used as a surface label in the BoozerLS algorithm.

### DIFF
--- a/src/simsopt/geo/surface.py
+++ b/src/simsopt/geo/surface.py
@@ -445,8 +445,8 @@ class Surface(Optimizable):
                    surface might not be self-intersecting at a given angle, but may be self-intersecting elsewhere.  To be certain
                    that the surface is not self-intersecting, it is recommended to run this check at multiple angles.  Also note
                    that angle is assumed to be in radians, and not divided by 2*pi as is commonly assumed.
-            thetas: the number of points to compute poloidally in a cross section.  If None, then there will be
-                    surface.quadpoints_theta.size points in the cross section.
+            thetas: the number of uniformly spaced points to compute poloidally in a cross section.  If None, then there will be
+                    surface.quadpoints_theta.size uniformly space points in the cross section.
         Returns:
             True if surface is self-intersecting at angle, else False.
         """
--- a/src/simsopt/geo/surface.py
+++ b/src/simsopt/geo/surface.py
@@ -438,7 +438,8 @@ class Surface(Optimizable):
         r"""
         This function computes a cross section of self at the input cylindrical angle.  Then,
         approximating the cross section as a piecewise linear polyline, the Bentley-Ottmann algorithm
-        is used to check for self-intersecting segments of the cross section.
+        is used to check for self-intersecting segments of the cross section.  NOTE: if this function returns False,
+        this is not a guarantee that the surface is everywhere not self-intersecting.
 
         Args:
             angle: the cylindrical angle at which we would like to check whether the surface is self-intersecting.  Note that a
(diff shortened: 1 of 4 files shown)


landreman:
I think `test_is_self_intersecting` is not being run in the CI, since `ground` and `bentley_ottman` are not installed there. Does it make sense to add these packages to simsopt's requirements? Or do you think these packages and their sub-dependencies are onerous to include for some users? If we don't include these packages in simsopt's general requirements, then I'd suggest manually installing them in the CI and extensive CI as in [this line](https://github.com/hiddenSymmetries/simsopt/blob/master/.github/workflows/extensive_test.yml#L101) so `test_is_self_intersecting` is run.

src/simsopt/geo/surface.py
```diff
@@ -481,6 +534,17 @@ def dminor_radius_by_dcoeff(self):
 
         return (0.5/np.pi)*self.dmean_cross_sectional_area_by_dcoeff()/np.sqrt(self.mean_cross_sectional_area() / np.pi)
 
+    def d2minor_radius_by_dcoeff_dcoeff(self):
+        """
+        Return the second derivative of the minor radius wrt surface coefficients
+
```
VascoSch92, line 540:
```suggestion
```

src/simsopt/geo/surface.py
```diff
@@ -508,6 +572,23 @@ def dmajor_radius_by_dcoeff(self):
...
+        dV_dt = self.dvolume_by_dcoeff()[None, :]
+        d2V_dsdt = self.d2volume_by_dcoeffdcoeff()
+        r = self.minor_radius()
+        dr_ds = self.dminor_radius_by_dcoeff()[:, None]
+        dr_dt = self.dminor_radius_by_dcoeff()[None, :]
+        d2r_dsdt = self.d2minor_radius_by_dcoeff_dcoeff()
+
+        return ((6*V*dr_dt*dr_ds)/r**4 - (2*dV_dt*dr_ds)/r**3 - \
+                (2*dr_dt*dV_ds)/r**3 - (2*V*d2r_dsdt)/r**3 + \
+                d2V_dsdt/r**2) * np.sign(V)/(2*np.pi**2)
```
VascoSch92, line 590:
```suggestion
        return (
            (6*V*dr_dt*dr_ds)/r**4 - (2*dV_dt*dr_ds)/r**3 - 
            (2*dr_dt*dV_ds)/r**3 - (2*V*d2r_dsdt)/r**3 + 
             d2V_dsdt/r**2
             ) * np.sign(V)/(2*np.pi**2)
```

I thought you were missing a parenthesis :-) but it was correct. in this way is a little bit easier to read

src/simsopt/geo/surface.py
```diff
@@ -625,6 +706,84 @@ def dmean_cross_sectional_area_by_dcoeff(self):
...
+        dxvarphi_dt = dg1_ds[:, :, 0, None, :]
+        dyvarphi_dt = dg1_ds[:, :, 1, None, :]
+        dzvarphi_dt = dg1_ds[:, :, 2, None, :]
+
+        dxtheta_dt = dg2_ds[:, :, 0, None, :]
+        dytheta_dt = dg2_ds[:, :, 1, None, :]
+        dztheta_dt = dg2_ds[:, :, 2, None, :]
+
+        mean_area = np.mean((1/r) * (ztheta*(x*yvarphi-y*xvarphi)-zvarphi*(x*ytheta-y*xtheta)))/(2.*np.pi)
+        d2mean_area_ds2 = np.sign(mean_area)*np.mean((2*(-(xvarphi*y*ztheta) + xtheta*y*zvarphi + x*(yvarphi*ztheta - \
```
VascoSch92, line 761: is it not possible to refactor that? because it is super hard to read

landreman (approved):

landreman (approved):

PR #408, opened 2024-04-23
Disable Avx512

mbkumar:
Intel is deprecating AVX512 from their chips. And AMD implementation of AVX512 is a non-native implementation with two 256 bit registers concatenated. The current AVX512 optimization code is not working when compiled with Intel compilers because xsimd is generating 4 doubles for simd_t when Intel compiler is used. However the disabled code is expecting 8 doubles.  We can forego this optimization for more portability.


landreman (approved):

src/simsoptpp/simdhelpers.h
```diff
@@ -129,6 +129,7 @@ using AlignedPaddedVec = std::vector<double, AlignedPaddedAllocator<double>>;
 #endif
 
 #if defined(USE_XSIMD)
+/*
 #if __AVX512F__
```
andrewgiuliani, line 133: If AVX-512 is disabled, shouldn't `__AVX512F__` not be defined, and this section of code would be avoided?

andrewgiuliani:
This is a neat optimization that would be a shame to remove. I can't reproduce the issue locally because I have an 11th generation intel CPU. 

> The current AVX512 optimization code is not working when compiled with Intel compilers because xsimd is generating 4 doubles for simd_t when Intel compiler is used. However the disabled code is expecting 8 doubles. 

I don't understand here, could you give more details?

andrewgiuliani:
thanks for the explanation, I'm ok with merging this PR

PR #409, opened 2024-04-24
vectorized BoozerLS residual evaluation

andrewgiuliani:
This PR introduces two capabilities:

1. calculating the Boozer residual for the BoozerLS algorithm using vectorized operations in cpp.  This is especially important for the Hessian evaluation, since one can quickly run out of memory using the original python implementation.
2. the boozer residual is the integrated squared PDE residual: `\int residual(varphi, theta)^2 dvarphi dtheta`.  This PR allows the user to weight the residual with 1/modB, preventing the residual from scaling with the coil currents, i.e., `\int [residual(varphi, theta)/modB(varphi, theta)]^2 dvarphi dtheta`.

Computing the Hessian for surfaces with mpol, ntor = N, I obtain the following timings in seconds:
```
N, non-vectorized, vectorized
1  0.12            0.06
2  0.82            0.10
3  2.95            0.29
4  7.59            0.67
5  17.56           1.25
6  killed          2.33
7  killed          4.02
8  killed          6.51
9  killed          9.98
```

### DIFF
--- a/src/simsopt/geo/boozersurface.py
+++ b/src/simsopt/geo/boozersurface.py
@@ -41,13 +41,12 @@ class BoozerSurface(Optimizable):
     where Newton is used to solve the first order necessary conditions for optimality.
     """
     
-    def __init__(self, biotsavart, surface, label, targetlabel, constraint_weight=None):
+    def __init__(self, biotsavart, surface, label, targetlabel):
         super().__init__(depends_on=[biotsavart])
         self.biotsavart = biotsavart
         self.surface = surface
         self.label = label
         self.targetlabel = targetlabel
-        self.constraint_weight = constraint_weight
         self.need_to_run_code = True
 
     def recompute_bell(self, parent=None):
--- a/src/simsopt/geo/boozersurface.py
+++ b/src/simsopt/geo/boozersurface.py
@@ -407,7 +407,7 @@ class BoozerSurface(Optimizable):
         self.need_to_run_code = False
         return res
 
-    def minimize_boozer_penalty_constraints_ls(self, tol=1e-12, maxiter=10, constraint_weight=1., iota=0., G=None, method='lm', vectorize=False):
+    def minimize_boozer_penalty_constraints_ls(self, tol=1e-12, maxiter=10, constraint_weight=1., iota=0., G=None, method='lm'):
         """
         This function does the same as :mod:`minimize_boozer_penalty_constraints_LBFGS`, but instead of LBFGS it
         uses a nonlinear least squares algorithm when ``method='lm'``.  Options for the method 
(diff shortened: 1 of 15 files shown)


src/simsopt/geo/boozersurface.py
```diff
@@ -51,7 +52,7 @@ def __init__(self, biotsavart, surface, label, targetlabel):
     def recompute_bell(self, parent=None):
         self.need_to_run_code = True
 
-    def boozer_penalty_constraints(self, x, derivatives=0, constraint_weight=1., scalarize=True, optimize_G=False):
+    def boozer_penalty_constraints(self, x, derivatives=0, constraint_weight=1., scalarize=True, optimize_G=False, weight_inv_modB=False):
```
landreman, line 55: Would it make sense for the default of weight_inv_modB to be True? I would think that option would be more robust.

src/simsopt/geo/surfaceobjectives.py
```diff
@@ -287,21 +287,40 @@ def dJ(self):
         return Derivative({self.surface: deriv})
 
 
-def boozer_surface_residual(surface, iota, G, biotsavart, derivatives=0):
+def boozer_surface_residual(surface, iota, G, biotsavart, derivatives=0, weight_inv_modB=False):
```
landreman, line 290: Consider default True?

landreman (commented):
Looks good. If the new vectorized routine is much better, did you consider removing the old routine? Similarly, it sounds like the new weight_inv_modB option improves robustness, so you could consider making it the only option, so there are fewer cases to maintain. Maybe there are backwards compatibility reasons to keep them, not sure.

src/simsopt/geo/surfaceobjectives.py
andrewgiuliani: I decided against defaulting to `True` here because the BoozerExact algorithm relies on this function, and it assumes that that the residuals are unweighted.

andrewgiuliani:
> If the new vectorized routine is much better, did you consider removing the old routine? 

The algorithms in `BoozerSurface.minimize_boozer_penalty_constraints_ls` rely on the old routine and I preferred to leave it alone, rather than to do a larger refactor of the BoozerSurface class.  Specifically, the old routines can return the residual evaluated at all collocation points (a vector) or the sum of all residuals squared (a scalar).  The new routines only return the scalarized residual, while `BoozerSurface.minimize_boozer_penalty_constraints_ls` require the vector residual.

> Similarly, it sounds like the new weight_inv_modB option improves robustness, so you could consider making it the only option, so there are fewer cases to maintain.

The BoozerExact algorithm assumes that the residuals are unweighted, so I think we should maintain both options.

landreman (approved):

PR #410, opened 2024-04-25
Rotating ellipse surface

landreman:
This PR adds a convenience function to make `SurfaceRZFourier` surfaces that are rotating ellipses.

I decided not to make this a `@classmethod` so users have the option to specify either `quadpoints_theta` and `quadpoints_phi`, or specify `range`, `nphi`, and `ntheta` with `@classmethod from_nphi_ntheta()`.

### DIFF
--- a/tests/geo/test_surface_rzfourier.py
+++ b/tests/geo/test_surface_rzfourier.py
@@ -712,7 +712,7 @@ class SurfaceRZFourierTests(unittest.TestCase):
             eq.indata.ntor = 4
             eq.indata.ftol_array[:2] = [1e-8, 1e-10]
 
-            # Try the case of elongation=1 with positive torsion:
+            # Try the case of elongation=1 with positive axis torsion:
             major_radius = 8.4
             minor_radius = 1.3
             elongation = 1.0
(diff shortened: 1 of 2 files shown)


tests/geo/test_surface_rzfourier.py
```diff
@@ -631,6 +638,98 @@ def test_shared_dof_serialization(self):
...
+            np.testing.assert_array_less(0, eq.wout.iotaf)
+
+            # Try the case of zero torsion with rotating elongation:
+            major_radius = 8.4
+            minor_radius = 1.3
+            elongation = 2.1
+            torsion = 0.0
+            eq.boundary.make_rotating_ellipse(major_radius, minor_radius, elongation, torsion)
+            eq.run()
+            np.testing.assert_array_less(0, eq.wout.iotaf)
```
rogeriojorge, line 731: What about a test to assert that eq.wout.iotaf is close to a specific value? This way, in the future, if somebody changes this function to have unexpected behavior, it provides a stronger test than simply a non-zero iota.

rogeriojorge (approved):
Looks good to me, just a comment on the possibility of an extra iota test.

tests/geo/test_surface_rzfourier.py
landreman: Done.

PR #412, opened 2024-05-02
Fix #411 compile flags on Apple Silicon Ubuntu Docker

### LINKED ISSUE #411: Compilation on arm64 Ubuntu-22.04 Docker on Apple M1 fails
`gcc` doesn't support `-mfma` flag on `arm64` CPUs, leading to compile error
    
    c++: error: unrecognized command-line option ‘-mfma’

on `pip install`.

Fix: Remove `-mfma` compile flags in `CMakeLists.txt`

krystophny:


krystophny:
@andrewgiuliani or @landreman - can you verify if this is fine? A cleaner solution would be a systematic check for compiler options, but the present fix at least makes it compile in our setup.

CMakeLists.txt
```diff
@@ -56,9 +56,14 @@ else()
...
-    elseif(${CMAKE_HOST_SYSTEM_PROCESSOR} STREQUAL "arm64")  
-    	set(CMAKE_CXX_FLAGS "-O3 -mcpu=apple-a14 -mfma -ffp-contract=fast")
+        message(STATUS "${CMAKE_HOST_SYSTEM_PROCESSOR}")
+        if(${CMAKE_HOST_SYSTEM_PROCESSOR} STREQUAL "aarch64")
+    	    set(CMAKE_CXX_FLAGS "-O3 -march=native -ffp-contract=fast")
+        else()
+            set(CMAKE_CXX_FLAGS "-O3 -march=native -mfma -ffp-contract=fast")
+        endif()
+    elseif(${CMAKE_HOST_SYSTEM_PROCESSOR} STREQUAL "arm64")
+    	set(CMAKE_CXX_FLAGS "-O3 -mcpu=apple-a14 -ffp-contract=fast")
```
andrewgiuliani, line 66: why are there two branches here, one for `aarch64` and another for `arm64`?

mbkumar:
Let me cross check this with what I got for docker build on Mac. Can we
also get the docker file uploaded?

I'll review this next week after I get back to work.

Bharat

CMakeLists.txt
krystophny: yeah, this is confusing because "arm64" is given on macos, and "aarch64" on linux on the same CPU.

krystophny:
Btw the Docker file we are using is `ghcr.io/itpplasma/devel-tex` based on Debian 12.

CMakeLists.txt
mbkumar: The clang compiler on Mac (v15.x) now supports march=native. So we need to change the conditions to reflect this. I'll adjust the conditions and push them.

mbkumar: I'd recommend to go with the below if else conditions. 
```
if(COMPILER_SUPPORTS_MARCH_NATIVE)
    set(CMAKE_CXX_FLAGS "-O3 -march=native -ffp-contract=fast")
elseif(${CMAKE_HOST_SYSTEM_PROCESSOR} STREQUAL "arm64")  
    set(CMAKE_CXX_FLAGS "-O3 -mcpu=apple-a14 -ffp-contract=fast")
```

krystophny: @mbkumar I did a few more changes as you hinted. I kept the -mfma option in the “old clang” branch but skipped it otherwise where it should be anyway added due to the other flags. Also, there’s a fallback case in the end. Can we merge this like it is now?

krystophny:
@mbkumar @andrewgiuliani can you quickly check? This is such a small change that we shouldn't spend more weeks on it.

CMakeLists.txt
```diff
@@ -56,11 +56,11 @@ else()
     unset(COMPILER_SUPPORTS_MARCH_NATIVE CACHE)
     CHECK_CXX_COMPILER_FLAG(-march=native COMPILER_SUPPORTS_MARCH_NATIVE)
     if(COMPILER_SUPPORTS_MARCH_NATIVE)
-    	set(CMAKE_CXX_FLAGS "-O3 -march=native -mfma -ffp-contract=fast")
-    elseif(${CMAKE_HOST_SYSTEM_PROCESSOR} STREQUAL "arm64")  
-    	set(CMAKE_CXX_FLAGS "-O3 -mcpu=apple-a14 -mfma -ffp-contract=fast")
+        set(CMAKE_CXX_FLAGS "-O3 -march=native -ffp-contract=fast")
+    elseif(${CMAKE_HOST_SYSTEM_PROCESSOR} STREQUAL "arm64")
+        set(CMAKE_CXX_FLAGS "-O3 -mcpu=apple-a14 -mfma -ffp-contract=fast")
```
mbkumar, line 61: Can you please remove `-mfma` on line 61 as well?

mbkumar (changes requested):

CMakeLists.txt
krystophny: Done!

krystophny:
Thanks! I updated the branch - I think the problem with the CI/CD is because the external fork has no password for the docker registry.

krystophny:
Thanks a lot!

PR #413, opened 2024-05-02
Make residuals file optional

landreman:
When running optimizations for least-squares problems, the vector of residuals is saved at each iteration. For objectives like quasisymmetry with lots of residuals, this file can get quite large. So this PR makes the residuals file optional.


rogeriojorge (commented):
Looks good to me!
In the future, it might be worth it to add a flag to save the 'jac_log' file as well.

PR #414, opened 2024-05-14
Stellarator non-symmetric configurations in vmec_splines

daringli:
Adds support for stellarator non-symmetric configurations in vmec_splines. A notable example of such a configuration is ITER, where the lower-null divertor breaks stellarator symmetry. 
For stellarator symmetric configuration, the additional outputs are all zero.


daringli:
Your commit looks like a strict improvement. Thank you for making these changes.
The X, Y, Z, etc. outputs from vmec_compute_geometry had snuck in from a previous edit that had nothing to do with this. Thanks for removing them.

landreman (approved):

PR #417, opened 2024-05-29
Added 1 FileNotFoundError exception handling to vmec.py

daringli:
Adds try-except blocks around a bare os.remove() call in vmec.py.

For some reason, the os.remove(filename) changed in this pull request gave me an unhandeled exception today. I've never seen it before and I don't know why it happened.
Since the file is supposed to get deleted anyway, it seems harmless to just print a warning and move on. It is also consistent with how os.remove is used elsewhere in vmec.py.


src/simsopt/mhd/vmec.py
```diff
@@ -752,7 +752,11 @@ def run(self):
 
             # Delete the previous output file, if desired:
             for filename in self.files_to_delete:
-                os.remove(filename)
+                try:
```
VascoSch92, line 755: you can also check if the file exists and if it is not the case you can thrown an error.

daringli: The way I've done it is similar to how it's done elsewhere in the same script, and it seems more pythonic to ask for forgiveness rather than permission.
There may also be practical reasons for doing it this way: the file probably ceased to exist just before os.remove tried to remove it (if it didn't exist recently, it'd not be in the files_to_delete list), so checking if it exists beforehand may not be reliable, since it may get deleted after the check is done but before the remove() is done, which is probably what happened here.

src/simsopt/mhd/vmec.py
```diff
@@ -752,7 +752,11 @@ def run(self):
...
+                    os.remove(filename)
+                except FileNotFoundError:
+                    logger.debug('Tried to delete the file {filename} but it was not found')
```
VascoSch92, line 758: I think you forgot an f-string here ;)

daringli: Definitely! Thanks a ton. I committed a fix.

landreman (approved):

PR #418, opened 2024-05-30
Spec refactoring/improvement (PR 1 of 3 for single-stage) 

smiet:
Updates to Spec and NormalField and a few helper functions . 

Spec class has been updated with the following features:
- detection when free-boundary solution is not converged, re-set boundary field guess when detected.
- default free-boundary configuration and classmethod to generate.
- re-written loops pythonically where possible (`for vol in range(self.inputlist.mvol+1)` -> `for surface in self.initial_guess`)
- implemented `array_translator` to reduce looping over index counters. 
- activate_profile helper function.

The NormalField class has been augmented to: 
- Always include the boundary on which it is defined (reference to associated SPEC boundary object).
- Store the Fourier components that describe it in array-form. 
- include getters and setters for the FCs in array form. 
- set the dofs in their entirety even when setting single (fixes bug where changing a DOF did not trigger recompute bell).

The Surface base class has two new methods: `fourier_transform_field` and `inverse_fourier_transform_field` to efficiently calculate fourier transforms in the VMEC convention (-`ntor` to `ntor` toroidal components, `0`:`mpol` poloidal). Allows for different Fourier normalization conventions through the 'normalization' parameter. 

This is PR 1 of 3, the following two: 
- CoilSet and ReducedCoilSet: helper classes to enable interchanging of coils, and create new optimization variables through SVD on coil DOFS
- CoilNormalField and misc changes:  class inheriting form NormalField that interfaces to the CoilSets, all other misc changes needed to make the single-stage optimizations work.


tests/geo/test_surface_rzfourier.py
```diff
@@ -732,6 +732,48 @@ def test_make_rotating_ellipse_iota(self):
...
+
+        # Transform the field to Fourier space:
+        ft_sines, ft_cosines = s.fourier_transform_field(field, stellsym=False)
+        self.assertAlmostEqual(ft_sines[2, 3+s.ntor], 0.8)
+        self.assertAlmostEqual(ft_sines[4, 5+s.ntor], 0.2)
+        self.assertAlmostEqual(ft_cosines[3, 3+s.ntor], 0.7)
+
+        # Test that all other elements are close to zero
+        sines_mask = np.ones_like(ft_sines, dtype=bool)
+        cosines_mask = sines_mask
```
landreman, line 765: You want a copy() here I think. Presently cosines_mask and sines_mask are two names for the same single object, so setting an entry in one modifies both.

tests/mhd/test_spec.py
```diff
@@ -195,6 +195,54 @@ def test_set_profile_cumulative(self):
...
+    def test_array_translator(self):
+        """
+        Test the array_translator method to convert a SPEC style array to simsopt style array.
+        """
+        spec = Spec()
+        array = spec.inputlist.rbc
+        translator = spec.array_translator(array)
+        translator2 = spec.array_translator(array, style='spec')
+        translator3 = spec.array_translator(translator.as_simsopt, style='simsopt')
+        #assert that translator.as_spec i
```
landreman, line 236: Did this sentence get cut off?

src/simsopt/mhd/spec.py
```diff
@@ -936,6 +1071,7 @@ def run(self, update_guess: bool = True):
                 logger.debug('About to call check_inputs')
                 spec.allglobal.check_inputs()
             logger.debug('About to call broadcast_inputs')
+            self.mpi.comm_groups.Barrier()  # wait for
```
landreman, line 1074: Sentence cut off

src/simsopt/field/normal_field.py
```diff
@@ -47,44 +53,52 @@ def __init__(self, nfp=1, stellsym=True, mpol=1, ntor=0,
...
             vns = np.zeros((self.mpol + 1, 2 * self.ntor + 1))
 
-        if vnc is None and not stellsym:
+        if not self.stellsym and vnc is None:
             vnc = np.zeros((self.mpol + 1, 2 * self.ntor + 1))
+        
+        if computational_boundary is None:
+            from simsopt.geo import SurfaceRZFourier
+            computational_boundary = SurfaceRZFourier(nfp=nfp, stellsym=stellsym, mpol=mpol, ntor=ntor)
+        computational_boundary.fix_all()
```
landreman, line 62: This seems like potentially dangerous and undocumented behavior. There might be use cases for NormalField unrelated to spec (maybe QFM surfaces?) where one would want the surface shape to change.

src/simsopt/field/normal_field.py
```diff
@@ -37,7 +43,7 @@ class NormalField(Optimizable):
     """
 
     def __init__(self, nfp=1, stellsym=True, mpol=1, ntor=0,
-                 vns=None, vnc=None):
+                 vns=None, vnc=None, computational_boundary=None):
```
landreman, line 46: The name `computational_boundary` sounds unwieldy. Why not `surface`? There could be use cases for NormalField unrelated to spec for which this surface is not a boundary of anything.

src/simsopt/geo/surface.py
```diff
@@ -839,6 +839,130 @@ def interpolate_on_arclength_grid(self, function, theta_evaluate):
...
+            nmin = -ntor
+            if m == 0: nmin = 1
+            for n in range(nmin, ntor+1):
+                angle = m * theta2d - n * self.nfp * phi2d
+                sinangle = np.sin(angle)
+                if not stellsym:
+                    cosangle = np.cos(angle)
+                field = field + A_mns[m, n + ntor] * sinangle
+                if not stellsym:
+                    field = field + A_mnc[m, n + ntor] * cosangle
```
landreman, line 957: The two `if` statements can be combined into one.

landreman (commented):
Having the new Fourier transform and inverse transform functions go in the base Surface class feels awkward to me. These functions need mpol and ntor, which not all Surface classes have (e.g. SurfaceHenneberg, SurfaceGarabedian). Also the transform and inverse-transform functions don't use any of the surface geometry, just the grid points. So I would suggest moving those two functions to `simsopt.util` (probably not in a class, just standalone functions).

It also feels awkward for NormalField to carry around `computational_boundary`. See the comments attached to specific line numbers for some specific issues. 

Codecov is flagging a lot of lines not covered by tests, 71% coverage for the changes compared to 91% for simsopt overall.

src/simsopt/geo/surface.py
```diff
@@ -839,6 +839,130 @@ def interpolate_on_arclength_grid(self, function, theta_evaluate):
 
         return function_interpolated
 
+    def fourier_transform_field(self, field, mpol=None, ntor=None, normalization=None, **kwargs):
```
andrewgiuliani, line 842: one suggestion to simplify this function is to use a prewritten FFT and inverse FFT algorithm here

smiet:
I have adapted most changes and would like to move the review forward. 

Almost all new untested lines are now covered. 

Especially awaiting comments including a `SurfaceRZFourier` object with the `NormalField`. I think it is useful to have here, but especially in the next PR, where I create a `CoilNormalField` that inherits from `NormalField` it will involve many more functions that have to access the `Surface` (to calculate the coefficients etc). 

Adapted `NormalField.from_spec` method to generate a `SurfaceRZFourier` class from the SPEC data. 

P.S. I might have accidentally 'started a review' myself, my apologies.

landreman (commented):
Hi @smiet, I see the testing was expanded but some of the issues above are still open. The main one is the location of the Fourier transform routines - I still think it doesn't make sense for them to be in the base Surface class, for the reasons explained above. There are also the issues about explaining that "field" is not a magnetic field vector, `cosines_mask = sines_mask`, and combining `if` statements

smiet (commented):
Hi Matt, 

I am sorry, had entered many comments and discussion that are <<pending>> (aparently not for a response from you, but not submitted and hence not visible to you!

My apologies, submitting these now and I will put the imports at the top of the file where requested.

tests/mhd/test_spec.py
```diff
@@ -196,6 +196,96 @@ def test_set_profile_cumulative(self):
...
+                else: 
+                    raise ValueError(f"Profile {profile} not recognized")
+
+
+    def test_freeboundary_default(self):
+        """
+        test the default freeboundary file, and re-set if Picard
+        iteration fails
+        """
+        from simsopt._core.util import ObjectiveFailure
```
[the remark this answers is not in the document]
smiet: done

src/simsopt/field/normal_field.py
```diff
@@ -91,6 +106,7 @@ def from_spec(cls, filename):
         if py_spec is None:
             raise RuntimeError(
                 "Initialization from Spec requires py_spec to be installed.")
+        from simsopt.geo import SurfaceRZFourier
```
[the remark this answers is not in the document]
smiet: done

src/simsopt/field/normal_field.py
```diff
@@ -47,39 +56,45 @@ def __init__(self, nfp=1, stellsym=True, mpol=1, ntor=0,
         if vns is None:
             vns = np.zeros((self.mpol + 1, 2 * self.ntor + 1))
 
-        if vnc is None and not stellsym:
+        if not self.stellsym and vnc is None:
             vnc = np.zeros((self.mpol + 1, 2 * self.ntor + 1))
+        
+        if surface is None:
+            from simsopt.geo import SurfaceRZFourier
```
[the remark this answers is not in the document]
smiet: done

tests/geo/test_surface_rzfourier.py
```diff
@@ -732,6 +732,44 @@ def test_make_rotating_ellipse_iota(self):
...
+        s.zs[1, 0] = 0.2
+
+        # Create the grid of quadpoints:
+        phi2d, theta2d = np.meshgrid(2 * np.pi * s.quadpoints_phi,
+                                     2 * np.pi * s.quadpoints_theta, 
+                                     indexing='ij')
+
+        # create a test field where only Fourier elements [m=2, n=3] 
+        # and [m=4,n=5] are nonzero:
+        field = field = 0.8 * np.sin(2*theta2d - 3*s.nfp*phi2d) + 0.2*np.sin(4*theta2d - 5*s.nfp*phi2d)+ 0.7*np.cos(3*theta2d - 3*s.nfp*phi2d)
```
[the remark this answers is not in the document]
smiet: corrected

mbkumar:
@smiet 
Can you please take a look at the SPEC failure [here](https://github.com/hiddenSymmetries/simsopt/actions/runs/9699158664/job/26767452881)?

smiet:
I had forgotten to include the python 3.8 exception to the numpy 2.0 build-time dependency in the python wrapper. Just merged the commit into SPEC master right before you triggered the tests. 

Your commit just now should trigger the tests which will pull the update and they should now pass

smiet:
@landreman, @mbkumar, could this PR be moved forward? 

Because the discussion is mixed in with a lot of changed code, I am summarizing the main open threads: 

#NormalField with a reference to the surface* 
@landreman noted: 
> It also feels awkward for NormalField to carry around computational_boundary.
I believe this is necessary for the user, as just the Fourier components are meaningless without the surface. I have also added methods as NormalField.get_real_space_field() that performs the Fourier transform (using the surface properties) and is useful for plotting and analyzing the real space field (and passing to a Focus-like algorithm if you want to generate coils for your Spec). My insistence on this is also because of future development in PR3, where I will introduce a CoilNormalField that inherits from it, and requires the Surface. This way the two classes will have more attributes in common. 

@landreman noted:
> Having the new Fourier transform and inverse transform functions go in the base Surface class feels awkward to me. These functions need mpol and ntor, which not all Surface classes have (e.g. SurfaceHenneberg, SurfaceGarabedian). Also the transform and inverse-transform functions don't use any of the surface geometry, just the grid points. So I would suggest moving those two functions to simsopt.util (probably not in a class, just standalone functions).

I agree that having the transform in the `Surface` base class is not optimal, but I do think it fits well in the `SurfaceRZFourier`, as it uses the same VMEC Fourier convention. As a user who has a SurfaceRZFourier, and evaluates some quantity on it, it is only natural for the surface to provide the method to transform it into the same representation as it uses itself. Having a separate function in `util` that you pass both a `Surface` and the quantity you want transformed seems unnatural to me. 
I have moved the function to the SurfaceRZFourier class
If you still prefer `util`, let me know and I will make the change

@andrewgiuliani  noted:
> one suggestion to simplify this function is to use a prewritten FFT and inverse FFT algorithm here
I tried this, but failed, the current code works and is clear (though not optimally efficient), hope it is OK.

landreman:
Presently, the `fourier_transform_scalar` and `inverse_fourier_transform_scalar` methods exist twice each, in both `Surface` and `SurfaceRZFourier`. Had you intended to remove them from `Surface`? Having these functions only in `SurfaceRZFourier` would be ok.

andrewgiuliani (changes requested):
Hi Chris, there are lines missing from code coverage and a merge conflict that needs to be resolved too.

smiet:
I have covered the new and functionally significant added lines of code. The patch has increased coverage over the old state of the SPEC code (though still under target, sorry). Uncovered are mainly `raise` statements and specific cases regarding the running of nonstellaratorsymmetric equilibria.

I hope to cover those in the future, but it will take time to generate a suitable nonstellaratorsymmetric spec file that runs in a reasonable time (for now we only test readin, the nonstellaratorsymmetric case takes cpu-hours to run and is unsitbable). 

With the increased coverage, passing tests, and removed duplicate transform method, I hope we are good to merge! 

Thank you all for your input and suggestions and time reviewing.

andrewgiuliani:
Here's an example unit test for exceptions, it's not a massive change and it makes sure no lines of code in production are untested:

https://github.com/hiddenSymmetries/simsopt/blob/2c0f05790272705fb6e07efa11bfc62bfd80b978/tests/geo/test_curve.py#L166

landreman (approved):

smiet:
I added the tests to the raises in NormalField. I will be doing some further additions during my next PR's is it OK if I catch the straggling uncovered exceptions in my next PR?
 (the push of these extra tests undid the approval, sorry @landreman)

landreman (approved):

PR #419, opened 2024-06-02
L_{grad B} scale length

landreman:
This pull request adds a calculation of the grad B tensor and L_{grad B} scale length from Kappel et al, "The Magnetic Gradient Scale Length Explains Why Certain Plasmas Require Close External Magnetic Coils", Plasma Phys. Control. Fusion 66 (2024) 025018,  https://doi.org/10.1088/1361-6587/ad1a3e


johntkappel (approved):
I have looked over the code related to Kappel et al, PPCF 66 025018 (2024), the L_grad_B term. The code accurately implements the code used in that paper. The code also adds additional tests to ensure the code works. These tests seem sensible and are either standard checks, or relate to the physics of L_grad_B (namely the test_L_grad_B_axisymmetry).

Although there are some failures in the tests/checks in Github actions, these are unrelated to any of this code added here (Matt has expressed to me that is related to the external library Numpy, which was updated to 2.0 on June 16th - 2 days ago - and that these issues need to be fixed within the main branch).

Given this, I think this branch should be merged into master.

PR #425, opened 2024-06-12
added a QUASR-downloader

smiet:
It would be amazing to have all of the QUASR configurations accessible with just one command. 

I implemented a simple interface using the 'requests' library, that downloads any QUASR field. 

I tried to make it so that it picks a random configuration, but apparently not all integers between 0 and 300,000 correspond to an entry in the database. 
@andrewgiuliani is there a list of which entries exist, so that a robust random picker can be returned? 
Or even better, the user could directly query the database for a list of constraints (NFP, A, iota, coil length, etc) and get a random config that fits their needs. 

Something like this must be already implemented on the website, one would only have to adapt this to python. I am happy to do this but would need more info on the database itself. Would this be of interest @andrewgiuliani?

### DIFF
--- a/src/simsopt/configs/zoo.py
+++ b/src/simsopt/configs/zoo.py
@@ -187,3 +187,52 @@ def get_w7x_data(Nt_coils=48, Nt_ma=10, ppp=2):
     ma.zs[:] = sZ[0:Nt_ma]
     ma.x = ma.get_dofs()
     return (curves, currents, ma)
+
+def get_QUASR_data(config_ID:int=None, return_style='default'): 
+    """
+    download a configuration from the QUASAR database
+
+    Args:
+        config_ID: the ID of the configuration to download. If None, a random configuration will be downloaded
+        return_style: 'default' or 'coils. If 'default', the function will return the curves, currents and magnetic axis
+         like the other configurations in the zoo. 
+         If 'coils', the function will return coils, the magnetic axis and a list of flux surfaces.
+
+    """
+    import json
+    from simsopt._core.json import GSONDecoder
+    import numpy as np
+    #test if you have the requests library: 
+    try:
+        import requests
+    except ImportError:
+        raise ImportError("You need to install the requests library to use this function. Run 'pip install requests'")
+    
+    if config_ID is None: 
+        # a random number between 1 and 300000:
+        config_ID = np.random.randint(1,300000)
+    
+    id_str = f"{config_ID:07d}"
+    
+    # string format randint to 7 digits
+    url = f'https://quasr.flatironinstitute.org/simsopt_serials/{id_str[0:4]}/serial{id_str}.json'
+
+    with requests.get(url) as r:
+        if r.status_code == 200:
+            print(f"Configuration with ID {config_ID} downloaded successfully")
+            surfaces, ma, coils = json.loads(r.content, cls=GSONDecoder)
+        else:
+            print(url)
+            raise ValueError(f"Could not download configuration with ID {config_ID:07d}. Status code: {r.status_code}")
+        
+    if return_style == 'default':
+        curves = [coil.curve for coil in coils]
+        currents = [coil.current for coil in coils]
+        return curves, currents, ma
+    elif return_style == 'coils':
+        return coils, ma, surfaces
+    else:
+        raise ValueError("return_style must be either 'default' or 'coils'")
+
+        
+    


src/simsopt/configs/zoo.py
```diff
@@ -187,3 +195,44 @@ def get_w7x_data(Nt_coils=48, Nt_ma=10, ppp=2):
...
+        nfp = ma.nfp
+        nc_per_hp = len(coils) // nfp // (1 + ma.stellsym)
+        coils = coils[:nc_per_hp]
+
+        curves = [coil.curve for coil in coils]
+        currents = [coil.current for coil in coils]
+        return curves, currents, ma
+
+    elif return_style == 'json':
+        return coils, ma, surfaces
```
landreman, line 238: Add `else:` case to raise an exception if `return_style` is something else

src/simsopt/configs/zoo.py
```diff
@@ -187,3 +195,44 @@ def get_w7x_data(Nt_coils=48, Nt_ma=10, ppp=2):
     ma.zs[:] = sZ[0:Nt_ma]
     ma.x = ma.get_dofs()
     return (curves, currents, ma)
+
+
+@SimsoptRequires(requests is not None, "You need to install the requests library to use this function. Run 'pip install requests'")
+def get_QUASR_data(ID, return_style='default'): 
+    """
+    Download a configuration from the QUASAR database.
```
landreman, line 203: Spelling: QUASR

src/simsopt/configs/zoo.py
```diff
@@ -187,3 +195,44 @@ def get_w7x_data(Nt_coils=48, Nt_ma=10, ppp=2):
...
+
+    Args:
+        ID: the ID of the configuration to download.  The database is navigatable at https://quasr.flatironinstitute.org/
+            Alternatively, you can download the latest full set of devices from https://zenodo.org/doi/10.5281/zenodo.10050655
+        
+        return_style: 'default' or 'json'. If 'default', the function will return the curves, currents and magnetic axis
+                      like the other configurations in the zoo. 
+                      If 'json' the function will return the full set of coils, magnetic axis and a number of surfaces
```
landreman, line 211: By "a number of surfaces", do you specifically mean a list of SurfaceXYZTensorFourier objects? If so it would be preferable to give this specific information.

tests/configs/test_zoo.py
```diff
@@ -0,0 +1,16 @@
+import unittest
+from simsopt.configs import get_QUASR_data
+
+class ZooTests(unittest.TestCase):
+    def test_QUASR_downloader(self):
+        curves, currents, ma = get_QUASR_data(952)
+        coils, ma, surfaces = get_QUASR_data(952, return_style='json')
+
```
landreman, line 8: How about also covering the case of a success with return_style="default".

src/simsopt/configs/zoo.py
```diff
@@ -2,11 +2,19 @@
 from simsopt.geo.curverzfourier import CurveRZFourier
 from simsopt.geo.curvexyzfourier import CurveXYZFourier
 from simsopt.field.coil import Current
+from simsopt._core.dev import SimsoptRequires
+from simsopt._core.json import GSONDecoder
+import json
+
+try:
+    import requests
```
landreman, line 10: Is "requests" already installed in the CI, or does it need to be added so the test will run?

src/simsopt/configs/zoo.py
```diff
@@ -187,3 +195,44 @@ def get_w7x_data(Nt_coils=48, Nt_ma=10, ppp=2):
...
+@SimsoptRequires(requests is not None, "You need to install the requests library to use this function. Run 'pip install requests'")
+def get_QUASR_data(ID, return_style='default'): 
+    """
+    Download a configuration from the QUASAR database.
+
+    Args:
+        ID: the ID of the configuration to download.  The database is navigatable at https://quasr.flatironinstitute.org/
+            Alternatively, you can download the latest full set of devices from https://zenodo.org/doi/10.5281/zenodo.10050655
+        
+        return_style: 'default' or 'json'. If 'default', the function will return the curves, currents and magnetic axis
```
landreman, line 209: The name `"json"` isn't an intuitive description of what the function returns. Maybe `"surfaces"`?

landreman (approved):

andrewgiuliani (changes requested):
I don't think this PR is ready yet.  We still need to rerun code coverage to make sure no lines are missing, and address @landreman's comments.

src/simsopt/configs/zoo.py
```diff
@@ -187,3 +195,44 @@ def get_w7x_data(Nt_coils=48, Nt_ma=10, ppp=2):
...
+                      like the other configurations in the zoo. 
+                      If 'json' the function will return the full set of coils, magnetic axis and a number of surfaces
+                      parametrized in Boozer coordinates.
+    """
+    
+    assert return_style in ['default', 'json']
```
jsoules, line 215: Assertions can be disabled by caller/system config [e.g. if `PYTHONOPTIMIZE` is set], so it might be better not to rely on `assert` for a true runtime (not-debugging) check.

src/simsopt/configs/zoo.py
```diff
@@ -187,3 +195,44 @@ def get_w7x_data(Nt_coils=48, Nt_ma=10, ppp=2):
...
+    
+    id_str = f"{ID:07d}"
+    # string to 7 digits
+    url = f'https://quasr.flatironinstitute.org/simsopt_serials/{id_str[0:4]}/serial{id_str}.json'
```
jsoules, line 219: This is fine, but note that we reserve the right to change the ID structure in new versions of the database, which would break this code. (Fair warning!)

I might be able to set up an endpoint on QUASR for converting an ID (arbitrary length) to the correct location, but I wouldn't have a chance to do that for at least a few weeks.

src/simsopt/configs/zoo.py
```diff
@@ -187,3 +195,44 @@ def get_w7x_data(Nt_coils=48, Nt_ma=10, ppp=2):
...
+
+    with requests.get(url) as r:
+        if r.status_code == 200:
+            print(f"Configuration with ID {ID:07} downloaded successfully")
+            surfaces, ma, coils = json.loads(r.content, cls=GSONDecoder)
+        else:
+            raise ValueError(f"Configuration ID {ID:07d} does not exist. Status code: {r.status_code}")
```
jsoules, line 226: This isn't strictly accurate--the download might've failed for some other reason.

tests/configs/test_zoo.py
```diff
@@ -0,0 +1,16 @@
+import unittest
+from simsopt.configs import get_QUASR_data
+
+class ZooTests(unittest.TestCase):
```
jsoules, line 4: Just noting that, as you're hitting an external resource (QUASR website), this isn't strictly a unit test but an integration test. Now, an integration test may be what you actually care about, but there's always the risk that it will fail (or start failing) due to factors that don't indicate an error in your code.
I would consider mocking the `requests.get` call for regular testing, and have a separate integration test that's run less frequently.

tests/configs/test_zoo.py
```diff
@@ -0,0 +1,16 @@
...
+    def test_QUASR_downloader(self):
+        curves, currents, ma = get_QUASR_data(952)
+        coils, ma, surfaces = get_QUASR_data(952, return_style='json')
```
jsoules, line 7: Might also be useful to have assertions verifying (from the returned data) that these are the same device record.

src/simsopt/configs/zoo.py
jsoules: Agreed--it also looks like "default" is only returning one half-period? Am I reading that right?

smiet:
Sorry to leave this hanging so long, but I think this would be a great addition, and am wanting to use this as I am looking at more QUASRs with pyoculus, and it would be great to directly grab them from a script. 

@jsoules, has the content of the json changed with newer entries to the database? For example https://quasr.flatironinstitute.org/model/1258083 only returns surfaces and coils


@andrewgiuliani, lets divide the work, it should be pretty minimal, hard-code a few numbers into the tests and adapt to the new format of QUASR (no magnetic axis anymore :'( )

jsoules:
> @jsoules, has the content of the json changed with newer entries to the database? For example https://quasr.flatironinstitute.org/model/1258083 only returns surfaces and coils

Hi,
Not sure what you mean--I'm looking at that link and I see it populating with surfaces, coils, modB, and currents, as well as populating the Poincare plots. (All those assets are present on the site back-end.)

The data in the individual-record JSON is an extract of the relevant row for the whole-database JSON; this is what populates the "Device Metadata" block on the right-hand side in the web UI.

The set of fields included in the JSON (i.e. the quantities of interest) have changed in different releases, but without knowing the time point you're comparing to, I'm hesitant to give a definitive answer. But any changes would affect all devices, as the same format is used for the entire database.

smiet:
When we wrote the scripts in June, the database would return a json file containing `surfaces, magnetic_axis, coils`, which allowed us to mirror the return style of other configurations in the simsopt `zoo` such as `simsopt.configs.get_w7x_data()`. 

This would allow a user to more easily swap out a quasr configuration in an existing script. 

Is it also possible for the user to request the entire row from the database? These quantities could be useful in scripts that further deal with this data.

andrewgiuliani:
Hi Chris, you're right! I have those json files too, we will swap them out so the axis is returned as well. As for the row of the data frame, that should be doable as well. Let me discuss with Jeff

missing-user:
Hi, I also think this is a great addition and would further suggest to optionally cache the requests https://github.com/hiddenSymmetries/simsopt/pull/468

mishapadidar:
@smiet is this PR still in progress or should we close?

smiet:
@andrew and I were planning to have a look-see this week and wrap it up. Currently stuck a bit with how to 'Mock' a request to a server in the testing... 

Also looking into the caching of previous results to minimize the clobbering of the database. Let us keep this PR open and add the code from #468 if we choose so (it creates a folder upon import, which I would rather not, we will discuss the solution).

jsoules:
> @andrew and I were planning to have a look-see this week and wrap it up. Currently stuck a bit with how to 'Mock' a request to a server in the testing...

Let me know if you need a hand with that--it can be a bit idiosyncratic until you get used to it (mocking a context manager, in particular, can be fussy).

src/simsopt/configs/zoo.py
andrewgiuliani: there is now an exception raised higher up if return_style is something else

src/simsopt/configs/zoo.py
andrewgiuliani: this is clarified in the latest commit

src/simsopt/configs/zoo.py
andrewgiuliani: you're right, just added it

src/simsopt/configs/zoo.py
andrewgiuliani: I didn't know this! thanks for letting me know, we've changed the assert to an if-statement

src/simsopt/configs/zoo.py
andrewgiuliani: noted, thanks!

src/simsopt/configs/zoo.py
andrewgiuliani: updated string here, thanks!

tests/configs/test_zoo.py
andrewgiuliani: done

andrewgiuliani (commented):
@mishapadidar @smiet @jsoules I've mocked the requests.get call so that this is a true unit test, please give this a look.  

I have not written an integration test - I don't know if we have the infrastructure set up to run a subset of tests only at merging. Can we do this @mbkumar ?

mbkumar:
> I don't know if we have the infrastructure set up to run a subset of tests only at merging. Can we do this @mbkumar ?

We can definitely do this. It's a bunch of if conditions in the Action file.

tests/configs/test_zoo.py
```diff
@@ -1,16 +1,70 @@
...
+        assert isinstance(curves[0], CurveXYZFourier)
+        assert isinstance(currents[0], ScaledCurrent)
+        np.testing.assert_allclose(curves[0].x, true_coils[0].curve.x)
+        np.testing.assert_allclose(currents[0].get_value(), true_coils[0].current.get_value())
+
+        surfaces, coils = get_QUASR_data(952, return_style='quasr-style')
+        assert isinstance(coils[0], Coil)
+        assert isinstance(surfaces[0], SurfaceXYZTensorFourier)
+        np.testing.assert_allclose(surfaces[0].x, true_surfaces[0].x)
+        np.testing.assert_allclose(coils[0].x, true_coils[0].x)
```
jsoules, line 46: Something to think about is the way that running the test might impact the environment where the test is being run. In general, it's not good if the test makes changes that are visible outside the test, because that could impact subsequent tests. I mention it because in this case I think the test will interact with the caching code--the `get_QUASR_data` function is expected to create a cache file the first time it runs for a particular ID, and then read from the cache file on subsequent runs for the same ID.

For this test, that means the very first time you run it you'll be creating the cache file, but on subsequent runs (and every time, for the code I've highlighted here) you'll be reading from the cache file.

1. I would recommend checking for, and cleaning up, the cache file at the end of the test.
2. If you care about the caching behavior, you might want to check that it's happening between these two calls.

I think you could handle deleting the cache file by using the `FILE_PATH = THIS_DIR / f'QUASR_cache/serial{id_str}.json'` logic you have in the actual implementing function (i.e. `get_QUASR_data()`) and then doing `FILE_PATH.is_file()` to confirm it exists and doing `FILE_PATH.unlink()` if it does (you'd still need to clear the `QUASR_cache` directory as well, I think there's a `rmdir()` method in `pathlib`).

Asserting `not FILE_PATH.is_file()` before the first call to `get_QUASR_data()`, and `FILE_PATH.is_file()` after it, would check that the cache file is getting created. If you'd like to check that the cache is actually getting used, you could also do `mock_get.assert_not_called()` before the first `get_QUASR_data()`, then `mock_get.assert_called_once()` after the SECOND `get_QUASR_data()` (to make sure it actually used the cache). Equivalently, you can also assert against the `mock_get.call_count` variable.

I could also see mocking `builtins.print` and inspecting the values it was called with, if you want to check anything about the function's reporting, but I'm not sure whether I'd bother realistically since "does the thing print this" is unlikely to be mission-critical functionality :smile: The only reason I'd do it is if you wanted to inspect what it was printing to confirm that it says it's using the cache.

tests/configs/test_zoo.py
```diff
@@ -1,16 +1,70 @@
...
+        # requests.get raises an exception
+        mock_response.side_effect = Exception("something went wrong")
         with self.assertRaises(Exception):
             curves, currents, ma = get_QUASR_data(952, return_style='')
+        
+        with self.assertRaises(Exception):
+            # mock os.makedir only here so it throws an error
+            with patch("simsopt.configs.zoo.os.makedirs") as mock_makedirs:
+                mock_makedirs.side_effect = Exception("Failed to create directory")
+                _, _ = get_QUASR_data(953, return_style='quasr-style')
```
jsoules, line 66: And here of course you've hard-coded 953, because if you used 952 again you'd hit the cache and `makedir` wouldn't be called. Subtle! :smile:

If you go with the defined `serial_number` variable I suggested earlier, you could use `serial_number + 1` or something.

Alternatively, if this were a separate test, you would theoretically have deleted the cached file at the end of whatever test created it.

src/simsopt/configs/zoo.py
```diff
@@ -198,38 +200,64 @@ def get_w7x_data(Nt_coils=48, Nt_ma=10, ppp=2):
...
     
     if return_style not in ['simsopt-style', 'quasr-style']:
         raise ValueError(f"invalid return_style: {return_style}, must be either simsopt-style or quasr-style")
     
-    id_str = f"{ID:07d}"
-    # string to 7 digits
+    id_str = f"{ID:07d}" # string to 7 digits
     url = f'https://quasr.flatironinstitute.org/simsopt_serials/{id_str[0:4]}/serial{id_str}.json'
+    
+    FILE_PATH = THIS_DIR / f'QUASR_cache/serial{id_str}.json'
```
jsoules, line 232: I'm unsure if this will still do the right thing on environments where `/` is not the path separator (e.g. windows). If you care, could change to `FILE_PATH = THIS_DIR / 'QUASR_cache' / f"serial{id_str}.json"` (and push it onto `pathlib`).

tests/configs/test_zoo.py
jsoules: You might be able to avoid having to do the cleanup by using a `TemporaryDirectory` context in this test; temporary directories are deleted when you exit the context. See https://adamj.eu/tech/2024/12/30/python-temporary-files-directories-unittest/ for an example (except you'd want to use https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory instead of `NamedTemporaryFile`).

However, using the temporary directory would require changing the definition of `THIS_DIR` in `configs/zoo.py` (which defines `get_QUASR_data()`), which might be complicated, so you're probably best off just deleting the cache file that you expect to have created in the test.

jsoules: It might be convenient to let the user configure where the parent directory for the downloads should be located (for the functions in `zoo.py`)--so that you can use that instead of the `THIS_DIR` which I think (?) is relative to where the `zoo.py` file is located. But that's obviously well outside the scope of this PR.

jsoules (commented):
At @andrewgiuliani's request, I've taken a look at this.

In general I think it's in really good shape; I caught a couple minor things, and made a few broader-scope suggestions that are not germane to this pull request.

The one thing that I think could be improved before merging is how this test will interact with the caching system in the implementation in `zoo.py` (explained in my in-line comments below).

I think you were also wondering about the integration test. I'll leave it to the local experts to determine how to actually flag a test to be run rarely (& what conditions should trigger running it), but apart from that it would just be as simple as running lines 34-40 of `test_QUASR_downloader` without mocking `request`, so that you actually hit the QUASR Navigator web server.

smiet:
It is getting there! I changed the caching to test if the directory (relative to zoo.py, in the install directory) is writeable, if not make a cache in the cwd. (in some types of installation this might be the case, i.e. it is installed in base system python and a user imports it). 

@andrewgiuliani can you address the last few issues by @jsoules? We should not let perfect be the enemy of good enough, and this is a really nice functionality to have, so let's merge it soon!

src/simsopt/configs/zoo.py
```diff
@@ -187,3 +197,82 @@ def get_w7x_data(Nt_coils=48, Nt_ma=10, ppp=2):
...
+        except:
+            raise Exception(f"requests failure on ID {ID:07}")
+
+        if r.status_code == 200:
+            print(f"ID={ID:07} downloaded successfully")
+            surfaces, coils = json.loads(r.content, cls=GSONDecoder)
+            success = True
+            
+            if use_cache: 
+                with open(FILE_PATH, 'wb') as f:
```
missing-user, line 262: This will fail if the directory doesn't exist yet.
```suggestion
                os.makedirs(FILE_PATH.parent, exist_ok=True)
                with open(FILE_PATH, 'wb') as f:
```

PR #428, opened 2024-06-18
ag/cross section fix

andrewgiuliani:
I have used the `Surface.cross_section` algorithm on the entire QUASR database, and notice that very infrequently, the algorithm fails, see below

image

I rewrote the algorithm, making it simpler, to fix the bug.  Rerunning the new algorithm on the entire dataset, I can't find any incorrect cross sections anymore.

image

Note that the jagged cross section where `varphi=0` is because it uniformly samples the cross section in the Boozer theta angle, which may be insufficient as you can see.


src/simsopt/geo/surface.py
```diff
@@ -354,64 +350,47 @@ def cross_section(self, phi, thetas=None):
...
+        cyl_phi = np.where(cyl_phi < 0, cyl_phi + 2*np.pi, cyl_phi)
+        cyl_phi[1, :] += 2*np.pi # second row is the same as the first row, but shifted by 2pi
+
+        varphi_left = varphigrid[0, :]
+        varphi_right = varphigrid[1, :]
+        cyl_phi_left = cyl_phi[0, :]
+        cyl_phi_right = cyl_phi[1, :]
+        
+        phi = phi*np.ones(varphi_left.size)
+        def put_angle_between_left_and_right(angle, left_bound, right_bound):
```
VascoSch92, line 370:
```suggestion
        phi = phi*np.ones(varphi_left.size)
        
        def put_angle_between_left_and_right(angle, left_bound, right_bound):
```

src/simsopt/geo/surface.py
```diff
@@ -354,64 +357,47 @@ def cross_section(self, phi, thetas=None):
...
         # sample the surface at the varphi and theta points
         gamma = np.zeros((varphigrid.shape[0], varphigrid.shape[1], 3))
         self.gamma_lin(gamma, varphigrid.flatten(), thetagrid.flatten())
 
-        # compute the cylindrical phi coordinate of each sampled point on the surface
         cyl_phi = np.arctan2(gamma[:, :, 1], gamma[:, :, 0])
+        cyl_phi = np.where(cyl_phi < 0, cyl_phi + 2*np.pi, cyl_phi)
+        cyl_phi[1, :] += 2*np.pi # second row is the same as the first row, but shifted by 2pi
+
+        varphi_left = varphigrid[0, :]
```
mishapadidar, line 364: Are we trying to determine bisection bounds here?

src/simsopt/geo/surface.py
```diff
@@ -425,7 +411,8 @@ def bisection(phia, a, phic, c):
                 err = np.max(np.abs(a - c))
             b = (a + c) / 2.
             return b
-
+        
+        phi = put_angle_between_left_and_right(phi, cyl_phi_left, cyl_phi_right)
         # bisect cyl_phi to compute the cross section
         sol = bisection(cyl_phi_left, varphi_left, cyl_phi_right, varphi_right)
```
mishapadidar, line 410: just trying to understand here: it looks like, for each `theta`, we are doing a bisection over `varphi`, so that `varphi2phi(varphi) = phi`. Why do we need to know `cyl_phi_right` and `cyl_phi_left`?

mishapadidar (commented):
I think your logic is correct. Most of the notes are for readability and code clean up. Still a bit confused on how the algorithm works though, namely the use of the `put_angle_between_left_and_right` function.

src/simsopt/geo/surface.py
andrewgiuliani: outdated in current state of code

src/simsopt/geo/surface.py
andrewgiuliani: Keep in mind the following:

1)  the range of atan2 is `[-pi, pi)`.  
2)  the initial varphi interval is `[varphi_left,  varphi_right] = [0, 1]`, and in cylindrical coordinates, this could map to e.g. `[cyl_phi_left,  cyl_phi_right] = [-0.1, -0.1+2*pi]`
3) there is no requirement on the value of `phi0` where we are computing the cross section, thus the function `put_angle_above_left_bound` maps `phi0` to `[cyl_phi_left,  cyl_phi_right]`.
4) over the course of the bisection algorithm, we compute `phi( 0.5*(varphi_left + varphi_right) )` using `atan2`, then `put_angle_above_left_bound` maps this new cylindrical value back to the proper cylindrical phi interval since atan2 will always lie in  `[-pi, pi)`.
5) the interval `[varphi_left,  varphi_right]` shrinks, and so too does `[cyl_phi_left,  cyl_phi_right]`, hence why we are updating both at each iteration of bisection.

src/simsopt/geo/surface.py
```diff
@@ -354,64 +357,47 @@ def cross_section(self, phi, thetas=None):
...
-        idx_left = idx_right - 1
-
-        varphi_left = varphigrid[idx_left, np.arange(idx_left.size)]
-        varphi_right = varphigrid[idx_right, np.arange(idx_right.size)]
-        cyl_phi_left = cyl_phi[idx_left, np.arange(idx_left.size)]
-        cyl_phi_right = cyl_phi[idx_right, np.arange(idx_right.size)]
-
-        # this function converts varphi to cylindrical phi, ensuring that the returned angle
-        # lies between left_bound and right_bound.
         def varphi2phi(varphi_in, left_bound, right_bound):
```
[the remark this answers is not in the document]
andrewgiuliani: added docstring, thanks

src/simsopt/geo/surface.py
```diff
@@ -333,19 +340,15 @@ def cross_section(self, phi, thetas=None):
         single curve.
```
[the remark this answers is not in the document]
andrewgiuliani: done

src/simsopt/geo/surface.py
```diff
@@ -354,64 +357,47 @@ def cross_section(self, phi, thetas=None):
...
-            pinc = (phi < left_bound).astype(int)
-            minc = (phi > right_bound).astype(int)
-            phi = phi + 2. * np.pi * (pinc - minc)
-            return phi
+            angle = np.arctan2(gamma[:, 1], gamma[:, 0])
+            angle = np.where(angle < 0, angle+2*np.pi, angle)
+            shifted_angle = put_angle_between_left_and_right(angle, left_bound, right_bound)
+            return shifted_angle
 
         def bisection(phia, a, phic, c):
```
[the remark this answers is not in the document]
andrewgiuliani: yes doable, but tricky due to the discontinuity in `phi(varphi)`.  scipy.bisection could converge to the jump from `-phi` to `pi` due to atan2, rather than the true root you seek.

src/simsopt/geo/surface.py
```diff
@@ -333,19 +340,15 @@ def cross_section(self, phi, thetas=None):
         single curve.
         """
 
-        # phi is assumed to be between [-pi, pi], so if it does not lie on that interval
+        # phi is assumed to be between [0, 2*pi], so if it does not lie on that interval
```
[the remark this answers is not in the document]
andrewgiuliani: done

src/simsopt/geo/surface.py
```diff
@@ -354,64 +357,47 @@ def cross_section(self, phi, thetas=None):
...
+        cyl_phi = np.where(cyl_phi < 0, cyl_phi + 2*np.pi, cyl_phi)
+        cyl_phi[1, :] += 2*np.pi # second row is the same as the first row, but shifted by 2pi
+
+        varphi_left = varphigrid[0, :]
+        varphi_right = varphigrid[1, :]
+        cyl_phi_left = cyl_phi[0, :]
+        cyl_phi_right = cyl_phi[1, :]
+        
+        phi = phi*np.ones(varphi_left.size)
+        def put_angle_between_left_and_right(angle, left_bound, right_bound):
```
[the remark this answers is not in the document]
andrewgiuliani: added docstrings

src/simsopt/geo/surface.py
```diff
@@ -333,19 +340,15 @@ def cross_section(self, phi, thetas=None):
         single curve.
         """
 
-        # phi is assumed to be between [-pi, pi], so if it does not lie on that interval
+        # phi is assumed to be between [0, 2*pi], so if it does not lie on that interval
         # we shift it by multiples of 2pi until it does
```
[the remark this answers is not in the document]
andrewgiuliani: any value of `phi` is acceptable, and I've updated the docstring to reflect this.

src/simsopt/geo/surface.py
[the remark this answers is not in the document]
andrewgiuliani: done

src/simsopt/geo/surface.py
```diff
@@ -333,19 +340,15 @@ def cross_section(self, phi, thetas=None):
...
         phi = phi - np.sign(phi) * np.floor(np.abs(phi) / (2 * np.pi)) * (2. * np.pi)
-        if phi > np.pi:
-            phi = phi - 2. * np.pi
-        if phi < -np.pi:
-            phi = phi + 2. * np.pi
-
```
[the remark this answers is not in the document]
andrewgiuliani: done, thanks for this nice suggestion

andrewgiuliani:
@mishapadidar for future reference, I'm including the figure illustrating what the algorithm is doing (all angles are normalized by 2*pi for consistency with Surface objects in simsopt).

The first graph shows the cylindrical angle as calculated by `atan2(y(varphi, theta),x(varphi, theta))/(2*pi)`.  The problem is that this has a discontinuity when the surface crosses +/- 1.  So we shift by `atan2(y(0, theta),x(0, theta))/(2*pi)` in the second graph.  Any part of the shifted curve that is less than 0 has crossed +/- 1, so we shift negative function values up by 1.  The resulting function is continuous provided that the surface crosses +/- 1 only once.

shift

src/simsopt/geo/surface.py
```diff
@@ -324,38 +324,31 @@ def to_RZFourier(self):
         """
         raise NotImplementedError
 
-    def cross_section(self, phi, thetas=None):
+    def cross_section(self, phi_prime, thetas=None, tol=1e-13):
```
mishapadidar, line 327: name `phi_prime` does not match documentation. As a second point we may want to keep this name as `phi` to prevent breaking code.

src/simsopt/geo/surface.py
andrewgiuliani: our code refactor makes this question irrelevant now, thanks

src/simsopt/geo/surface.py
```diff
@@ -324,38 +324,31 @@ def to_RZFourier(self):
...
-            phi: float
-                toroidal angle
-            thetas: float array
+            phi (float):
+                toroidal angle, the standard cylindrical angle normalized by :math:`2\pi`.
+                There is no restriction on :math:`\phi`, i.e. is can be larger than 1, or
+                smaller than 0.
+
+            thetas (float, array, optional):
                 collocation points to compute cross-section with
```
mishapadidar, line 342: What happens is `thetas` is not passed?

src/simsopt/geo/surface.py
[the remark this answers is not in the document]
andrewgiuliani: after our code refactor, we could have used scipy.bisection, but since we're applying bisection to a set of independent equations (one per value of theta), I implemented the algorithm myself

src/simsopt/geo/surface.py
```diff
@@ -324,38 +324,31 @@ def to_RZFourier(self):
...
-
-        # varphi are the search intervals on which we look for the cross section in
-        # at constant cylindrical phi
-        # The cross section is sampled at a number of points (theta_resolution) poloidally.
-        varphi = np.asarray([0., 0.5, 1.0])
-
+        
         if thetas is None:
             theta = np.asarray(self.quadpoints_theta)
         elif isinstance(thetas, np.ndarray):
```
mishapadidar, line 354: should this be 
`elif isinstance(thetas, np.ndarray)`?

src/simsopt/geo/surface.py
mishapadidar: nit: can we set the default value for `thetas` to be an `int`? This gives some hint as to what it is.

src/simsopt/geo/surface.py
```diff
@@ -324,38 +324,34 @@ def to_RZFourier(self):
...
+                should be computed, between 0 and 1.  If thetas is not provided, then the cross section will be calculated at the positions given by
+                self.quadpoints_theta.
+            tol (float): 
+                the tolerance for the bisection root-finding
 
         Returns
         -------
-            cross_section: float
+            cross_section (np.array): 
                 The cross-section evaluated at :math:`\phi` given support points `thetas`
```
mishapadidar, line 352: perhaps add shape of array `(ntheta,)`?

src/simsopt/geo/surface.py
```diff
@@ -324,38 +324,34 @@ def to_RZFourier(self):
...
-            phi = phi + 2. * np.pi
-
-        # varphi are the search intervals on which we look for the cross section in
-        # at constant cylindrical phi
-        # The cross section is sampled at a number of points (theta_resolution) poloidally.
-        varphi = np.asarray([0., 0.5, 1.0])
-
+        
         if thetas is None:
             theta = np.asarray(self.quadpoints_theta)
```
mishapadidar, line 356: doesnt seem to affect functionality, but why do we change from `thetas` to `theta`?

src/simsopt/geo/surface.py
```diff
@@ -364,82 +360,75 @@ def cross_section(self, phi, thetas=None):
...
+            return angle
+        
+        # set the initial brackets for bisection
+        varphia = np.zeros(theta.size)
+        phia = np.zeros(theta.size)
+        varphic = np.ones(theta.size)
+        phic = np.ones(theta.size)
+
+        err = np.inf
+        while err > tol:
```
mishapadidar, line 411: should we try scipy root here?

tests/geo/test_surface_xyzfourier.py
```diff
@@ -74,12 +74,12 @@ def test_toRZFourier_lossless_at_quadrature_points(self):
     def subtest_toRZFourier_lossless_at_quadraturepoints(self, surface_type):
```
mishapadidar, line 74: can you describe what the subtest is?

mishapadidar (commented):
looks mostly good to me:
- minor comments on code.
- can you take a look at the compiled documentation and make sure it is alright?
- if scipy root finding works it would clean up code.

src/simsopt/geo/surface.py
andrewgiuliani: using `scipy.optimize.root_scalar` or `scipy.optimize.bisect` is doable, but you have to solve each independent scalar equation with separate solve calls since these functions do not support independent vector equations.  I tried that, but it's much much slower than the bisection algorithm currently in the code.

I also tried `scipy.optimize.fsolve`, and it is much slower than the vectorized bisection algorithm currently implemented (geo.test_surface.isSelfIntersecting takes 15s with fsolve and 7s using the vectorized bisection).

mishapadidar: okay the, lets keep the bisection

mishapadidar:
its looking good. just address the couple comments about documentation and we can get this thing landed.

src/simsopt/geo/surface.py
andrewgiuliani: kept `phi_prime` as is, so it's compatible with the figure in this thread.  changed default value for `thetas` to `int`.

src/simsopt/geo/surface.py
andrewgiuliani: updated docstring

src/simsopt/geo/surface.py
```diff
@@ -324,38 +324,34 @@ def to_RZFourier(self):
...
         ----------
-            phi: float
-                toroidal angle
-            thetas: float array
-                collocation points to compute cross-section with
+            phi (float):
+                toroidal angle, the standard cylindrical angle normalized by :math:`2\pi`.
+                There is no restriction on :math:`\phi`, i.e. is can be larger than 1, or
+                smaller than 0.
+            thetas (float, array, optional):
```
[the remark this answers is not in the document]
andrewgiuliani: updated docstring

src/simsopt/geo/surface.py
andrewgiuliani: changed `thetas` to `theta`

tests/geo/test_surface_xyzfourier.py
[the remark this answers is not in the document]
andrewgiuliani: done

tests/test_files/serial2680021.json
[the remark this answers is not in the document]
andrewgiuliani: this is a serialized device on which the previous cross section algorithm failed, and the new implementation in this PR passes

src/simsopt/geo/surface.py
```diff
@@ -324,145 +324,141 @@ def to_RZFourier(self):
         """
         raise NotImplementedError
 
-    def cross_section(self, phi, thetas=None):
+    def cross_section(self, phi_prime, theta=32, tol=1e-13):
```
mishapadidar, line 327: is it possible to revert the notation `phi_prime` back to `phi` and `theta` back to `thetas`? This ensures that we arent breaking any code.

src/simsopt/geo/surface.py
```diff
@@ -324,145 +324,141 @@ def to_RZFourier(self):
...
+        
+        varphi_root = (varphia + varphic) / 2.
+        cross_section = np.zeros((varphi_root.size, 3))
+        self.gamma_lin(cross_section, varphi_root, theta)
         return cross_section
 
     @SimsoptRequires(get_context is not None, "is_self_intersecting requires ground package")
     @SimsoptRequires(contour_self_intersects is not None, "is_self_intersecting requires the bentley_ottmann package")
-    def is_self_intersecting(self, angle=0., thetas=None):
+    def is_self_intersecting(self, angle=0., theta=None):
```
mishapadidar, line 440: same here. This is a breaking change. Consider using thetas everywhere.

src/simsopt/geo/surface.py
andrewgiuliani: added

src/simsopt/geo/surface.py
andrewgiuliani: reverted to avoid compatibility issues

mishapadidar (approved):
just make sure that the argument naming `phi_prime` is reverted to `phi` to not break code.

mishapadidar (approved):

PR #430, opened 2024-06-24
Numpy type hinting

mbkumar:
This PR replaces nptyping with numpy.typing for type hinting numpy objects.
Numpy is also constrained to be <2.0.0 to allow other projects such as f90wrap to catch up.


smiet (approved):
Looks good, thanks for the fast work!

PR #432, opened 2024-06-27
Fix for python 3.8 and numpy dependency

mbkumar:
This PR adds python-version specific dependencies to address numpy compatibility with v1.x and v2.x.

There is one spec test failing in 3.8. Could that be due to numpy 2.0 and numpy 1.x differences?
@smiet, can you please invetigate?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,8 +1,9 @@
 [build-system]
 requires = [
-  "scikit-build-core>=0.3.3",
+  "scikit-build-core",
   "pybind11",
-  "oldest-supported-numpy",
+  "numpy >= 2.0.0; python_version > '3.8'",
+  "oldest-supported-numpy; python_version <= '3.8'",
   "setuptools_scm>=8.0",
   'tomli; python_version < "3.11"',]
 build-backend = "scikit_build_core.build"


smiet:
all passing now, fixed on current spec master by not requiring 2.0 if python<3.8. 

There is a new PR waiting that removes numpy dependency on install, but that shouldn't affect this.

mbkumar:
@smiet 

Spec related tests are still failing. https://github.com/hiddenSymmetries/simsopt/actions/runs/9713949797/job/26811901758 Can you please check?

smiet:
@mbkumar  the failures you are referring to are caused by an identifier collision in the python cache of the Fortran memory. They occur because f90rwap uses an identifier based on the pointer in memory for its python-cache, and re-loading a Spec object results in sucessive deallocation and allocation, with possibility of occupying the same location on memory-constrained systems (such as runners). This results in the collision such that python references an outdated and misshapen array. 

Will be addressed and fixed in PR  #431, if you want to speed it up, give it an approve, or directly merge it ;)

smiet (approved):
I trust you in that the scikit-build-core version is necessary, and not too-constraining. The numpy versioning is what it should be until 3.8 reaches EOL.

Issue #435, opened 2024-07-11
Wrong syntax in logging statements


eguiraud-pf:
Hello,

the logging code here:

https://github.com/hiddenSymmetries/simsopt/blame/f6a8412047ad418ba287ee9c32c707802f064970/src/simsopt/mhd/boozer.py#L181-L182

seems wrong, and triggers the following exception when running with log level INFO:

```
    self.boozer.run()
  File "/home/devuser/.cache/pants/named_caches/pex_root/venvs/s/f700390b/venv/lib/python3.10/site-packages/simsopt/mhd/boozer.py", line 179, in run
    logger.info('mnmax:', wout.mnmax, ' len(xm):', len(wout.xm), ' len(xn):', len(wout.xn))
Message: 'mnmax:'
Arguments: (72, ' len(xm):', 72, ' len(xn):', 72)
```

You can see the syntax is wrong by running the following at the prompt (with `error` instead of `info` so the logging actually happens):

```
import logging
logging.error('mnmax:', 242, ' len(xm):', 24, ' len(xn):', 24)
```

landreman:
@eguiraud-pf do you agree that #436 would resolve this?

eguiraud-pf:
thanks for the quick reaction @landreman , it should, yes!

eguiraud:
Hello, sorry to bother, would it be possible to release a version with this fix? :pray:

landreman:
Ok, just created a release v1.6.4.

eguiraud-pf:
Thank you very much!

PR #436, opened 2024-07-11
Fix logging error in boozer.py to address #435

landreman:
Fix for github issue #435


rogeriojorge (approved):
Looks good and it seems to fix the underlying bug

PR #437, opened 2024-07-17
[VMEC,SurfaceRZFourier] fix inconsistency regarding mpol

jons-pf:
VMEC only ever uses poloidal modes up to `m=mpol-1` with `mpol` being specified in the VMEC input. `SurfaceRZFourier` calls their poloidal Fourier resolution also `mpol`, but includes poloidal modes including up to `m=mpol`. This leads to inconsistencies when converting between VMEC input/output files and `SurfaceRZFourier`. This PR addresses this inconsistency.

### DIFF
--- a/src/simsopt/geo/surfacerzfourier.py
+++ b/src/simsopt/geo/surfacerzfourier.py
@@ -35,6 +35,10 @@ class SurfaceRZFourier(sopp.SurfaceRZFourier, Surface):
 
     and the same for :math:`z(\theta, \phi)`.
 
+    Note that the `mpol` in this is different than the `mpol` in VMEC:
+    VMEC uses poloidal mode numbers `m=0, 1, ..., (mpol-1)`,
+    but here we use `m=0, 1, ..., mpol`.
+
     Here, :math:`(r,\phi, z)` are standard cylindrical coordinates, and theta
     is any poloidal angle.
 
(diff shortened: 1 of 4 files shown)


eguiraud-pf:
Hi, to add to @jons-pf 's comment, some examples of inconsistencies. When the indata file contains RBC/ZBS Fourier coefficients for `m == mpol` (which VMEC ignores but SurfaceRZFourier includes):

- `simsopt.mhd.Vmec._boundary` does not reflect exactly what VMEC uses as boundary condition 
- similarly `SurfaceRZFourier.from_vmec_input("input.best")` does not correspond exactly to the boundary VMEC optimizes for

src/simsopt/geo/surfacerzfourier.py
```diff
@@ -190,7 +194,10 @@ def from_wout(cls, filename: str, s: float = 1.0,
             interp = interp1d(s_full_grid, zmnc, kind=interp_kind, axis=0)
             zbc = interp(s)
 
-        mpol = int(np.max(xm))
+        # VMEC only considers poloidal modes up to `m=(mpol-1)`,
+        # but `SurfaceRZFourier` includes `m=mpol`.
+        # Here, `xm` comes from VMEC and `mpol` goes into `SurfaceRZFourier`.
+        mpol = int(np.max(xm)) - 1
```
landreman, line 200: I don't think this -1 should be here. mpol here is SurfaceRZFourier's mpol, so it should equal the maximum m.

src/simsopt/mhd/vmec.py
```diff
@@ -523,7 +528,9 @@ def set_indata(self):
         mpol_capped = np.min([boundary_RZFourier.mpol, 101])
         ntor_capped = np.min([boundary_RZFourier.ntor, 101])
         # Transfer boundary shape data from the surface object to VMEC:
-        for m in range(mpol_capped + 1):
+        # The highest mode number that VMEC can use is `m=100`
+        # if the max resolution is `mpol=101`.
+        for m in range(mpol_capped):
```
landreman, line 533: It doesn't seem correct to remove the +1 here. mpol_capped is essentially mpol of a SurfaceRZFourier object (via `boundary_RZFourier = self.boundary.to_RZFourier()` and `mpol_capped = np.min([boundary_RZFourier.mpol, 101])`. So the m=mpol mode should be included rather than excluded. If you want to change the 101 to 100 on line 523/528, that would be fine.

jons-pf:
Just to give a signal of life here - I have not forgotten about this issue, but sadly have not found time yet to continue digging into this.

PR #438, opened 2024-07-22
Single-stage optimization with spec: CoilSet, ReducedCoilSet and CoilNormalField (PR 2 of 2)

smiet:
This PR adds new features, and enables the optimizations  presented at the Simons' retreat and meeting, and available in preprint (https://arxiv.org/abs/2407.02097). 

The main additions are: 

### class `CoilSet`
A helper class to handle a set of `coil` objects as a single object:
 -  convenience class
 -  provides helper methods for all the coil-related optimization targets
 - serves as the base class for  `ReducedCoilSet`
 - allows the `ReducedCoilSet` and `CoilNormalField` to be implemented with minimal code duplication.

### class `ReducedCoilSet`
Identical in use to `CoilSet`, but it's degrees-of-freedom are linear combinations of the coil degrees-of-freedom. 
- Calculates its degrees-of-freedom by performing a singular-value-decomposition on an arbitrary user-provided mapping. 
- Ability to re-calculate the reduced basis when changing the function or  other relevant properties of the CoilSet.
- Removes some overdetermination inherent in direct-from-coil optimization. 
- Reduces the dimensionality of the optimization problem, speeds up optimization.  

### class `CoilNormalField`
Identical in use to a NormalField and interfaces with SPEC, but has degrees-of-freedom specified by a `CoilSet` or `ReducedCoilSet`. 
- Provides the normal field on a boundary in the SPEC conventions.
- Uses a cache to avoid redundant Fourier transformations.
- Handles both `CoilSet` and `ReducedCoilSet`, provides method to reduce the `CoilSet` using the mapping to SPEC input parameters (normal field Fourier components).   

With these changes the functionality I describe in my preprint will be added to the simsopt master branch. I look forward to your suggestions to improve my code.


src/simsopt/geo/surfacerzfourier.py
```diff
@@ -421,6 +421,75 @@ def from_pyQSC(cls, stel: Qsc, r: float = 0.1, ntheta=20, mpol=10, ntor=20, **kw
 
         surf.local_full_x = surf.get_dofs()
         return surf
+    
+    def copy(self, **kwargs):
```
rogeriojorge, line 425: What is the difference between this function and simply setting the dofs of a new surface object as, for example

`
surf = SurfaceRZFourier.from_vmec_input(filename, range="half period", nphi=nphi, ntheta=ntheta)
surf_big = SurfaceRZFourier(dofs=surf.dofs,nfp=nfp, mpol=surf.mpol,ntor=surf.ntor,
                            quadpoints_phi=new_quadpoints_phi,quadpoints_theta=new_quadpoints_theta)
`

tests/field/test_mgrid.py
```diff
@@ -1,9 +1,8 @@
 #!/usr/bin/env python
 
 import unittest
-import os
```
rogeriojorge, line 4: Did os fail in same particular case? Is it why it was removed?

tests/geo/test_surface_rzfourier.py
```diff
@@ -771,6 +771,24 @@ def test_fourier_transform_scalar(self):
         # Check that the result is the same as the original field:
         np.testing.assert_allclose(field/2*np.pi**2, field2)
 
+    def test_copy_method(self):
```
rogeriojorge, line 774: Some lines of copy are not being tested (see codecov)

src/simsopt/geo/surfacerzfourier.py
```diff
@@ -705,7 +774,7 @@ def fourier_transform_scalar(self, scalar, mpol=None, ntor=None, normalization=N
         if not stellsym:
             A_mnc[0, ntor] = np.sum(scalar) / (ntheta_grid * nphi_grid)
         if normalization is not None:
-            if isinstance(normalization, float):
+            if not isinstance(normalization, float):
```
rogeriojorge, line 777: Good catch!

rogeriojorge (commented):
This looks really good. I am just not sure if the "copy" method is actually needed.

tests/geo/test_surface_rzfourier.py
smiet: Should be in the latest push, with a logic fix!

src/simsopt/geo/surfacerzfourier.py
smiet: I'd be much happier if the mistake weren't my own ;)

tests/field/test_mgrid.py
smiet: most tests use `monty.ScratchDir` but the `test_mgrid` used `tempfile` and `os.chdir`, without changing back. 

This led my subsequent tests to fail mysteriously as the directory that the test-kernel had moved into had been deleted. 

Reverted this test to the more readable and less error-prone `moty.ScratchDir`, and therefore `os` import was also no longer needed to effect the `chdir`, so I removed it.

smiet:
Sorry to push this, but could we move forward with this PR? All tests are now pasing (after the numpy2.0 speedbump), and the tests now cover all new code. 

The only exception is a plotting utility function that I am not testing because not all runners have visualization capabilities, but I do think it's inclusion is valuable as a demonstration. 

@landreman, @rogeriojorge @abaillod

rogeriojorge:
Looks good to me. However, I'll let a day or two go by before accepting in case @landreman or @abaillod wants to add something here.

rogeriojorge:
There's a conflict with the recent merge. Could you fix it @smiet? I'll merge in afterwords.

rogeriojorge (approved):

rogeriojorge:
Merged

PR #440, opened 2024-07-31
Enable open upper limits for the grids in interpolation code

mbkumar:
This PR enables open endpoints for the upper limits of the meshes when doing interpolation. The main reason to enable this functionality is to allow for [0, phi_max) generated by BMW mesh.

### DIFF
--- a/src/simsopt/field/magneticfieldclasses.py
+++ b/src/simsopt/field/magneticfieldclasses.py
@@ -862,10 +862,14 @@ class InterpolatedField(sopp.InterpolatedField, MagneticField):
         Args:
             field: the underlying :mod:`simsopt.field.magneticfield.MagneticField` to be interpolated.
             degree: the degree of the piecewise polynomial interpolant.
-            rrange: a 3-tuple of the form ``(rmin, rmax, nr)``. This mean that the interval :math:`[rmin, rmax]` is
-                    split into ``nr`` many subintervals.
-            phirange: a 3-tuple of the form ``(phimin, phimax, nphi)``.
-            zrange: a 3-tuple of the form ``(zmin, zmax, nz)``.
+            rrange: a 3-tuple of the form ``(rmin, rmax, nr)`` or a 4-tuple of the form ``(rmin, rmax, nr, include_endpoint)``.
+            The default for `include_endpoint` is True. This mean that the interval :math:`[rmin, rmax]` is
+            split into ``nr`` many subintervals. When `include_endpoint` is False, interval :math:`[rmin, rmax)` is
+            split into ``nr`` many subintervals.
+            phirange: a 3-tuple of the form ``(phimin, phimax, nphi)`` or a 4-tuple of the form
+            ``(phimin, phimax, nphi, include_endpoint)``.
+            zrange: a 3-tuple of the form ``(zmin, zmax, nz)`` or a 4-tuple of the form
+            ``(zmin, zmax, nz, include_endpoint)``.
             extrapolate: whether to extrapolate the field when evaluate outside
                          the integration domain or to throw an error.
             nfp: Whether to exploit rotational symmetry. In this case any angle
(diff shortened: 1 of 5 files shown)


src/simsopt/field/magneticfieldclasses.py
```diff
@@ -862,10 +862,14 @@ def __init__(self, field, degree, rrange, phirange, zrange, extrapolate=True, nf
         Args:
             field: the underlying :mod:`simsopt.field.magneticfield.MagneticField` to be interpolated.
             degree: the degree of the piecewise polynomial interpolant.
-            rrange: a 3-tuple of the form ``(rmin, rmax, nr)``. This mean that the interval :math:`[rmin, rmax]` is
-                    split into ``nr`` many subintervals.
-            phirange: a 3-tuple of the form ``(phimin, phimax, nphi)``.
-            zrange: a 3-tuple of the form ``(zmin, zmax, nz)``.
+            rrange: a 3-tuple of the form ``(rmin, rmax, nr)`` or a 4-tuple of the form ``(rmin, rmax, nr, include_endpoint)``.
+            The default for `include_endpoint` is True. This mean that the interval :math:`[rmin, rmax]` is
```
andrewgiuliani, line 866: typo, this `means`

andrewgiuliani (commented):
Hi Bharat, I do not understand why this PR is needed.  

You want to interpolate the BMW field on the cylindrical domain [rmin, rmax] x [phimin, phimax] x [zmin, zmax].  If you do not include the endpoint, what does the field interpolator do on the interval [rmax-h, rmax)?

The field is periodic in phi, so I'm not sure why you can't just always include the endpoint there.

src/simsopt/field/magneticfieldclasses.py
```diff
@@ -862,10 +862,14 @@ def __init__(self, field, degree, rrange, phirange, zrange, extrapolate=True, nf
...
+            split into ``nr`` many subintervals. When `include_endpoint` is False, interval :math:`[rmin, rmax)` is
+            split into ``nr`` many subintervals.
```
andrewgiuliani, line 868: what does it mean [rmin, rmax]` is split into ``nr`` many subintervals as opposed to `[rmin, rmax)` is split into ``nr`` many subintervals ?

mbkumar: To concretize this, lets suppose you want 5 points between 0 and 1. If you include the end point, the mesh would [0, 0.25, 0.5, 0.75, 1]. If you don't want to include the end point, the mesh generated would be [0, 0.2, 0.4, 0.6, 0.8]. So the mesh would be quite different. Unfortunately, BMW generates phi grid with end point not included.

andrewgiuliani: But then what happens on the small subinterval (2pi/nfp - h, 2pi/nfp)?  I guess the behavior would be dictated by the `extrapolate` keyword argument.

Since the BMW field is nfp-periodic, you can just rotate the B-field values from B(R, phi=0, Z) to B(R, phi=2pi/nfp, Z) and obtain the endpoint value.

mbkumar:
Good catch. I didn't think about the implications. Should I add  periodic
keyword?

landreman:
My reaction is the same as @andrewgiuliani 's - to avoid extrapolating, the thing to do would be to rotate the BMW field from phi=0 by 2pi/nfp and add the result as a new grid point at phi=2pi/nfp. This could be done in whatever function you have that passes data from BMW to the InterpolatedField.

It seems unlikely that a user would have `include_endpoint=False` in the R or Z coordinates, so I'd be inclined not to support those cases for simplicity.

smiet:
I agree with @landreman and @andrewgiuliani, BMW will interpolate between the last phi plane and the first (flipped and copied according to the relevant symmetry), and the equivalently functioning `InterpolatedField` will have the last phi plane appended. 

Technically `InterplolatedField` contains redundant information, but the extra cost is almost negligible and evaluation potentially faster.

PR #446, opened 2024-09-14
Fix MPI issue with vmec crashes in 1st evaluation

landreman:
When vmec fails to converge on the first function evaluation of an optimization, MPI proc 0 exits with an uncaught exception while other procs do not, causing the code to hang due to the MPI communication getting out of sync. This can result is wasted time on the cluster, as jobs don't exit cleanly and keep occupying nodes even though they are not actually working on an optimization. This PR fixes this issue, and the new test here provides coverage for this issue.


rogeriojorge (approved):
Looks ok to me

PR #450, opened 2024-10-12
Coil forces

smhurwitz:
This is to pull all the force optimization work into master.


landreman:
In `selffield.py`, the functions `G`, `K`, `local_field`, `local_field_rect`, `field_from_other_coils`, and `field_from_other_coils_pure` are not used in any test or example. [Codecov is flagging](https://app.codecov.io/gh/hiddenSymmetries/simsopt/commit/e1c699818e25d57e4918fe530730c543168d99c1) these functions for having no test coverage. The function `local_field_rect` would need some work anyway to use a general frame instead of the Frenet frame, and both `local_field_rect`, `field_from_other_coils` should be vectorized to avoid the explicit `for` loop. So, I propose we remove these functions for now from the PR, since getting these functions polished and tested would delay the merge of the functions for Lorentz force, and they can be merged in a separate PR later. If anyone objects, let me know.

landreman:
Hearing no objections, I'll merge.

landreman (approved):

PR #451, opened 2024-10-17
Specify coil group names in load_coils_from_makegrid_file

daringli:
Adds an extra argument 'group_names' to the load_coils_from_makegrid_file() function, allowing the user to provide a list of strings that specify which coil groups to fetch from the makegrid file. Only coil groups with names that match entries in the list will be returned.

This is useful to for example only visualize the 'banana coils' for the Henneberg-Plunk compact QA. The circular toroidal field coils add a lot of clutter to the plot if they are not excluded.

Potential usability issues: if the user provides a single group name not in a list ('RotatedCurve1' instead of the correct ['RotatedCurve1']), the string itself might be interpreted as a list of characters. What is the standard simsopt way to resolve this?


landreman:
I think some of the Extensive CI tests are failing because this fork is a bit behind master (the workflow should use python 3.11 instead of 3.8), so the latest master should be merged into the fork. Don't worry about the failing Docker workflow - that is due to a permission issue with that workflow for forks.

Also please add a unit test or two to cover the case in which `group_names` is not None.

daringli:
> I think some of the Extensive CI tests are failing because this fork is a bit behind master (the workflow should use python 3.11 instead of 3.8), so the latest master should be merged into the fork. Don't worry about the failing Docker workflow - that is due to a permission issue with that workflow for forks.
> 
> Also please add a unit test or two to cover the case in which `group_names` is not None.

I've added the relevant tests, and I now also cover the case where group_names is just a single string rather than a list of strings (the sole string is put inside a list of length one). This case is also covered by a test.

The CI on my fork has always been failing: I'm not sure what to do about it but doing nothing seems fine for now.

landreman:
> The CI on my fork has always been failing: I'm not sure what to do about it 

You can look at the github actions logs to see why the tests are failing. Here the extensive CI is failing because the file `coils.M16N08` was not yet added to the repository.

daringli:
Looks like the file wasn't found due to my usage of relative paths to reach the test_files directory. Updated to use the TEST_DIR import.

landreman (approved):
Not sure why codecov is failing, but it looks good to me.

Issue #455, opened 2024-11-12
NERSC perlmutter shifter container: importing mpi4py in 'hiddensymmetries/simsopt:latest' results in fatal errors.


daringli:
Possibly related to issue #395.

On 'latest', when importing mpi4py, I get the following fatal error
```
Traceback (most recent call last):
  File "/global/u2/s/sbuller/test_MPI_singularity/test.py", line 1, in <module>
    from mpi4py import MPI
ImportError: /venv/lib/python3.10/site-packages/mpi4py/MPI.cpython-310-x86_64-linux-gnu.so: undefined symbol: MPI_Neighbor_alltoallv_c
```
This occurs in optimization scripts, but also in a minimal working example consisting solely of
`from mpi4py import MPI` when called from a job script:
```
#!/bin/bash -l
#SBATCH --time=00:01:00
#SBATCH --qos=debug
#SBATCH --nodes=1
#SBATCH --ntasks-per-node=128
#SBATCH --cpus-per-task=1
#SBATCH --constraint=cpu
#SBATCH --image=hiddensymmetries/simsopt:latest
#SBATCH --account=m4529
#SBATCH -o out.%j
#SBATCH -e err.%j

module list
echo $LD_LIBRARY_PATH
time srun shifter /venv/bin/python test.py
```
where 'test.py' just contains `from mpi4py import MPI`. I've attached the needed files in a zip (although you'd need to edit the 'account' field in the job script to run it).

The error does not occur with `image=hiddensymmetries/simsopt:v1.6.4`.

Zip with minimal example: [test_MPI_shifter.zip](https://github.com/user-attachments/files/17721569/test_MPI_shifter.zip)

landreman:
I just had exactly this same issue with a separate project, so can comment. mpi4py recently released version 4.0, and the problem started then. Forcing mpi4py<4 when pip-installing mpi4py in the Dockerfile resolves the problem. I notified NERSC of the issue.

landreman:
Resolved by #456

PR #456, opened 2024-11-12
Force mpi4py < 4 in Dockerfile due to shifter problem

landreman:
This should address #455

### DIFF
--- a/ci/Dockerfile.ubuntu
+++ b/ci/Dockerfile.ubuntu
@@ -40,7 +40,10 @@ RUN python3 -m pip install wheel
 RUN python3 -m venv /venv/
 
 RUN /venv/bin/pip install -U pip
-RUN /venv/bin/python -m pip install numpy scipy jax jaxlib f90nml mpi4py jupyter notebook ipython qsc sympy scikit-build ninja "pybind11[global]" cmake f90wrap h5py
+# 2024-11-12 MJL: mpi4py < 4.0 in the next line is due to
+# https://github.com/hiddenSymmetries/simsopt/issues/455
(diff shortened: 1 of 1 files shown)


mbkumar:
Nersc is moving away from shifter to Podman. Should we attempt to do the
same?

daringli:
How would I create (or where do I find?) a shifter container from this pull request so I can try it out on perlmutter?

mbkumar:
I don't think the built Docker image is uploaded to Docker. You can copy
the Dockerfile and try building the image somewhere and then copy the image
to Perlmutter.

Bharat Medasani

landreman:
> How would I create (or where do I find?) a shifter container from this pull request so I can try it out on perlmutter?

I created a branch `ml/shifter_mpi4py_fix2` which triggers the docker build in github actions and should push the updated container `hiddensymmetries/simsopt:test` to docker hub. It's [running now](https://github.com/hiddenSymmetries/simsopt/actions/runs/11808542318/job/32897281638) and should push the container in about half an hour, if it works.

mbkumar:
Matt,

The one for branches runs all the times. It's not a big deal, but if you
can add an if clause that would be great.

Bharat Medasani

daringli (approved):
I can confirm that running with
#SBATCH --image=hiddensymmetries/simsopt:test
on perlmutter works.

landreman:
> Matt, The one for branches runs all the times. It's not a big deal, but if you can add an if clause that would be great. 

The `ml/shifter_mpi4py_fix2` isn't intended to merge in to master - it was just a hack to build and push the container once to test it, and we'll never use that branch again. In `ml/shifter_mpi4py_fix` without the 2 at the end, the container is never pushed to docker hub for branches.

PR #457, opened 2024-11-13
Fix logical errors and typos in Spec equilibrium object

missing-user:
There were a few bugs in the Spec equilibrium object that I spotted:
- Many error checks were written, but the `raise` clause was omitted, so they didn't take effect
- The RHS of the assignments to `si.iota[0:self.nvol+1]` and `si.oita[0:self.nvol+1]` had the wrong shape
- array shape check in `as_spec` compared tuple to list and always raised

### DIFF
--- a/src/simsopt/mhd/spec.py
+++ b/src/simsopt/mhd/spec.py
@@ -442,11 +442,11 @@ class Spec(Optimizable):
 
         # Check inputs
         if not isinstance(pressure_profile, ProfileSpec):
-            ValueError('Input should be a ProfileSpec')
+            raise ValueError('Input should be a ProfileSpec')
 
         # Check size
         if pressure_profile.dofs.full_x.size != self.mvol:
-            ValueError('Invalid number of dofs. Shoudl be equal to Mvol!')
+            raise ValueError('Invalid number of dofs. Shoudl be equal to Mvol!')
 
         # Update pressure profile
         if pressure_profile is not self._pressure_profile:
@@ -478,17 +478,17 @@ class Spec(Optimizable):
         """
 
         if not isinstance(volume_current_profile, ProfileSpec):
-            ValueError('Input should be a ProfileSpec')
+            raise ValueError('Input should be a ProfileSpec')
 
         # Check size
         if volume_current_profile.dofs.full_x.size != self.mvol:
-            ValueError('Invalid number of dofs. Shoudl be equal to Mvol!')
+            raise ValueError('Invalid number of dofs. Shoudl be equal to Mvol!')
 
         # Volume current is a cumulative property
         volume_current_profile.cumulative = True
 
         if volume_current_profile is not self._volume_current_profile:
-            logging.debug('Replacing pressure_profile in setter')
+            logging.debug('Replacing volume_current_profile in setter')
             if self._volume_current_profile is not None:
                 self.remove_parent(self._volume_current_profile)
             self._volume_current_profile = volume_current_profile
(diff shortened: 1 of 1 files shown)


src/simsopt/mhd/spec.py
```diff
@@ -1029,10 +1028,10 @@ def run(self, update_guess: bool = True):
             si.curtor = si.ivolume[self.nvol - 1] + np.sum(si.isurf)
 
         if self.iota_profile is not None:
-            si.iota[0:self.nvol+1] = self.iota_profile.get(np.arange(0, self.nvol))
+            si.iota[0:self.nvol+1] = self.iota_profile.get(np.arange(0, self.nvol+1))
 
         if self.oita_profile is not None:
-            si.oita[0:self.nvol+1] = self.oita_profile.get(np.arange(0, self.nvol))
+            si.oita[0:self.nvol+1] = self.oita_profile.get(np.arange(0, self.nvol+1))
```
missing-user, line 1034: nvol+1 is the correct size here, compare to:

https://github.com/missing-user/SPEC/blob/6a231606a626968fc1b83eed7dfa01c234842905/InputFiles/Verification/forcefree/solovev/generate_spec_nl_from_vmec.m#L167

all other iota sizes in spec wrapper must be adjusted

missing-user:
I noticed, that parts of the codebase assume all profiles to be the same size (mvol), in this example according to the error message `local_full_x = self.mvol`  
https://github.com/hiddenSymmetries/simsopt/blob/9fc2c36c7888aa3f677af3ad4f71d63a3e673a21/src/simsopt/mhd/profiles.py#L93-L98
or here, where `volume_current_profile` is simply overwritten in the vacuum region for freeboundary runs, so the profile entry has to be at least `mvol` in size: 
https://github.com/hiddenSymmetries/simsopt/blob/9fc2c36c7888aa3f677af3ad4f71d63a3e673a21/src/simsopt/mhd/spec.py#L1001-L1019

while other parts create different sizes for different profiles.
https://github.com/hiddenSymmetries/simsopt/blob/9fc2c36c7888aa3f677af3ad4f71d63a3e673a21/src/simsopt/mhd/spec.py#L766-L781

What is the intended design @smiet ? I think variable size will result in stricter error handling, whereas having the same size might confuse users, on which elements actually affect the result. Of course, having all profiles the same size makes them easier to switch out, iterate over, etc.

smiet:
@missing-user Thank you for the great work! Sorry to keep you waiting.  The inconsistencies are not intentional and need to be addressed.

The design relies on the user to enforce that the 'ProfileSpec' object is the right size and type for the profile in question. There are some helper functions that aid the user in generating profiles. From the `simsopt` side it is important that the profiles only contain the degrees-of-freedom that actually affect the plasma, so having them all `mvol` is not an option. 

> I noticed, that parts of the codebase assume all profiles to be the same size (mvol), in this example according to the error message `local_full_x = self.mvol`

The error is in the message here, the test compares to the number of degrees-of-freedom of the SpecProfile instance against the 'volume' (interface label) that is trying to be set, which depends on the profile, and the message should be changed to reflect that.  


> or here, where `volume_current_profile` is simply overwritten in the vacuum region for freeboundary runs, so although `nvol` elements are used, the profile has to be at least `mvol` in size:

This is due to the (confusing) use of 'cumulative profiles'. The profile on the `simsopt` side has nvol degrees-of-freedom, but on the SPEC side needs to populate mvol elements of the array. 
The value of the volume current in the vacuum needs to be changed if the current inside of it is changed, but this cannot be to an arbitrary value and should not be a DoF. 

We have to keep the DoFs on the `simsopt` side correct, as the optimizer dislikes having degrees-of-freedom that do not change anything (or worse, they do change the elements of the array, leads to calculations that do not reflect physical reality, i.e. current in the vacuum region). 

I have not tested the profiles extensively, and there are indeed likely inconsistensies that you point  out, like:

> Also @smiet , shouldn't iota profiles be `nvol+1` elements and not `mvol`? How can the vacuum computational boundary have a prescribed `iota`? [#457 (comment)](https://github.com/hiddenSymmetries/simsopt/pull/457#discussion_r1845259154)

Yes, this must be wrong.

I now see that the the `Spec.activate_profile()` function might not work correctly for all profiles, volume-current should skip the first element for example. 

I see you refactored the NormalField, but there is a `simsopt` specific issue with this: when changing a single element in the array, the recompute bell is not triggered. Setting the entire `Optimizable.local_full_x` does trigger it, or you can do it manually. 

I did very much appreciate the full-array setters and getters `set_vns_vns_asarray` as it made code more readable, and it is used in CoilNormalField which gets the full array from a Fourier Transform, and needs it to be set. Could you rename the deleted get_dofs to _array_to_dofs that acts on the input array and sets the `local_full_x`?

missing-user:
Thanks for the thorough feedback @smiet , I'll pick this back up after I submit my thesis next month. Just wanted to confirm know that I haven't forgotten about this.

PR #458, opened 2024-11-18
np.NINF to -np.inf for np2.0 compatibility

missing-user:
NINF was deprecated and should be replaced with -np.inf


landreman (approved):
Thanks for this fix!

PR #459, opened 2024-11-21
(Minor Fix) two broken examples in docs

missing-user:
Two examples in the documentation folder weren't working with the current simsopt version, because a `.J` was missing


landreman (approved):
Thanks for the fix!

PR #461, opened 2024-11-25
Cbs/curvealongz

smiet:
I wanted to add a toroidal field to a configuration, because I just wanted to jiggle the rotational transform a bit without too much ado. That would be nice with a straight coil along the z-axis, so I coded it up using the JaxCurve. 

(Yes upon writing this PR I see that ToroidalBfield exists, but it does not allow one to shift the location of the curve along Z (playing with a 1/1 component), or optimize with it). 

Think this might be useful for others, @krystophny has mentioned wanting to do something similar but resorting to using very large radius coils. 

Some of the unittests are failing, mostly because the curve is not curved, and the evaluation of Kappa and it's derivatives fail. In that regard, what would be the best fix?

- overload the `kappa`, `torsion` etc functions and their derivatives in the `CurveAlongZ` class with a default choice. 
- adapt the base `Curve` methods of evaluation of `kappa`, `torsion` etc and their derivatives to deal with straight sections. 

option 1 would work, but lead to code repitition if other curves with straight sections are implemented
option 2 could lead to unneccesary computational overhead if every field evaluation `nan`s need to be replaced

@andrewgiuliani @florianwechsung, any suggestions as to the best approach?


krystophny:
@smiet I opt for option 1. There will be no code duplication if you extract the required functionality into lower level functions that are used by several classes. Performance and simplicity of class hierarchy are higher priority than implementation details in special cases. Can you add that?

src/simsopt/geo/curvealongz.py
```diff
@@ -0,0 +1,81 @@
...
+    gamma = jnp.stack((x, y, z), axis=1)
+    return gamma
+
+
+class CurveAlongZ(JaxCurve):
+    r'''
+    Straight vertical curve, parallel to the z-axis. 
+
+    Useful for quickly generating a toroidal field, comparing to tokamak equilibria
+    where an axisymmetric 1/R field is present, and testing.
```
andrewgiuliani, line 34: do you really want this approximation of an axisymmetric 1/R field? why not just implement the analytical magnetic field instead of this approximation?

Issue #389, opened 2024-01-17
MPILogHandler not working and logger example fails


smiet:
I am having issues running the MPILogHandler, which would be very useful for debugging/understanding some large runs. 

The example `examples/1_simple/logger_example.py` fails because 'initialize_logging'  should be imported directly from 'simsopt.util'. Fixing that, we get the error: 
```python
Traceback (most recent call last):
  File "/my_mount/1_Simple/logger_example.py", line 26, in <module>
    initialize_logging(mpi=True, filename='mpi.log')
  File "/venv/lib/python3.10/site-packages/simsopt/util/logger.py", line 46, in initialize_logging
    logging.config.dictConfig(config_dict)
  File "/usr/lib/python3.10/logging/config.py", line 811, in dictConfig
    dictConfigClass(config).configure()
  File "/usr/lib/python3.10/logging/config.py", line 572, in configure
    raise ValueError('Unable to configure handler '
ValueError: Unable to configure handler 'mpi_file_handler'
```
which, after some digging, I found out is due to a failed relative import of `from .._core/dev import SimsoptRequires` in `simsopt.util.mpi_logger.py`. 

This is probably because `MPILogHandler` is spawned in a child process that is only fed the file and cannot resolve the relative import as this requires rest of simsopt. 

Commenting out the `SimsoptRequires` produces acceptable log files, but now the logger does not exit when the script completes. 

does anyone have experience with MPI logging, and knows a fix? MPILogHandler seems based on 2013 code, there might be better solutions now... Looking for suggestions and willing to implement them when pointed in the right direction.

PR #462, opened 2024-12-04
Fix MPI Logging class and example

missing-user:
Fix the MPI logging example, as suggested by @smiet in https://github.com/hiddenSymmetries/simsopt/issues/389 
I did not experience the mentioned issue, about the logging process persisting on exit, so it seems fine to merge.
Add `logger_example.py` to the parallel test cases.


missing-user:
Sorry for wasting GH Action minutes, hopefully someone can cancel the job.
 
Apparently the mpi_logger process still doesn't exit in the action file, but I cannot reproduce the behavior on my machine.

PR #463, opened 2024-12-04
Respect prob.bounds in all solver wrappers

missing-user:
Only the constrained_solve overloads respected the problem bounds, the least_squares_solve overloads ignored them, unless passed with `bounds=prob.bounds` explicitly. This is unintuitive, since these functions are specifically overloads for `LeastSquaresProblem`, so I would expect it to use all the properties set on the object.

I suggest harmonizing the usage of `constrained_solve` and `least_squares_solve` to in both cases respect `prob.bounds` (as implemented in this PR), or at the very least raising a warning message to users, if `prob.bounds` were set, despite being ignored (I will adjust the PR if required).

### DIFF
--- a/src/simsopt/solve/mpi.py
+++ b/src/simsopt/solve/mpi.py
@@ -200,6 +200,12 @@ def least_squares_mpi_solve(prob: LeastSquaresProblem,
         logger.debug(f"residuals are {residuals}")
         return residuals
 
+    if "bounds" in kwargs:
+        import warnings
+        warnings.warn("The bounds argument has been deprecated and is being ignored, \
+                      please use prob.bounds instead.", DeprecationWarning, 2)
+        kwargs.pop("bounds", None)
+
     # For MPI finite difference gradient, get the worker and leader action from
     # MPIFiniteDifference
     if grad:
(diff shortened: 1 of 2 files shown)


missing-user:
I also replaced print with logger.info in two places that stood out to me.

missing-user:
Replaced deprecated np.product with np.prod for numpy 2.0 support

landreman:
Looks good to me. I think the failing tests are unrelated to the changes in this PR. Let's see if we can fix those in master. Anyone have ideas how to resolve those?

missing-user:
> Looks good to me. I think the failing tests are unrelated to the changes in this PR. Let's see if we can fix those in master. Anyone have ideas how to resolve those?

Thanks, it also seemed to me that way, tried looking into the issues with pyoculus myself but got sidetracked unfortunately.

landreman:
I think to correct my mistaken merge, it would work to do
~~~~
git reset --hard d473a37
git merge upstream master
~~~~
from the warn-unused-bounds branch, where d473a37 is the last intended commit before I merged. Let me know if you agree and if I should push this. Sorry again!

PR #464, opened 2024-12-16
Planar coil array branch merge

akaptano:
Hi all,

Here is a branch I have that is built off of some of the machinery from Siena's coil forces branch. Some notes:
1. The branch implements various pointwise and net force and torque calculations, and the total vacuum energy.
2. The branch implements a JaxCurvePlanarFourier class and fixes the Jacobian calculation (thanks Alex) for the CurvePlanarFourier class, which had a bug. 
3. It adds some geometry-related functions and various other functionality for effectively optimizing arrays of dipole coils, including with forces and torques. 
4. There is still a fair bit of stuff in the 3_Advanced/ examples directory to reproduce the paper I'm putting on arXiv shortly, which I can put its own Zenodo repo if need be. I also included Siena's Pareto-scan scripts, since I actually think future coil optimizers would find these helpful to not have to recreate. 
5. There are some lingering issues. As the number of coils gets large (~40), the number of Optimizable weak references seem to spiral out of control in any of the force/torque objectives that need to generate their own BiotSavart objects. I have been able to somewhat control this by regularly erasing weak references when making calls to this terms. This seems to have no bearing on the optimization itself. I have also not yet correctly implemented the fast version of the self inductance for a rectangular coil from Siena and Matt. 
6. I am planning on adding passive coil functionality, but figure I should stop here and continue on a branch split off from this one. 

Let me know if you have any questions, and thanks for any time you can spare to take a look!


landreman:
@akaptano on zoom the other day I believe you said this PR was going to be subsumed by another PR you're about to submit. Should we close this one then? Or still review it?

PR #467, opened 2025-01-08
Fix CI

landreman:
The CI appears to be broken. For the extensive tests with python 3.10 and 3.11 but not 3.9, there is a failure early on when pyoculus is installed.

This PR doesn't yet have a fix, just documenting that the problem occurs even for the master branch.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -160,7 +160,6 @@ target_include_directories(profiling PRIVATE  "thirdparty/xtensor/include" "thir
 target_link_libraries(profiling PRIVATE fmt::fmt-header-only)
 
 
-
 #get_cmake_property(_variableNames VARIABLES)
 #list (SORT _variableNames)
 #foreach (_variableName ${_variableNames})


landreman:
The error appears to be related to f2py:
~~~~
    File "/tmp/pip-build-env-78kkn94r/overlay/lib/python3.11/site-packages/numpy/f2py/capi_maps.py", line 377, in getpydocsign
      sig = '%s : %s %s%s' % (a, opt, c2py_map[ctype], init)
                                      ~~~~~~~~^^^^^^^
  KeyError: 'void'
~~~~
I seem to remember this error appearing in the simsopt CI also last summer when numpy v2 was released, but I'm not certain. @smiet @mbkumar any ideas how to fix this?

mbkumar:
Looks like numpy>=2.0 issue with pyoculus. I solved a similar issue for VMEC. Perhaps it could be replicated here. I'll do a more thorough check and get back to you.

mbkumar:
In VMEC, we moved away from numpy.distutils which is deprecated and not working properly anymore. Pyoculus is still using numpy.distutils. This approach should be updated and a more robust `setup` function such as the one from scikit-build should be used.
@smiet used a `meson` build system for SPEC. Perhaps he may be interested in using the same one for PyOculus. Otherwise, I can take the lead and update the build system to a new one based on `scikit-build` and `cmake`.

smiet:
There is a branch with meson-build, but it still has an issue (see: [this issue on pyoculus](https://github.com/zhisong/pyoculus/issues/12). 

Issue seems to originate from numpy 2.1 or 2.2 change.  Have not reproduced the issue myself, but we might need to clamp 2.0 <= numpy <= 2.1. 

I have currently clamped the numpy version used for the pyoculus build to numpy 2.0, this should fix it temporarily, hopefully until the new pyoculus version I am working on is ready for prime-time.

smiet (changes requested):
Temporarily fixed by change to pyoculus that clamps build version of numpy to 2.0 (this allows backwards compatibility, but should also work for people using numpy >2.0 at runtime). 

Not a complete fix, but will hold until I return to work on Monday and I can figure out the issue with numpy>=2.1

We can close this PR

landreman:
Some updates:
The test
~~~~
test_iotas_derivative (geo.test_surface_objectives.IotasTests)
~~~~
is now passing - I just slightly loosened a tolerance from 0.3 to 0.31. The test
~~~~
test_Bn (geo.test_pm_grid.Testing)
~~~~
is now passing, after loosening an atol slightly from 0 to 1e-15. That leaves 3 failing tests.

The failing tests in
~~~~
test_boozer_surface_optimisation_convergence (geo.test_boozersurface.BoozerSurfaceTests)
test_convergence_cpp_and_notcpp_same (geo.test_boozersurface.BoozerSurfaceTests)
~~~~
for python 3.10 and 3.11 are still not understood. In the branch [ml/20250113-debug_CI](https://github.com/hiddenSymmetries/simsopt/tree/refs/heads/ml/20250113-debug_CI) the `Tests` CI runs only these tests so we can debug them faster. Feel free to push to this branch for debugging.

The test failure
~~~~
test_curve_first_derivative (geo.test_curve_optimizable.Testing)
~~~~
for all python versions is also not understood - now `obj.J()` returns nan. In the `20250113-debug_CI` branch, this test is run by itself, and it passes. How could the result of `obj.J()` depend on whether or not other tests are run?

smiet:
Given that a few tests are solved with fixes to the tolerances, this might be related to changes in data type promotion in numpy 2.x . 
Though during execution numpy 1.26 is used, the C-API is compiled using numpy 2.2. 
There have been changes to these parts, and the 2.2.1 update specifically mentions breakage in downstream projects due to typing changes. 

In some cases this causes lower-precision datatypes which could fail the tests. Though this is still strange, as runtime 1.26 should keep the promotion the same. 

see: https://numpy.org/neps/nep-0050-scalar-promotion.html#nep50

the numpy2.0 migration guide does list a way to raise a warning when this happens: https://numpy.org/doc/2.2/numpy_2_0_migration_guide.html#changes-to-numpy-data-type-promotion 

We could test `np.set_promotion_state90('weak_and_warn')` (or `'legacy'`) in the failing tests to see if this is the cause. 

Alternatively fixing the build-time numpy version to 2.0 is potentially a short-term fix that I will try now.

landreman:
I agree with @mbkumar that the ubuntu version must be part of the story. In [this run of the ml/20250113-debug_CI branch](https://github.com/hiddenSymmetries/simsopt/actions/runs/12871897346), which directly compares ubuntu-22.04 vs ubuntu-latest, the `test_boozer_surface_optimisation_convergence` tests all pass in the former and fail in the latter. The default ubuntu version in the CI changed in December, so the timing is right. Some differences between the os versions are documented [here](https://github.com/actions/runner-images/issues/10636). Looking at all the versions of python packages at the end of the "Install simsopt package" step, it looks like they are all the same between the older vs newer ubuntu runs, so that is not the issue. While we could have the CI use the older ubuntu version for now, it would be better to get the CI working with the latest ubuntu version.

In that CI run, I printed out the residual vs iteration for those failing tests, and it looks like Newton's method is diverging rather than converging. Example:
~~~~
!!!!!! Testing surfacetype SurfaceXYZTensorFourier, stellsym True, optimize_G True, second_stage newton, get_data <function get_giuliani_data at 0x7f87650a5510>, vectorize False
Residual norm after LBFGS 682 2.294509402657461e-03
minimize_boozer_penalty_constraints_newton iter: 0  val: 2.6323866994417486e-06  norm: 0.0006022726438895579
minimize_boozer_penalty_constraints_newton iter: 1  val: 0.00039005450363802563  norm: 0.5041954646638603
minimize_boozer_penalty_constraints_newton iter: 2  val: 5.094615223534711e-06  norm: 0.007185679457999476
minimize_boozer_penalty_constraints_newton iter: 3  val: 4.089872578117214e-05  norm: 0.14172172054575086
minimize_boozer_penalty_constraints_newton iter: 4  val: 8.809182188159334e-06  norm: 0.01242660254456349
minimize_boozer_penalty_constraints_newton iter: 5  val: 0.00013128372620558918  norm: 0.2830093092565208
minimize_boozer_penalty_constraints_newton iter: 6  val: 1.6468981915618705e-05  norm: 0.013756818781153798
minimize_boozer_penalty_constraints_newton iter: 7  val: 10.157595247966423  norm: 194.07052961535098
minimize_boozer_penalty_constraints_newton iter: 8  val: 1.3450970606902033  norm: 50.677055615637116
minimize_boozer_penalty_constraints_newton iter: 9  val: 128.27102807457686  norm: 1449.0650971488906
minimize_boozer_penalty_constraints_newton iter: 10  val: 153.97224755727544  norm: 1855.6508088107519
minimize_boozer_penalty_constraints_newton iter: 11  val: 2779207607344.2334  norm: 84211407168.62497
minimize_boozer_penalty_constraints_newton iter: 12  val: 548978693020.2609  norm: 24951522381.755825
minimize_boozer_penalty_constraints_newton iter: 13  val: 108439976999.2  norm: 7393039889.858839
minimize_boozer_penalty_constraints_newton iter: 14  val: 21420127411.04032  norm: 2190527818.519551
minimize_boozer_penalty_constraints_newton iter: 15  val: 4231085186.618325  norm: 649043599.9812926
minimize_boozer_penalty_constraints_newton iter: 16  val: 835747207.6700873  norm: 192308094.79325154
minimize_boozer_penalty_constraints_newton iter: 17  val: 165075792.47637716  norm: 56979388.17807232
minimize_boozer_penalty_constraints_newton iter: 18  val: 32786359.931228556  norm: 16899121.641762033
minimize_boozer_penalty_constraints_newton iter: 19  val: 6599061.6332561895  norm: 5108009.228141096
minimize_boozer_penalty_constraints_newton iter: 20  val: 1373211.6883342883  norm: 1574328.522892679
minimize_boozer_penalty_constraints_newton norm: 1574328.522892679  tol: 1e-10  success: False
~~~~
@andrewgiuliani @mbkumar can you think of any reason that Newton iterations would diverge for these tests due to a change in os/gcc version? Do you have any suggestions how to fix this?

The test `geo.test_boozersurface.BoozerSurfaceTests.test_convergence_cpp_and_notcpp_same` is fixed now by slightly loosening a tolerance.

For the other failing test, `geo.test_curve_optimizable.Testing.test_curve_first_derivative`: By printing info at curveobjectives.py line 43 in the CI, it looks like the problem is coming from `self.curve.incremental_arclength()`, which for `RotatedCurve` is returning NaNs. Any ideas why?

mbkumar:
Thanks Matt for the detailed investigation. I don't have answers now. I'll
start debugging these issues from tomorrow.

akaptano:
@landreman can we close this? I think this is solved by #486

landreman:
CI is finally passing with several recent PRs like #580. Some of the issues here were addressed in #486

PR #470, opened 2025-01-13
Support for current profiles that set I(s) directly instead of I'(s).

daringli:
Simsopt currently assumes that for any current profile, curtor in the input file is given by the integral of the profile.
However, there are advanced profiles (see https://princetonuniversity.github.io/STELLOPT/VMEC%20Advanced%20Profiles) that set I(s) directly, i.e. the total current inside the flux-surface s. For such profiles, no integration is necessary. They are also often smoother than the current density, which might make it easier to represent them with fewer degrees of freedom and thus render them suitable for optimization.

This simple pull request implements support for such profiles by adding a code-path where curtor is set by the value of the profile at s=1.


landreman:
Can you add a sentence or 2 about this behavior to the docstring of the Vmec class?

daringli:
> Can you add a sentence or 2 about this behavior to the docstring of the Vmec class?

Added! Interestingly, the docstring already mentioned "cubic_spline_i", despite this not being properly supported prior to this pull request.

landreman:
It looks like the test `mhd.test_vmec.VmecTests.test_current_profile` is failing as a result of the change here. Can you sort this out?

daringli:
> It looks like the test `mhd.test_vmec.VmecTests.test_current_profile` is failing as a result of the change here. Can you sort this out?

Thanks for pointing this out.
It looks like the problem was caused by vmec.vmec_input.pcurr_type being a binary string rather than just a string. I fixed this and also added a test case for setting a current profile with pcurr_type 'cubic_spline_i', which covers the new code path added by this pull request.

landreman (approved):
Looks good, thanks for this contribution

PR #473, opened 2025-02-02
Fix wheel workflow

landreman:
The wheel workflow recently [stopped working](https://github.com/hiddenSymmetries/simsopt/actions/runs/13074530769). With this PR it is working again. I've mainly followed [this example from cibuildwheel](https://github.com/pypa/cibuildwheel/blob/main/examples/github-deploy.yml). 

The workflow takes longer to run now because wheels are built for more platforms now, in particular `musllinux` and linux `i686` (in addition to `manylinux` and `x86_64`).  This is done by default in the cibuildwheel example workflow that I copied, and the wheels were built without error, so I kept these platforms. We could potentially remove these cases if nobody would use them and the longer workflow runtime is a problem.

### DIFF
--- a/.github/workflows/wheel.yml
+++ b/.github/workflows/wheel.yml
@@ -1,5 +1,7 @@
 name: Wheel Builder
 
+# This file is based on https://github.com/pypa/cibuildwheel/blob/main/examples/github-deploy.yml
+
 on: [push, pull_request]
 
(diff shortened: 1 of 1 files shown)


mbkumar (approved):
Looks good. All the arch variations are captured

PR #486, opened 2025-04-25
Use ubunut 24.04 in place of other ubuntu runners

mbkumar:
This PR is to push code changes to make the CI run on Ubuntu 24.04. This serves two purposes.
1. To ensure, our code is not broken on Ubuntu 24.04 so users can install simsopt on Ubuntu 24.04
2. Helps devs to use the `ubuntu-latest` runner for GitHub Actions.

### DIFF
--- a/docs/source/example_islands.rst
+++ b/docs/source/example_islands.rst
@@ -47,7 +47,7 @@ configuration contains only modes up to m=3, we must increase the
 number of modes in the boundary shape in order to have m=6 modes
 available to vary::
 
-  s.boundary.change_resolution(6, s.boundary.ntor)
+  s.boundary = s.boundary.change_resolution(6, s.boundary.ntor)
 
 Now we can pick out a few modes of the boundary shape to vary in the
 optimization::
--- a/examples/2_Intermediate/eliminate_magnetic_islands.py
+++ b/examples/2_Intermediate/eliminate_magnetic_islands.py
@@ -33,7 +33,7 @@ s = Spec(os.path.join(os.path.dirname(__file__), 'inputs', 'QH-residues.sp'),
          mpi=mpi)
 
 # Expand number of Fourier modes to include larger poloidal mode numbers:
-s.boundary.change_resolution(6, s.boundary.ntor)
+s.boundary = s.boundary.change_resolution(6, s.boundary.ntor)
 # To make this example run relatively quickly, we will optimize in a
 # small parameter space. Here we pick out just 2 Fourier modes to vary
 # in the optimization:
--- a/examples/2_Intermediate/tracing_boozer.py
+++ b/examples/2_Intermediate/tracing_boozer.py
@@ -69,8 +69,7 @@ modB_inits = field.modB()
 Ekin = 100000*ONE_EV
 mass = PROTON_MASS
 vpar = np.sqrt(2*0.8*Ekin/mass)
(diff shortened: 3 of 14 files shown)


akaptano:
Happy to get this over the finish line, but what is the advantage of this change?

andrewgiuliani:
@mbkumar weren't the failing tests happening on Ubuntu 24.04? why would forcing 24.04 fix the issue?

mbkumar:
They are happening on 24.04, but we are not forcing 24.04 on people. It is
the other way, people want to run simsopt on 24.04. So we need to resolve
the issues. This PR is to address that.
Bharat Medasani

akaptano:
Writing some notes about the bug here:

1. The issue is essentially that L-BFGS seems to arrive at a different minima (a very poor one) with the non-vectorized. I cannot reproduce this with any version of python on my Mac. My speculation is that there is some ubuntu-specific issue with the python version of BFGS. Because L-BFGS returns a nonsense minima, the Newton method blows up in the next step of the unit test in test_boozersurface. This appears to be very robust behavior, not just from roundoff. I tried adding regularization and damping to the Newton solve, having the BFGS solve do several solves from slightly different initial conditions and take the best output, and other tricks that should work if this was just a roundoff error. None of this was able to fully avoid the fundamental issue that BFGS was returning very poor minima that either make Newton blow up or don't pass the checks. My guess is that there is a real bug in the ubuntu implementation somehow. 

2. Apart from this, there is also a weird difference in how the vectorized and non-vectorized versions normalize the function values/Jacobian/Hessian. In boozer_penalty_constraints it is:
boozer = tuple([b/np.sqrt(num_res) for b in boozer])
and in  boozer_penalty_constraints_vectorized it is:
boozer = tuple([b/num_res for b in boozer])
The normalization also doesn't happen at the same point in the constraint set-up. I fixed this during debugging (although I have since reverted to keep changes minimal) and the CI issues persisted (although this does seem to reduce the variability between the two version of the problem). 

3. Given the vectorized function works and the non-vectorized doesn't, but they do the same thing + the vectorized is faster, I recommend that we deprecate the non-vectorized. I have also changed it so the CI skips the vectorize=False convergence tests in the CI. I have added a warning when minimize_boozer_penalty_constraints_LBFGS is used with vectorize=False.

akaptano:
@andrewgiuliani @landreman @mbkumar Assuming that we can chop the non-vectorized boozersurface stuff, this is ready for a merge. However, our CI troubles are not behind us. Deactivating the SPEC checks in the CI is still keeping the test coverage fairly low. Switching to ubuntu24 also necessitated dropping python 3.12 runs -- there seems to be some complication getting pip to be recognized correctly in the CI environment. This appears to be an issue with the github CI, not with our code (I tried many fixes to no avail). So this is not super satisfying beyond fixing the boozersurface issues which were the biggest issue.

.github/workflows/extensive_test.yml
akaptano: Can simplify the logic in the VMEC/SPEC coverage checks, don't need the last one especially

mbkumar:
Looks like docker website is broken. Seeing failures to access the container registry in `master` as well.

landreman:
Looks fine to me, but it would be nice to hear from @andrewgiuliani if the changes to the BoozerSurface machinery are ok.

PR #490, opened 2025-05-05
Bug fix for adding dof names in CurveRZFourier Class

jhalpern30:
This bug-fix is part of **Simsopt fix-it week 2025**. 
It addresses a bug in the class `CurveRZFourier`. According to the documentation, the degrees of freedom (dofs) should be listed as 

```[r_{c,0}, \cdots, r_{c,\text{order}}, r_{s,1}, \cdots, r_{s,\text{order}}, z_{c,0},....]```

if `stellsym` is false, and

```[r_{c,0},...,r_{c,order},z_{s,1},...,z_{s,order}]```

if stellsym is True. 

However, a dof naming function did not exist so the dofs were being stored as 

```[x0, x1, /cdots, x{num_dofs]```.

This pull request fixes this issue by adding a `_make_names` function, which names the dofs according to the documentation. It is modeled after the existing function used to names the dofs in `CurveXYZFourier`. Its functionality has been tested for both stellarator symmetric and non-stellarator symmetric curves below

---------------------------------

**Test 1:**
Non-stellarator symmetric, order = 3: dofs are stored as [r_{c,0}, \cdots, r_{c,\text{order}}, r_{s,1}, \cdots, r_{s,\text{order}}, z_{c,0},....]
image

**Test 2:**
Stellarator symmetric, order = 6: dofs are stored as [r_{c,0},...,r_{c,order},z_{s,1},...,z_{s,order}]
image

As a comparison, this is the output for a CurveXYZFourier object where the dofs are appropriately named and stored as 
```[x_{c,0}, x_{s,1}, x_{c,1},\cdots x_{s,\text{order}}, x_{c,\text{order}},y_{c,0},y_{s,1},y_{c,1},\cdots]```
image

### DIFF
--- a/src/simsopt/geo/curverzfourier.py
+++ b/src/simsopt/geo/curverzfourier.py
@@ -38,19 +38,15 @@ class CurveRZFourier(sopp.CurveRZFourier, Curve):
             Curve.__init__(self, x0=self.get_dofs(), names=self._make_names(order, stellsym),
                            external_dof_setter=CurveRZFourier.set_dofs_impl)
         else:
-            Curve.__init__(self, dofs=dofs, 
+            Curve.__init__(self, dofs=dofs,
                            external_dof_setter=CurveRZFourier.set_dofs_impl)
 
     def _make_names(self, order, stellsym):
-        r_names = ['rc(0)']
-        r_names += [f'rc({i})' for i in range(1, order + 1)]
+        r_names = [f'rc({i})' for i in range(0, order + 1)]
+        z_names = [f'zs({i})' for i in range(1, order + 1)]
         if not stellsym:
             r_names += [f'rs({i})' for i in range(1, order + 1)]
-        z_names = []
-        if not stellsym:
-            z_names += ['zc(0)']
-            z_names += [f'zc({i})' for i in range(1, order + 1)]
-        z_names += [f'zs({i})' for i in range(1, order + 1)]
+            z_names = [f'zc({i})' for i in range(0, order + 1)] + z_names
         return r_names + z_names
 
     def get_dofs(self):
--- a/src/simsopt/geo/curverzfourier.py
+++ b/src/simsopt/geo/curverzfourier.py
@@ -38,8 +38,8 @@ class CurveRZFourier(sopp.CurveRZFourier, Curve):
             Curve.__init__(self, x0=self.get_dofs(), names=self._make_names(order, stellsym),
                            external_dof_setter=CurveRZFourier.set_dofs_impl)
         else:
-            Curve.__init__(self, dofs=dofs,
-                           external_dof_setter=CurveRZFourier.set_dofs_impl)
+            Curve.__init__(self, external_dof_setter=CurveRZFourier.set_dofs_impl,
+                           dofs=dofs)
 
     def _make_names(self, order, stellsym):
         r_names = [f'rc({i})' for i in range(0, order + 1)]
--- a/src/simsopt/geo/curverzfourier.py
+++ b/src/simsopt/geo/curverzfourier.py
@@ -1,4 +1,5 @@
 import numpy as np
+from itertools import chain
 
 import simsoptpp as sopp
 from .curve import Curve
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/simsopt/geo/curverzfourier.py [hunk 1]
```diff
@@ -35,11 +35,23 @@ def __init__(self, quadpoints, order, nfp, stellsym, dofs=None):
...
+                           external_dof_setter=CurveRZFourier.set_dofs_impl)
         else:
-            Curve.__init__(self, external_dof_setter=CurveRZFourier.set_dofs_impl,
-                           dofs=dofs)
+            Curve.__init__(self, dofs=dofs, 
+                           external_dof_setter=CurveRZFourier.set_dofs_impl)
+
+    def _make_names(self, order, stellsym):
+        r_names = ['rc(0)']
+        r_names += [f'rc({i})' for i in range(1, order + 1)]
```
mishapadidar, line 46: can line 45 and 46 be made into one line?

src/simsopt/geo/curverzfourier.py
```diff
@@ -35,11 +35,23 @@ def __init__(self, quadpoints, order, nfp, stellsym, dofs=None):
...
+        if not stellsym:
+            r_names += [f'rs({i})' for i in range(1, order + 1)]
+        z_names = []
+        if not stellsym:
```
mishapadidar, line 50: can we use only one if statement?
```
if not stellsym:
    r_names = ...
    z_names = ...
```

mishapadidar:
looks good to me!

mishapadidar (approved):
looks good to me

andrewgiuliani:
Hi Jake, all looks good to me, but could you write a unit test to verify that the dof names ordering corresponds to what is actually done in the code.  i.e. something like:

```
rc0 = curve.get('rc(0)')
idx = np.where(['rc(0)' in curve.dof_names])[0]
assert curve.x[idx] == rc0
assert curve.rc[0] == rc0
```

for a few dofs in rc, zs, zc, zs?

src/simsopt/geo/curverzfourier.py
```diff
@@ -35,11 +35,19 @@ def __init__(self, quadpoints, order, nfp, stellsym, dofs=None):
...
         if dofs is None:
-            Curve.__init__(self, external_dof_setter=CurveRZFourier.set_dofs_impl,
-                           x0=self.get_dofs())
+            Curve.__init__(self, x0=self.get_dofs(), names=self._make_names(order, stellsym),
+                           external_dof_setter=CurveRZFourier.set_dofs_impl)
         else:
-            Curve.__init__(self, external_dof_setter=CurveRZFourier.set_dofs_impl,
-                           dofs=dofs)
+            Curve.__init__(self, dofs=dofs,
+                           external_dof_setter=CurveRZFourier.set_dofs_impl)
```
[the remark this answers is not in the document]
jhalpern30: Fixed this

jhalpern30:
> Hi Jake, all looks good to me, but could you write a unit test to verify that the dof names ordering corresponds to what is actually done in the code. i.e. something like:
> 
> ```
> rc0 = curve.get('rc(0)')
> assert curve.rc[0] == rc0
> ```
> 
> for a few dofs in rc, zs, zc, zs?

I added a unit test specifically for CurveRZFourier. This is the first unit test I've written, so let me know if there's anything I need to fix. 

I looked into making it more a more general test for all curves, but ran into two issues:

1. The existing function that loops through the curve object ```get_curve``` with ```curvetypes``` listed in line 139 of ```test_curve.py``` only considers a stellarator symmetric form of the coil, and I think testing on both a ```stellsym=True``` and ```stellsym=False``` coil is necessary to effectively test. 
2. All curves still don't have the dofs properly named, making looping through this array more difficult. It looks like after this PR, ```CurveXYZFourier, CurveXYZFourierSymmetries```, and ```CurveRZFourier``` will have named dofs, but other classes such as ```CurvePlanarFourier, JaxCurveXYZFourier```, and ```CurveHelical``` still don't have this feature. I think this is necessary to add first before making this a more general unit test. 

On a similar note, I think once I fix the ```SurfaceXYZFourier``` and ```SurfaceXYZTensorFourier``` classes (in progress), all surfaces named in ```surface_types``` within ```test_surface.py``` will have appropriately named dofs, so I could look into a more general unit test there.

mishapadidar (approved):

andrewgiuliani:
Thanks @jhalpern30 I'm ok with keeping the unit test you've written simple, and specialized to `CurveRZFourier`.  Let's wait until the unit tests + code coverage finish.

mbkumar:
> @mbkumar any idea why the docker unit test is failing here?

Pull request from forked simsopt, not from a branch on HS simsopt. His account does not have the login privileges because the workflow can't use the secrets from HS Github org. You can ignore the failure.

PR #494, opened 2025-05-06
Surface self intersecting dependencies

daringli:
Surface.is_self_intersecting() requires packages 'ground' and 'bentley_ottman'. These dependencies were previously undocumented and also not installed in the containers. This PR adds these dependencies to a new optional dependency category 'ALGS', and also installs them in the docker and shifter containers.


andrewgiuliani (commented):
Taking a quick look at simsopt's additional optional dependency, I noticed that pyQSC  could be included in ALGS too.

https://github.com/hiddenSymmetries/simsopt/blob/53c4f29aea3e92133aed7ab7338f7923419febb9/src/simsopt/geo/surfacerzfourier.py#L481

looks good to me in any case, but I'd like @mbkumar to take a look to.

andrewgiuliani (approved):

daringli:
> Taking a quick look at simsopt's additional optional dependency, I noticed that pyQSC could be included in ALGS too.
Added. It was already installed in the containers, so only pyproject.toml was changed.

mbkumar (approved):

mishapadidar (approved):
looks good to me.

PR #493, opened 2025-05-05
Developer Guides with Derivative Tutorial

mishapadidar:
**PR part of Fix-It Week 2025**

In this PR, we add a new section to the ReadTheDocs sidebar: the "Developer Guides". The Developer Guides section is meant to hold tutorials to help developers build code with SIMSOPT's features. As the first tutorial in the developer guides we added the "Using the Derivative class" tutorial, which teaches developers how to build an objective function with derivatives in SIMSOPT. Learning how to add derivatives to Optimizable objects has been a challenge for many developers, in our experience.

Attached are two screenshots showing the new entry in the sidebar and the tutorial.

dev_guides
deriv_tut

\
\
**Tests**
No tests are needed since this is a documentation fix.

PR #495, opened 2025-05-06
Adds C++ API's to simsopt documentation

mbkumar:
This PR adds Simsopt++ C++ API's to simsopt documentation. Please take a look at https://simsopt.readthedocs.io/en/doxygen/. The C++ documentation appears at Simsopt++ C++ API and Simsopt++ C++ Internals buttons at the bottom of the left hand pane. Needs to be coordinated with #493 to make the organization in the Right Hand Panel consistent.


mishapadidar:
I like the idea of a `For Develoeprs` section. I have a few general comments on organization. 

1. We have two sections related to the API: `API Reference: Public Functions and Classes` and `Full API Listing`. Why do we have both? Is it possible to just have a single API?
2. If the `Full API Listing` was moved to the API section, then everything in the `For Developers` section is related to C++.
3. Generally, what do we see as the main use case of the `For Developers` section? Is it for reference, learning, links? There are other related sections like `contributing to simsopt` that are not in this section, probably because they have a different purpose.

mbkumar:
1. In the User API Reference, we don't list some functions/classes of
simsopt that are not relevant for users. Mainly to simplify the learning
curve for the users. Full API listing shows all the classes and functions
of simsopt and useful for developers.
2. Contributing to simsopt points to users on how to contribute to simsopt.
Treat it as a stepping stone for users to become developers.
3. We can expand "For Developers" with additional content. There are few
features in simsopt that we don't expand on, such as jax usage, XSIMD based
optimization, caching mechanism in C++. In future we can expand on these
topics in "For Developers" documentation.

Bharat Medasani

mishapadidar:
Thanks for the responses, those are great ideas. 
I think that we could remove the Public API and just have the full API in the API section. Perhaps we should get others to chime in here though, @andrewgiuliani @landreman @smiet 
Everything else looks good to me.

andrewgiuliani:
Is there a simple way to go from a function in the documentation directly to the corresponding code on github? I have never figured out a simple way to do this, so I typically just read directly the code + docstrings on github rather than use these readthedoc pages.

can we merge with master so can observe what it will look like with #493 now live?

mbkumar:
That's sounds good

mbkumar:
@andrewgiuliani 
Screenshot 2025-05-07 at 7 41 29 PM

mishapadidar:
Two changes:
1. Can we move the Simsopt++ C++ API into the API section and rename it `Simsopt C++ API`?
2. Can we rename `simsopt package` in the API Reference to `Simsopt Python API`?

Otherwise the changes all look good

mbkumar:
Done. 
https://simsopt.readthedocs.io/en/doxygen/

smiet (approved):
LGTM!

PR #498, opened 2025-05-06
ran run_autopep for the main branch.

akaptano:
Just ran the ./run_autopep script in the main directory and linted the master branch.


mishapadidar (commented):
Thanks for running this, Alan. 

I believe we have a linter running during the CI. Does this linter conform to the same style guidelines as autopep? @mbkumar @landreman 

It is a bit concerning to see so many files changed by the linter. I am certain that many of these files were linted before their last edit. This suggests that there may have been a change to the autopep style rules, either on our end or theirs. For example, in many files, autopep is removing trailing blank spaces.

mbkumar:
Not sure why that many issues were not caught before by the linter. Anyhow the fixes by autopep are in correct direction. So lets merge these.

landreman (approved):

mbkumar:
I looked into this issue. Trailing whitespace is only a warning and not an error. So this is not treated as an error by `black` and `ruff`.

PR #499, opened 2025-05-06
Updates Vmec to change nfp when a boundary with a new nfp is assigned. Adds tests for the boundary setter.

daringli:
This PR does a few things related to VMEC boundaries:
- Fixes a bug where Vmec would not update nfp when a boundary with a new nfp was assigned.
- Fixes a bug where get_max_mn() would not be updated after assigning a new boundary unless set_indata() was called.
- Adds tests for assigning a new boundary. This increases the test coverage.
- Adds a test for free boundary VMEC. This test is current skipped and is also untested since free boundary VMEC crashes on my laptop when run through simsopt¹. This will be fixed tomorrow. Apart from this, the PR is ready for review.
- Linted the affected files.
 
¹: Note that the external VMEC2000 does not segfault, but raises a ton of warnings.
```
WARNING: MPI_IN_PLACE in vacuum.f
 WARNING: MPI_IN_PLACE in scalpot.f 2
```


daringli:
Closed since we opened a new one

PR #500, opened 2025-05-07
Stop building wheels for non-HPC architectures

mbkumar:
Currently wheels are being built for i68c and musllinux architectures. Not worth building wheels not going to be used.


landreman (approved):
Looks fine. In #473 I remember that when the i686 and musllinux wheels were introduced, it significantly increased the time required for the wheel CI to run.

PR #502, opened 2025-05-07
(Fix-It Week) Update VMEC object to throw an error if a boundary with a different nfp is assigned

mishapadidar:
**Summary**
This pull-request throws an error if a boundary is assigned to a VMEC object that has an NFP that is different from the NFP of the VMEC object.

Only two lines of code need to be added. 

**Tests**
A unit test is added to test this feature.


tests/mhd/test_vmec.py
```diff
@@ -685,78 +687,33 @@ def test_write_input(self):
...
+            self.assertAlmostEqual(A1, 4.3549675967508055, places=7,
+                                   msg='Aspect ratio calculation does not match precalculated result.')
+            self.assertAlmostEqual(A2, 10.758596041584697, places=7,
+                                   msg='Aspect ratio calculation does not match precalculated result.')
+            self.assertEqual(
+                mn1, (4, 3), msg='max_mn does not match precalculated results')
+            self.assertEqual(
+                mn2, (11, 12), msg='max_mn not recalculated correctly after boundary updated')
+            self.assertEqual(
+                v.indata.nfp, 5, msg='nfp not updated correctly when boundary updated.')
```
andrewgiuliani, line 715: can you add a few more tests checking e.g. mean_iota, edge_iota, and axis_iota are the same as a precalculated ones? Right now you're only checking the aspect ratio is correct, but what about the actual VMEC solution?

andrewgiuliani (commented):
I would have thought changing the nfp would mess with the VMEC fortran internals...I think we should have a few more unit tests to ensure that we're not missing something

tests/mhd/test_vmec.py
andrewgiuliani: won't changing NFP clash with the input file used to initialize the vmec object?

daringli: I added an iota_edge check. But I'm pretty sure we are properly updating the VMEC 'internals'.
Reason 1: in some sense, VMEC is fully controlled by the input file, and it is correct.
Reason 2: we do update the VMEC2000 python-package singleton (which is how the value is updated in the input file). I don't think there is any other internal?
It will overwrite with the original value of nfp in the input file, but I think this is OK. NFP is basically a surface property that controls how the surface modes are interpreted.

tests/mhd/test_vmec.py
```diff
@@ -685,78 +687,38 @@ def test_write_input(self):
...
+            v.boundary = nfp5_boundary
+            mn2 = v.get_max_mn()
+            A2 = v.aspect()
+            iota_edge = v.iota_edge()
+            
+            self.assertAlmostEqual(A1, 4.3549675967508055, places=7,
+                                   msg='Aspect ratio calculation does not match precalculated result.')
+            self.assertAlmostEqual(A2, 10.758596041584697, places=7,
+                                   msg='Aspect ratio calculation does not match precalculated result.')
+            self.assertAlmostEqual(iota_edge, 1.62182836599054, places=7,
```
andrewgiuliani, line 712: Is this value of iota edge calculated from VMEC as if you had loaded a different input file with the modified value of nfp?

tests/mhd/test_vmec.py
andrewgiuliani: I'm not fully familiar with python VMEC implementation, maybe someone with more experience with it can chime in on this.

@mbkumar  would you know if modifying nfp as is done here is safe to do and won't mess with anything on the VMEC Fortran side?

src/simsopt/mhd/vmec.py
```diff
@@ -515,6 +515,7 @@ def set_indata(self):
         vi = vmec.vmec_input  # Shorthand
         # Convert boundary to RZFourier if needed:
         boundary_RZFourier = self.boundary.to_RZFourier()
+        vi.nfp = boundary_RZFourier.nfp
```
andrewgiuliani, line 518: What happens if boundary_RZFourier has a different value of mpol and/or ntor than what VMEC was initialized with? Can this be accounted for here too?

tests/mhd/test_vmec.py
daringli: It's calculated using the external VMEC2000 binary using an input file that I have modified myself to have the new boundary and the new nfp.

src/simsopt/mhd/vmec.py
daringli: It's something worth thinking about. Right now, mpol and ntor are not updated. Which means that we may be truncating modes. This truncation may happen at any time if the values of mpol and ntor are not consistent with the mpol and ntor of the surface; it's not solely something that could happen due to assigning a new boundary.
Our VMEC inputs do not generally have to have mpol and ntor consistent with the boundary; they are parameters the user has to update if they don't want to truncate modes.
I personally think this is potentially harmful and unintuitive, but it is the approach SIMSOPT has taken so far and there are legitimate use-cases for scanning 'surface complexity' by just changing ntol and mpol while keeping the boundary fixed.

tests/mhd/test_vmec.py
daringli: I agree. I'm fairly certain it is OK for fixed-boundary VMEC; I don't think nfp is used for anything but the boundary in this case. I am much less certain that it's OK for free-boundary VMEC, as nfp could have implications for how the coils are interpreted.
A more conservative approach might be to just throw an error if the user tries to assign a boundary with a different nfp. In practice, I've never wanted to do this and I can't think of a compelling use-case.

tests/mhd/test_vmec.py
```diff
@@ -685,78 +687,45 @@ def test_write_input(self):
...
-
-        equil.finalize()
-"""
-
+    def test_update_boundary(self):
+        """Test that the VMEC is able to update the boundary correctly.
+        Runs with very few iterations for quick CI.
+        Previously, get_max_mn() would not updated immediately upon
+        """
+        #with ScratchDir("."):
```
mbkumar, line 695: Can you please use the Scratchdir? We don't want to leave test stubs hanging.

tests/mhd/test_vmec.py
```diff
@@ -24,6 +24,8 @@
 from simsopt.util.constants import ELEMENTARY_CHARGE
 
 from simsopt.mhd.vmec import Vmec
+from simsopt.configs import get_w7x_data
+
```
mbkumar, line 29: Delete unused imports

src/simsopt/mhd/vmec.py
```diff
@@ -393,6 +393,9 @@ def boundary(self):
     def boundary(self, boundary):
         if boundary is not self._boundary:
             logging.debug('Replacing surface in boundary setter')
+            #TODO: should mpol and ntor be updated to match the new boundary?
+            if boundary.nfp != self.boundary.nfp:
+                raise ValueError('Trying to assign a new boundary to VMEC object, but the new boundary has an nfp diffrent from the one in the VMEC object.')
             self.remove_parent(self._boundary)
```
mbkumar, line 399: Can't we update nfp in the VMEC object?

mbkumar (changes requested):

src/simsopt/mhd/vmec.py
daringli: For fixed boundary VMEC, I'm fairly sure we can since I think NFP is only used to interpret the boundary, so changing NFP when changing the boundary makes sense.
For free-boundary, I don't know. Would it change how the coils are interpreted? If you think it makes sense, we certainly can updated NFP (and this was done in an earlier version of the PR).

tests/mhd/test_vmec.py
daringli: Oops! Fixed.

src/simsopt/mhd/vmec.py
daringli: Can do!
Although I should stress that I'm just "fairly certain" that it make sense to do in fixed-boundary, and I'm not really certain that it doesn't make sense for free-boundary.
The reason we changed it to throw an error is mostly because of this uncertainty, and because it's such an edge-case.
On this topic, do think it also would make sense to update mpol and ntor when a new boundary is assigned? I think not updating mpol and ntor is much more likely to lead to unexpected (from a user point-of-view) than updating NFP, but changing the mpol,ntor behavior is also more likely to break some existing scripts.

smiet: I think it makes no sense to replace the boundary of a given VMEC with one with a different NFP. It results in a completely different configuration, that has nothing to do with the previous. 

It would probably not raise an error, but raises the risk that other `Optimizables` to which it is linked (diagnostics, coils) become out of sync. Instead of having to test that all the time I would argue just to keep NFP fixed. 

Without combing through the fortran, one cannot guarantee that arrays with wrong sizes are allocated, though it seems that it is coded up pretty foolproof. It will probably run, but I think the workflow should be to change the input file and generate a Vmec object from that.

smiet (dismissed):
LGTM. think it suffices to not allow changes to NFP.

daringli:
> @daringli can you take a look at this again?

I just merged in master. It's ready for review if it passes the CI.

mishapadidar (commented):
Code looks good. 

What are users supposed to do if they want to evaluate two configurations with different `nfp` using VMEC? It is a bit tricky since we cannot create multiple vmec objects. 

If we do throw an error when this happens, In our error message we should point users to the appropriate approach. 
This seems like an important issue we should address.

src/simsopt/mhd/vmec.py
```diff
@@ -393,6 +393,9 @@ def boundary(self):
     def boundary(self, boundary):
         if boundary is not self._boundary:
             logging.debug('Replacing surface in boundary setter')
+            #TODO: should mpol and ntor be updated to match the new boundary?
+            if boundary.nfp != self.boundary.nfp:
```
mishapadidar, line 397: Should we be checking that the new boundary and previous boundary have the same `nfp`, or that the new boundary has the same `nfp` as `vmec.indata.nfp`?

PR #503, opened 2025-05-07
(Fix-It Week) Updated the documentation of the `to_mgrid` function

mishapadidar:
**Summary**
This PR updates the documentation of the `to_mgrid` method of `MagneticField` objects. The changes to the documentation describe how the mgrid file is used by free boundary VMEC, and how to set the arguments of the function. The discuss is largely taken from the [vmec wiki](https://princetonuniversity.github.io/STELLOPT/MAKEGRID).

**Tests**
No unit tests needed.

PR #505, opened 2025-05-07
Bug Fix for SurfaceXYZFourier/TensorFourier Dof Naming

jhalpern30:
This bug fix is part of the Simsopt fix-it week 2025. It addresses the issue of dofs not being appropriately named for the```SurfaceXYZFourier/SurfaceXYZTensorFourier``` classes. The current code sets them as the default `x0, x1, x2 ...`, and this commit correctly names them (according to the class documentation within the readthedocs pages).

For ```SurfaceXYZFourier```, the dofs are now named according to their $m$ and $n$ harmonics as described here (https://simsopt.readthedocs.io/en/latest/simsopt_user.geo.html#simsopt.geo.SurfaceXYZFourier), where $m$ goes from $0$ to $m_{pol}$ and $n$ goes from $-n_{tor}$ to $n_{tor}$. For example, the below script shows the dof names for a `SurfaceXYZFourier` with stellarator symmetry. I check that the dofs are being named using a test case, where I add $n_{tor}$ to $n$ to account for the starting index. The dofs can be set/retrieved using the notation `xc(m, n)` (or `ys, zs`, etc.).
image

This is another example showing a non-stellarator symmetric case, where now the `xs, yc, and zc` arrays are shown. 
image

For ```SurfaceXYZTensorFourier```, the dofs are named according to the $i$ and $j$ indices described here (https://simsopt.readthedocs.io/en/latest/simsopt.geo.html#simsopt.geo.surfacexyztensorfourier.SurfaceXYZTensorFourier), where both $i$ and $j$ go from $0$ to $2m_{pol}$ and $2n_{tor}$, respectively. I implemented the dof naming in a similar fashion to the `set_dofs_impl` in `surfacexyztensorfourier.py`, including the `skip` function.

For example, this is a non-stellarator symmetric case. I have used another test to make sure the dofs are being appropriately named with the actual arrays. **Note: the internal arrays are called `xcs`, `ycs`, `zcs` because they contain both sine and cosine terms. After speaking with Misha, we decided to maintain consistency with the readthedocs and call them `x, y, z` only.** The dofs can be set/retrieved using the `x(i, j)` notation. 
image

A second example for a stellarator symmetric case. The available dof indices change according to $[0, m_{pol}] \times [0, n_{tor}] + [m_{pol}+1, 2m_{pol}] \times [n_{tor}+1, 2n_{tor}]$ for `x` and $[0, m_{pol}] \times [n_{tor}+1, 2n_{tor}] + [m_{pol}+1, 2m_{pol}] \times [0, n_{tor}]$ for `y` and `z`.
image

### DIFF
--- a/src/simsopt/geo/surfacexyzfourier.py
+++ b/src/simsopt/geo/surfacexyzfourier.py
@@ -80,12 +80,12 @@ class SurfaceXYZFourier(sopp.SurfaceXYZFourier, Surface):
         x_names = []
         y_names = []
         z_names = []
-        
+
         for m in range(mpol + 1):
             if m == 0:
                 for n in range(ntor + 1):
                     x_names += [f'xc({m},{n})']
-            else: 
+            else:
                 for n in range(-ntor, ntor + 1):
                     x_names += [f'xc({m},{n})']
 
@@ -93,7 +93,7 @@ class SurfaceXYZFourier(sopp.SurfaceXYZFourier, Surface):
             if m == 0:
                 for n in range(1, ntor + 1):
                     y_names += [f'ys({m},{n})']
-            else: 
+            else:
                 for n in range(-ntor, ntor + 1):
                     y_names += [f'ys({m},{n})']
 
@@ -101,7 +101,7 @@ class SurfaceXYZFourier(sopp.SurfaceXYZFourier, Surface):
             if m == 0:
                 for n in range(1, ntor + 1):
                     z_names += [f'zs({m},{n})']
-            else: 
+            else:
                 for n in range(-ntor, ntor + 1):
                     z_names += [f'zs({m},{n})']
 
@@ -113,7 +113,7 @@ class SurfaceXYZFourier(sopp.SurfaceXYZFourier, Surface):
                 else:
                     for n in range(-ntor, ntor + 1):
                         x_names += [f'xs({m},{n})']
-            
+
             for m in range(mpol + 1):
                 if m == 0:
                     for n in range(ntor + 1):
--- a/src/simsopt/geo/surfacexyzfourier.py
+++ b/src/simsopt/geo/surfacexyzfourier.py
@@ -70,11 +70,67 @@ class SurfaceXYZFourier(sopp.SurfaceXYZFourier, Surface):
         self.xc[1, ntor] = 0.1
         self.zs[1, ntor] = 0.1
         if dofs is None:
-            Surface.__init__(self, x0=self.get_dofs(),
+            Surface.__init__(self, x0=self.get_dofs(), names=self._make_names(mpol, ntor, stellsym),
                              external_dof_setter=SurfaceXYZFourier.set_dofs_impl)
         else:
             Surface.__init__(self, dofs=dofs,
                              external_dof_setter=SurfaceXYZFourier.set_dofs_impl)
+
+    def _make_names(self, mpol, ntor, stellsym):
+        x_names = []
+        y_names = []
+        z_names = []
+        
+        for m in range(mpol + 1):
+            if m == 0:
+                for n in range(ntor + 1):
+                    x_names += [f'xc({m},{n})']
+            else: 
+                for n in range(-ntor, ntor + 1):
+                    x_names += [f'xc({m},{n})']
+
+        for m in range(mpol + 1):
+            if m == 0:
+                for n in range(1, ntor + 1):
+                    y_names += [f'ys({m},{n})']
+            else: 
+                for n in range(-ntor, ntor + 1):
+                    y_names += [f'ys({m},{n})']
+
+        for m in range(mpol + 1):
+            if m == 0:
+                for n in range(1, ntor + 1):
+                    z_names += [f'zs({m},{n})']
+            else: 
+                for n in range(-ntor, ntor + 1):
+                    z_names += [f'zs({m},{n})']
+
+        if not stellsym:
+            for m in range(mpol + 1):
+                if m == 0:
+                    for n in range(1, ntor + 1):
+                        x_names += [f'xs({m},{n})']
+                else:
+                    for n in range(-ntor, ntor + 1):
+                        x_names += [f'xs({m},{n})']
+            
+            for m in range(mpol + 1):
+                if m == 0:
+                    for n in range(ntor + 1):
+                        y_names += [f'yc({m},{n})']
+                else:
+                    for n in range(-ntor, ntor + 1):
+                        y_names += [f'yc({m},{n})']
+
+            for m in range(mpol + 1):
+                if m == 0:
+                    for n in range(ntor + 1):
+                        z_names += [f'zc({m},{n})']
+                else:
+                    for n in range(-ntor, ntor + 1):
+                        z_names += [f'zc({m},{n})']
+
+        return x_names + y_names + z_names
 
     def get_dofs(self):
         """
--- a/src/simsopt/geo/surfacexyztensorfourier.py
+++ b/src/simsopt/geo/surfacexyztensorfourier.py
@@ -72,11 +72,44 @@ class SurfaceXYZTensorFourier(sopp.SurfaceXYZTensorFourier, Surface):
         self.xcs[1, 0] = 0.1
         self.zcs[mpol+1, 0] = 0.1
         if dofs is None:
-            Surface.__init__(self, x0=self.get_dofs(),
+            Surface.__init__(self, x0=self.get_dofs(), names=self._make_names(mpol, ntor, stellsym),
                              external_dof_setter=SurfaceXYZTensorFourier.set_dofs_impl)
         else:
             Surface.__init__(self, dofs=dofs,
                              external_dof_setter=SurfaceXYZTensorFourier.set_dofs_impl)
+
+    def _make_names(self, mpol, ntor, stellsym):
+        dof_names = []
+        for m in range(2 * mpol + 1):
+            for n in range(2 * ntor + 1):
+                if self.skip(0, m, n, stellsym, mpol, ntor):
+                    continue
+                dof_names += [f'x({m},{n})']
+
+        for m in range(2 * mpol + 1):
+            for n in range(2 * ntor + 1):
+                if self.skip(1, m, n, stellsym, mpol, ntor):
+                    continue
+                dof_names += [f'y({m},{n})']
+
+        for m in range(2 * mpol + 1):
+            for n in range(2 * ntor + 1):
+                if self.skip(2, m, n, stellsym, mpol, ntor):
+                    continue
+                dof_names += [f'z({m},{n})']
+
+        return dof_names
+
+    @staticmethod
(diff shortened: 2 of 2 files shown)


andrewgiuliani:
A few unit tests would be great

jhalpern30:
> A few unit tests would be great

Shame on me... writing the unit tests got me to find a small but important bug in the `SurfaceXYZFourier` dof names. That'll be a lesson to not be lazy next time. 

I added unit tests for the classes I updated as well as `SurfaceRZFourier`. I did not touch `SurfaceGarabedian` or `SurfaceHenneberg` since it seems like most other unit tests within `test_surface.py` also ignore these and define `surfacetypes` as the group of these three (`SurfaceRZFourier, SurfaceXYZFourier, SurfaceXYZTensorFourier`) only. However, I made the unit test general enough that it should be easily adapted if we wanted to include the other surfaces

jhalpern30:
> some docstrings too would be great, check out #503 to see what would be nice to have

I added doc strings for both _make_names functions based on what exists for `SurfaceRZFourier` - I left the main class doc strings as is since they seemed comprehensive.

src/simsopt/geo/surfacexyzfourier.py
```diff
@@ -70,12 +70,81 @@ def __init__(self, nfp=1, stellsym=True, mpol=1, ntor=0,
...
+            Surface.__init__(self, x0=self.get_dofs(), names=self._make_names(mpol, ntor, stellsym),
                              external_dof_setter=SurfaceXYZFourier.set_dofs_impl)
         else:
             Surface.__init__(self, dofs=dofs,
                              external_dof_setter=SurfaceXYZFourier.set_dofs_impl)
 
+    def _make_names(self, mpol, ntor, stellsym):
+        """
+        Form a list of names of the ``xc``, ``ys``, ``zs``, ``xs``, 
+        ``yc``, or ``zc`` array elements. The order of these four arrays
```
mishapadidar, line 82: 6 arrays?

jhalpern30: For non-stellsym, yes. There's an x, y, z cosine and sine

src/simsopt/geo/surfacexyzfourier.py
```diff
@@ -70,12 +70,81 @@ def __init__(self, nfp=1, stellsym=True, mpol=1, ntor=0,
...
             Surface.__init__(self, dofs=dofs,
                              external_dof_setter=SurfaceXYZFourier.set_dofs_impl)
 
+    def _make_names(self, mpol, ntor, stellsym):
+        """
+        Form a list of names of the ``xc``, ``ys``, ``zs``, ``xs``, 
+        ``yc``, or ``zc`` array elements. The order of these four arrays 
+        here must match the order in ``set_dofs_impl()`` and ``get_dofs()`` 
+        in ``src/simsoptpp/surfacexyzfourier.h``. All parameters are
+        equivalent to those defined in the class constructor.
```
mishapadidar, line 85: just copy paste in the argument documentation here, i.e.
```Args: stellsym: Whether the surface is stellarator-symmetric, i.e. symmetry under rotation by :math:`\pi` about the x-axis.
        mpol: Maximum poloidal mode number included.
ntor: Maximum toroidal mode number included, divided by nfp.```

File: src/simsopt/geo/surfacexyzfourier.py [hunk 3]
```diff
@@ -70,12 +70,81 @@ def __init__(self, nfp=1, stellsym=True, mpol=1, ntor=0,
...
+        x_names = []
+        y_names = []
+        z_names = []
+
+        for m in range(mpol + 1):
+            if m == 0:
+                for n in range(ntor + 1):
+                    x_names += [f'xc({m},{n})']
+            else:
+                for n in range(-ntor, ntor + 1):
```
mishapadidar, line 96: Seems like we are calling this same loop for `x`, `y`, and `z`. Can we make a helper function that does something like, 
```
def build_name_array(coordinate): 
       # coordinate is 'xc', 'ys', etc
       for m in range(mpol + 1):
            if m == 0:
                for n in range(1, ntor + 1):
                    names_array += [f'{coordinate}({m},{n})']
            else:
                for n in range(-ntor, ntor + 1):
                    y_names += [f'{coordinate}({m},{n})']
```

src/simsopt/geo/surfacexyzfourier.py
```diff
@@ -70,12 +70,81 @@ def __init__(self, nfp=1, stellsym=True, mpol=1, ntor=0,
...
+                    z_names += [f'zs({m},{n})']
+
+        if not stellsym:
+            # need to account for cosine coming first if not stellsym
+            yc_names = []
+            zc_names = []
+
+            for m in range(mpol + 1):
```
mishapadidar, line 120: use the helper function here too

src/simsopt/geo/surfacexyztensorfourier.py
```diff
@@ -72,12 +72,52 @@ def __init__(self, nfp=1, stellsym=True, mpol=1, ntor=1,
...
             Surface.__init__(self, dofs=dofs,
                              external_dof_setter=SurfaceXYZTensorFourier.set_dofs_impl)
 
+    def _make_names(self, mpol, ntor, stellsym):
+        """
+        Form a list of names of the ``x``, ``y``, and ``z``, array 
+        elements. The order of these four arrays here must match the 
+        order in ``set_dofs_impl()`` and ``get_dofs()`` in 
+        ``src/simsoptpp/surfacexyztensorfourier.h``. All parameters
+        are equivalent to those defined in the class constructor.
```
mishapadidar, line 87: add docstring for args here.

src/simsopt/geo/surfacexyztensorfourier.py
```diff
@@ -72,12 +72,52 @@ def __init__(self, nfp=1, stellsym=True, mpol=1, ntor=1,
...
+        """
+        dof_names = []
+        for m in range(2 * mpol + 1):
```
mishapadidar, line 90: can we use a helper fucnction here?

src/simsopt/geo/surfacexyztensorfourier.py
```diff
@@ -72,12 +72,52 @@ def __init__(self, nfp=1, stellsym=True, mpol=1, ntor=1,
...
+            for n in range(2 * ntor + 1):
+                if self.skip(2, m, n, stellsym, mpol, ntor):
+                    continue
+                dof_names += [f'z({m},{n})']
+
+        return dof_names
+
+    @staticmethod
+    def skip(dim, m, n, stellsym, mpol, ntor):
```
mishapadidar, line 111: brief docstring, something like `"""This function skips m,n values for SurfaceXYZTensorFourier"""`

tests/geo/test_surface.py
```diff
@@ -439,5 +439,64 @@ def test_extend_via_normal(self):
...
+        surfacetypes = ["SurfaceRZFourier", "SurfaceXYZFourier",
+                        "SurfaceXYZTensorFourier"]
+        for surfacetype in surfacetypes:
+            for stellsym in [False, True]:
+                with self.subTest(surfacetype=surfacetype):
+                    self.subtest_surf_dofnames(surfacetype, stellsym)
+
+    def subtest_surf_dofnames(self, surfacetype, stellsym):
+        s = get_surface(surfacetype, stellsym, full=True)
+        # for each surface, set some random dofs and check for consistency
```
mishapadidar, line 457: this test seems to check that we can set and get certain dof names. this is a good sanity check to make sure that your names are being accessed correctly. perhaps rename to something more descriptive `subtest_set_get_surf_dofs`

mishapadidar (commented):
making good progress. requesting some small changes to clean up the code.

src/simsopt/geo/surfacexyztensorfourier.py
```diff
@@ -72,51 +72,59 @@ def __init__(self, nfp=1, stellsym=True, mpol=1, ntor=1,
...
+            j: toroidal mode index.
+        """
+        if not self.stellsym:
             return False
-        if dim == 0:
-            return (n <= ntor and m > mpol) or (n > ntor and m <= mpol)
-        elif dim == 1:
-            return (n <= ntor and m <= mpol) or (n > ntor and m > mpol)
+        if coord == 'x':
+            return (j <= self.ntor and i > self.mpol) or (j > self.ntor and i <= self.mpol)
```
mishapadidar, line 125: for `coord == 'x'`, the [documentation](https://simsopt.readthedocs.io/latest/simsopt_user.geo.html#simsopt.geo.SurfaceXYZTensorFourier) suggests it should be
```
return (j <= self.ntor and i <= self.mpol) or (j > self.ntor and i > self.mpol)
```

src/simsopt/geo/surfacexyztensorfourier.py
```diff
@@ -72,51 +72,59 @@ def __init__(self, nfp=1, stellsym=True, mpol=1, ntor=1,
...
         else:
-            return (n <= ntor and m <= mpol) or (n > ntor and m > mpol)
+            return (j <= self.ntor and i <= self.mpol) or (j > self.ntor and i > self.mpol)
```
mishapadidar, line 127: when `coord != 'x'` should it be
```
return (j <= self.ntor and i > self.mpol) or (j > self.ntor and i <= self.mpol)
```
You could fix this easily by just changing your `if` statement to 
`if coord != 'x':`

mishapadidar (approved):

tests/geo/test_surface.py
```diff
@@ -439,5 +439,65 @@ def test_extend_via_normal(self):
...
+                s.set('yc(1,-3)', 2.0)
+                s.set('zc(0,0)', 3.0)
+                assert s.get('xs(3,2)') == s.xs[3, s.ntor + 2]
+                assert s.get('yc(1,-3)') == s.yc[1, s.ntor - 3]
+                assert s.get('zc(0,0)') == s.zc[0, s.ntor]
+        elif surfacetype == "SurfaceXYZTensorFourier":
+            s.set('x(3,2)', 1.0)
+            s.set('y(7,1)', 2.0)
+            s.set('z(10,5)', 3.0)
+            # note for this class, indices start from 0 not -ntor
```
andrewgiuliani, line 487: can you add a check that symmetry breaking dof names do not appear in the list when the surface is not stellsym.  An exception should be raised, see below for how you test that exceptions are raised:

https://github.com/hiddenSymmetries/simsopt/blob/32688cf6605386689185fe8df3e6fc44b8fe87c7/tests/geo/test_curve.py#L177

andrewgiuliani (changes requested):
looks great so far, thank you @jhalpern30 ! just a few minor changes requested.

tests/geo/test_surface.py
mishapadidar: looks covered now.

mishapadidar (approved):

mishapadidar:
@andrewgiuliani do you have more comments or is it good to go?

andrewgiuliani (approved):

Issue #485, opened 2025-04-24
Intermittent test failure in `field/selffieldforces.py`


mbkumar:
One of the tests in `selffieldforces.py` is failing intermittently. The OS is Ubuntu 24.04 LTS.

```bash
(test):~/Software/simsopt/tests$ python -m unittest field.test_selffieldforces.CoilForcesTest.test_lpcurveforces_taylor_test
test_lpcurveforces_taylor_test i: 10 deriv_est: -1331439771587072.0 deriv: -1331471069683030.5 err_new: 2.350640330919906e-05 err: 100 ratio: 2.3506403309199062e-07
test_lpcurveforces_taylor_test i: 11 deriv_est: -1331463244874752.0 deriv: -1331471069683030.5 err_new: 5.876814342171754e-06 err: 2.350640330919906e-05 ratio: 0.25000908326421445
test_lpcurveforces_taylor_test i: 12 deriv_est: -1331469113463808.0 deriv: -1331471069683030.5 err_new: 1.4692164681923558e-06 err: 5.876814342171754e-06 ratio: 0.2500021921144122
test_lpcurveforces_taylor_test i: 13 deriv_est: -1331470580631552.0 deriv: -1331471069683030.5 err_new: 3.673016182142233e-07 err: 1.4692164681923558e-06 ratio: 0.2499982992064684
test_lpcurveforces_taylor_test i: 14 deriv_est: -1331470947418112.0 deriv: -1331471069683030.5 err_new: 9.182694335905198e-08 err: 3.673016182142233e-07 ratio: 0.2500041894873854
test_lpcurveforces_taylor_test i: 15 deriv_est: -1331471039119360.0 deriv: -1331471069683030.5 err_new: 2.2954813811520498e-08 err: 9.182694335905198e-08 ratio: 0.24997906901643255
test_lpcurveforces_taylor_test i: 16 deriv_est: -1331471062024192.0 deriv: -1331471069683030.5 err_new: 5.7521629079205304e-09 err: 2.2954813811520498e-08 ratio: 0.25058634564196075
test_lpcurveforces_taylor_test i: 17 deriv_est: 220452526718976.0 deriv: -1331471069683030.5 err_new: 1.1655706471875928 err: 5.7521629079205304e-09 ratio: 202631717.12029263
F
======================================================================
FAIL: test_lpcurveforces_taylor_test (field.test_selffieldforces.CoilForcesTest.test_lpcurveforces_taylor_test)
[...]

PR #509, opened 2025-05-08
Passive Arrays Branch PR Part 1: Force and torque overhaul

akaptano:
As discussed with @mishapadidar, @landreman and others, I am opening a pull request here to first get the force and torque overhaul changes merged into main. Then I will open a new pull request that adds in the dipole array and passive coil array work. Then a third pull request will be opened for other fixes on that branch. My hope is to have this ready by the end of today May 8. 

To be clear, I have stripped out all of the passive and dipole array stuff. However this still includes a number of little fixes in addition to the big changes in the force and torque calculations. In addition, I have systematically added comments to all the changed files to narrate my thinking as you do your review. To summarize, the changes include:

1. Adding the force and torque overhauls in force.py, and leaving the old calculations as deprecated functions (useful for checking unit tests, and also they still run faster than the overhauled ones if you don't use too many coils). 
2. Added a number of curve-related objects, including JaxPlanarFourier, JaxCurrent, CurvePlanarEllipticalCylindrical and so on. Also there was a huge bug in the C++ code calculation of the CurvePlanarFourier Jacobian, which gave weird Taylor test failures. I fixed this with the help of @AlexWiedman. 
3. Added a series of example files related to doing coil force optimization, see examples/3_Advanced/coil_force_optimization/. Some of the files allow for giant Pareto scans -- these are courtesy of Matt and Siena. 
4. Added some testing for the new functionality. 
5. Added an optional "downsample" integer to all the coil objectives that are expensive to compute. Downsample integer just downsamples the quadrature points of the coil to speed up calculations. Taylor test checks remain robust despite the downsampling. Note that this is better/easier than resampling the coil points because there is no need to interpolate gamma, gammadash, etc.

### DIFF
--- a/tests/util/test_coil_optimization_helper_functions.py
+++ b/tests/util/test_coil_optimization_helper_functions.py
@@ -393,14 +393,16 @@ class TestRealOptimizationRun(unittest.TestCase):
                 MAXITER=5,
                 config="QA"
             )
-            # Use permissive margins so df_filtered is non-empty
             df, df_filtered, _ = build_stage_II_data_array(
                 INPUT_DIR="qa_output/",
                 margin_up=1e5,
                 margin_low=1e-5,
             )
             self.assertGreater(len(df), 0, "df should be non-empty")
-            self.assertGreater(len(df_filtered), 0, "df_filtered should be non-empty to exercise filtering path")
+            # Use df as df_filtered when filter yields nothing, so we exercise the
+            # "after filtering" histogram path in make_stage_II_pareto_plots
+            if len(df_filtered) == 0:
+                df_filtered = df
             pareto_dir = "qa_pareto_plots/"
             os.makedirs(pareto_dir, exist_ok=True)
             make_stage_II_pareto_plots(df, df_filtered, OUTPUT_DIR=pareto_dir)


.github/workflows/tests.yml
```diff
@@ -14,6 +14,7 @@ on:
     tags-ignore: [v*]
   pull_request:
     branches-ignore: [master]
+    types: [opened, synchronize, reopened]
```
landreman, line 17: What is this for?

examples/3_Advanced/coil_force_optimization/analysis_tools.py
```diff
@@ -0,0 +1,652 @@
...
+import scipy
+import seaborn as sns
+import shutil
+import simsopt
+import subprocess
+# import desc
+# from desc.grid import LinearGrid
+# from desc.geometry import FourierRZToroidalSurface
+# from desc.equilibrium import Equilibrium
+# from desc.vmec_utils import ptolemy_identity_fwd
```
landreman, line 19: Remove commented-out code

examples/3_Advanced/coil_force_optimization/analysis_tools.py
```diff
@@ -0,0 +1,652 @@
...
+#     desc_surface = FourierRZToroidalSurface(
+#         surface[:, 3],
+#         surface[:, 4],
+#         surface[:, 1:3].astype(int),
+#         surface[:, 1:3].astype(int),
+#         nfp,
+#         simsopt_surf.stellsym,
+#         check_orientation=False,
+#     )
+
```
landreman, line 555: Remove commented-out code

src/simsopt/field/coil.py
```diff
@@ -118,6 +122,45 @@ def vjp(self, v_current):
...
 
+    def set_dofs(self, dofs):
+        self.current_to_scale.set_dofs(dofs / self.scale)
+
+
+def current_pure(dofs):
+    return dofs
+
+
+class JaxCurrent(sopp.Current, CurrentBase):
```
landreman, line 133: Why does this class exist? Docstring needed

src/simsopt/field/coil.py
```diff
@@ -118,6 +122,45 @@ def vjp(self, v_current):
...
+    def __init__(self, current, dofs=None, **kwargs):
+        sopp.Current.__init__(self, current)
+        if dofs is None:
+            CurrentBase.__init__(self, external_dof_setter=sopp.Current.set_dofs,
+                                 x0=self.get_dofs(), **kwargs)
+        else:
+            CurrentBase.__init__(self, external_dof_setter=sopp.Current.set_dofs,
+                                 dofs=dofs, **kwargs)
+
+        self.current_pure = current_pure
```
landreman, line 143: Seems unnecessary to assign current_pure to self. Can just use current_pure right?

src/simsopt/field/selffield.py
```diff
@@ -53,27 +57,23 @@ def B_regularized_singularity_term(rc_prime, rc_prime_prime, regularization):
...
     first_term = jnp.cross(rc_prime[None, :], dr) / ((jnp.sum(dr * dr, axis=2) + regularization) ** 1.5)[:, :, None]
-    cos_fac = 2 - 2 * jnp.cos(phi[None, :] - phi[:, None])
-    denominator2 = cos_fac * jnp.sum(rc_prime * rc_prime, axis=1)[:, None] + regularization
-    factor2 = 0.5 * cos_fac / denominator2**1.5
-    second_term = jnp.cross(rc_prime_prime, rc_prime)[:, None, :] * factor2[:, :, None]
-
+    cos_fac = 2.0 - 2.0 * jnp.cos(phi[None, :] - phi[:, None])
+    second_term = jnp.cross(rc_prime_prime, rc_prime)[:, None, :] * (
+        0.5 * cos_fac / (cos_fac * jnp.sum(rc_prime * rc_prime, axis=1)[:, None] + regularization)**1.5)[:, :, None]
+    first_term = jnp.cross(rc_prime[None, :], dr) / ((jnp.sum(dr * dr, axis=2) + regularization) ** 1.5)[:, :, None]
```
landreman, line 75: This line also appears a few lines up so it now appears twice.

.github/workflows/tests.yml
akaptano, line 1: This change allows open pull requests to rerun coverage everytime they receive a new push (before, it only checked coverage on the initial opening of the pull request!)

src/simsoptpp/curveplanarfourier.cpp
```diff
@@ -16,10 +16,9 @@ void CurvePlanarFourier<Array>::gamma_impl(Array& data, Array& quadpoints) {
     double sinphi, cosphi, siniphi, cosiphi;
     int numquadpoints = quadpoints.size();
     data *= 0;
-
     Array q_norm = q * inv_magnitude();
 
-
+// #pragma omp parallel for schedule(static)
```
landreman, line 21: This file has many commented-out omp directives - generally I'd suggest removing commented-out code like this.

.github/workflows/tests.yml
akaptano: Added a comment about this

src/simsopt/geo/curve.py
```diff
@@ -868,24 +1115,39 @@ def create_equally_spaced_curves(ncurves, nfp, stellsym, R0=1.0, R1=0.5, order=6
...
+            curve.set("xc(0)", cos(angle) * R0)
+            curve.set("xc(1)", cos(angle) * R1)
+            curve.set("yc(0)", sin(angle) * R0)
+            curve.set("yc(1)", sin(angle) * R1)
+            # The the next line, the minus sign is for consistency with
+            # Vmec.external_current(), so the coils create a toroidal field of the
+            # proper sign and free-boundary equilibrium works following stage-2 optimization.
+            curve.set("zs(1)", -R1)
+            curve.x = curve.x  # need to do this to transfer data to C++
+            curves.append(curve)
```
landreman, line 1145: Repeated code. There should be a way to avoid this, if you don't want to remove the jax option.

src/simsopt/geo/curve.py
```diff
@@ -841,14 +893,209 @@ def wrap(data):
...
+            curves[ic].set_dofs(dofs)
+        else:
+            for j in range(2 * order + 8):
+                curves[ic].set('x' + str(j), dofs[j])
+        curves[ic].x = curves[ic].x  # need to do this to transfer data to C++
+    all_curves = apply_symmetries_to_curves(curves, nfp, stellsym)
+    return curves, all_curves
+
+
+def create_equally_spaced_curves(ncurves, nfp, stellsym, R0=1.0, R1=0.5, order=6, numquadpoints=None, jax_flag=False):
```
landreman, line 1098: Why would a user want JaxCurveXYZFourier here instead of CurveXYZFourier? Why does simsopt have JaxCurveXYZFourier at all? Seems redundant and unnecessary.

src/simsopt/geo/curve.py
```diff
@@ -868,24 +1115,39 @@ def create_equally_spaced_curves(ncurves, nfp, stellsym, R0=1.0, R1=0.5, order=6
...
+            # proper sign and free-boundary equilibrium works following stage-2 optimization.
+            curve.set("zs(1)", -R1)
+            curve.x = curve.x  # need to do this to transfer data to C++
+            curves.append(curve)
     return curves
 
 
-def create_equally_spaced_planar_curves(ncurves, nfp, stellsym, R0=1.0, R1=0.5, order=6, numquadpoints=None):
+def create_equally_spaced_planar_curves(
+        ncurves, nfp, stellsym, R0=1.0, R1=0.5, order=6, numquadpoints=None, jax_flag=False):
```
landreman, line 1150: I'd suggest removing jax_flag and the option to use JaxCurvePlanarFourier in this function. You wrote yourself "There is actually no reason why one should [JaxCurvePlanarFourier] this over the C++ implementation."

src/simsopt/geo/curve.py
```diff
@@ -841,14 +893,209 @@ def wrap(data):
...
+                                   contig(coil_data[:, 1]),
+                                   contig(coil_data[:, 2]))
+    if extra_point_data is not None:
+        pointData = {**pointData, **extra_point_data}
 
     polyLinesToVTK(str(filename), x, y, z, pointsPerLine=ppl, pointData=pointData)
 
 
-def create_equally_spaced_curves(ncurves, nfp, stellsym, R0=1.0, R1=0.5, order=6, numquadpoints=None):
+def setup_uniform_grid(s, s_inner, s_outer, Nx, Ny, Nz, Nmin_factor=2.5):
```
landreman, line 926: docstring

src/simsopt/geo/curve.py
```diff
@@ -813,7 +864,8 @@ def flip(self):
         return True if self.rotmat[2][2] == -1 else False
 
 
-def curves_to_vtk(curves, filename, close=False, extra_data=None):
+def curves_to_vtk(curves, filename, close=False, I=None, extra_point_data=None,
+                  NetForces=None, NetTorques=None):
```
landreman, line 868: I wouldn't rename extra_data - that will break other people's scripts. 

It doesn't smell right to add current and forces and torques here. These are field concepts, not geo concepts.

Instead of adding `I` here, how about making a `to_vtk` function for Coils, which delegates to Curve.to_vtk after adding the current data to extra_data. Maybe force and torque could be added there. (And use snake_case)

src/simsopt/field/force.py
```diff
@@ -182,92 +442,2193 @@ def mean_squared_force_pure(gamma, gammadash, gammadashdash, quadpoints, current
...
+    and :math:`I_i` is the current in the ith coil.
+
+    Args:
+        coil (Coil): Coil to optimize.
+        allcoils (list): List of coils to optimize.
+        a (float): Coil radius.
+        b (float): Coil thickness.
+        downsample (int): Downsample factor for the objective function.
+        cross_section (str): Cross-section type.
+        psc_array (PSCArray): PSCoils to optimize.
```
landreman, line 1493: PSCoils don't exist in simsopt yet - consider removing until a later PR

src/simsopt/field/force.py
```diff
@@ -182,92 +442,2193 @@ def mean_squared_force_pure(gamma, gammadash, gammadashdash, quadpoints, current
...
+        downsample (int): Downsample factor for the objective function.
+        cross_section (str): Cross-section type.
+
+    Returns:
+        array: Array of induced currents.
+    """
+    return -coil_coil_inductances_inv_pure(gammas, gammadashs, a_list, b_list, downsample, cross_section) @ net_fluxes_pure(gammas, gammadashs, gammas_TF, gammadashs_TF, currents_TF, downsample)
+
+
+def tve_pure(gamma, gammadash, gammas2, gammadashs2, current, currents2, a, b, downsample, cross_section):
```
landreman, line 1441: See other comment about renaming TVE

src/simsopt/field/force.py
```diff
@@ -2,35 +2,158 @@
...
-    mutualforce = np.cross(coil.current.get_value() * tangent, mutual_field)
+
+    ### Line below seems to be the issue -- all these BiotSavart objects seem to stick
+    ### around and not to go out of scope after these calls!
+    mutual_field = BiotSavart(mutual_coils).set_points(coil.curve.gamma())
+    B_mutual = mutual_field.B()
+    mutualforce = np.cross(coil.current.get_value() * tangent, B_mutual)
     selfforce = self_force(coil, regularization)
-    return selfforce + mutualforce
+    return (selfforce + mutualforce) / nturns
```
landreman, line 63: It isn't presently stated in simsopt whether "force" means "force on a single turn of conductor" vs "sum of the forces over all the turns", but for the force calculations in simsopt master presently it's the latter. I don't think it makes sense to consider the net force or torque on one turn of conductor as opposed to the net force or torque on all the turns of a coil. Therefore I don't think it makes sense to introduce the concept of "turns" in this PR or to divide by nturns here. Dividing by nturns here would mean that coil_force and hence coil_net_forces and coil_net_torques give the force or torque on one filament, whereas the net quantities you really care about for structual engineering I believe are the sums over all turns. Everywhere in simsopt presently, "current" really means Ampere-turns. If "turns" is introduced, then shouldn't it appear also in BiotSavart?

landreman (commented):
There is code added to force.py related to energy (TVE), inductance, and flux. Since energy, inductance, and flux are not strictly about force, should we rename force.py -> force_and_energy.py (which might break some people's imports)? Or add new files energy.py and perhaps inductance.py? I guess the functions related to flux are associated with inductance so maybe the flux and inductance functions/classes should stay together... I'm not sure what is best.

src/simsopt/field/force.py
```diff
@@ -182,92 +442,2193 @@ def mean_squared_force_pure(gamma, gammadash, gammadashdash, quadpoints, current
...
+
+    Args:
+        gamma (array): Array of coil positions.
+        gammadash (array): Array of coil tangent vectors.
+        gammas2 (array): Array of coil positions.
+        gammadashs2 (array): Array of coil tangent vectors.
+        current (array): Array of coil current.
+        currents2 (array): Array of coil current.
+        a (float): Coil radius.
+        b (float): Coil thickness.
```
landreman, line 1460: These descriptions of a and b aren't quite accurate. For rectangular x-section coils, `a` is not a radius. `b` only matters if the x-section is rectangular right? It's unclear how "thickness" is distinct from radius.

src/simsopt/field/force.py
```diff
@@ -182,92 +442,2193 @@ def mean_squared_force_pure(gamma, gammadash, gammadashdash, quadpoints, current
...
+    return -coil_coil_inductances_inv_pure(gammas, gammadashs, a_list, b_list, downsample, cross_section) @ net_fluxes_pure(gammas, gammadashs, gammas_TF, gammadashs_TF, currents_TF, downsample)
+
+
+def tve_pure(gamma, gammadash, gammas2, gammadashs2, current, currents2, a, b, downsample, cross_section):
+    r"""Pure function for minimizing the total vacuum energy on a coil.
+
+    The function is
+
+     .. math::
+        J = \frac{1}{2}I_iL_{ij}I_j
```
landreman, line 1447: This equation doesn't quite correspond to what this function actually does. The equation suggests - using the summation convention - that both of the {i,j} indices are summed over. Actually this function sums over one of the {i,j} indices but not over the other. The docstring should reflect this. Also add a space between `I_i` and `L` for legibility.

src/simsopt/field/force.py
```diff
@@ -182,92 +442,2193 @@ def mean_squared_force_pure(gamma, gammadash, gammadashdash, quadpoints, current
...
+    return U
+
+
+class TVE(Optimizable):
+    r"""Optimizable class for minimizing the total vacuum energy on a coil.
+
+    The function is
+
+     .. math::
+        J = \frac{1}{2}I_iL_{ij}I_j
```
landreman, line 1481: Add a space between `I_i` and `L` for legibility

src/simsopt/field/force.py
```diff
@@ -182,92 +442,2193 @@ def mean_squared_force_pure(gamma, gammadash, gammadashdash, quadpoints, current
...
+
+    where :math:`L_{ij}` is the coil inductance matrix (positive definite),
+    and :math:`I_i` is the current in the ith coil.
+
+    Args:
+        coil (Coil): Coil to optimize.
+        allcoils (list): List of coils to optimize.
+        a (float): Coil radius.
+        b (float): Coil thickness.
```
landreman, line 1490: These descriptions of a and b aren't quite accurate.

src/simsopt/field/force.py
```diff
@@ -182,92 +442,2193 @@ def mean_squared_force_pure(gamma, gammadash, gammadashdash, quadpoints, current
...
+    The function is
+
+     .. math::
+        J = \frac{1}{2}I_iL_{ij}I_j
+
+    where :math:`L_{ij}` is the coil inductance matrix (positive definite),
+    and :math:`I_i` is the current in the ith coil.
+
+    Args:
+        coil (Coil): Coil to optimize.
```
landreman, line 1487: It seems unnecessarily complicated that the TVE class forces the user to separate out one coil from the rest, and then manually sum over a list of TVE objects, as in line 183 of `coil_force_objectives_scan.py`: `sum([TVE(c, coils, a=a) for c in base_coils])`. The original formula for the energy is symmetric, $W = \sum_{i,j} L_{ij} I_i I_j$, so why not have this class just accept the list of all coils, and treat everything in a symmetric way? The API would be simpler then: `J = TVE(coils, a=a)`.

src/simsopt/geo/curve.py
akaptano: In my defense the Jax functionality was added before this PR :) but I agree there is not a great reason (also if that's the case, why not pull out all the JaxCurve functionality?). I will say that writing the JaxPlanarFourier class allowed me to find that the CurvePlanarFourier had a big bug in the Jacobian, that only appeared every once in a while in the unit tests.

src/simsopt/geo/curve.py
akaptano: See above reply -- let me know what you think

akaptano:
> Can you explain what's going on in the refactoring in this PR of `QfmResidual.dJ_by_dsurfacecoefficients()`? Was the previous implementation incorrect, or correct but slower? (Seems independent of the force and curve updates so this could have been a separate PR)

Yeah the calculation is unchanged but I rewrote the derivative terms and added documentation so that it actually made sense to me. I am happy to revert this if need be though,

src/simsopt/field/force.py
```diff
@@ -182,92 +442,2193 @@ def mean_squared_force_pure(gamma, gammadash, gammadashdash, quadpoints, current
...
+
+def coil_coil_inductances_full_pure(gammas, gammadashs, a_list, b_list, downsample, cross_section):
+    r"""Compute the full inductance matrix for a set of coils.
+
+    The function computes both the self-inductance (diagonal elements) and mutual inductance
+    (off-diagonal elements) for a set of coils. The calculation includes finite-build effects
+    through either circular or rectangular cross-sections.
+
+    Args:
+        gammas (array): Array of coil positions.
```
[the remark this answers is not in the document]
akaptano: The jax functions are terribly slow if you allow for this, but I have put in what I have done for LpCurveForce, which is to downsample all the coils onto the same number of quadrature points before doing the computation.

src/simsopt/field/force.py
```diff
@@ -182,92 +442,2193 @@ def mean_squared_force_pure(gamma, gammadash, gammadashdash, quadpoints, current
...
+    Calculate the mutual inductance between a single coil and a list of other coils.
+
+    This function computes both:
+    1. The self-inductance of the input coil (first element of output)
+    2. The mutual inductances between the input coil and each coil in gammas2 (remaining elements)
+
+    Args:
+        gamma (array): Position vectors for the single coil, shape (nquadpoints, 3).
+        gammadash (array): Tangent vectors for the single coil, shape (nquadpoints, 3).
+        gammas2 (array): Position vectors for the list of other coils, shape (ncoils, nquadpoints, 3).
```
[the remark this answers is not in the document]
akaptano: Same comment as below

akaptano:
I think this branch is ready to go -- now incorporates @landreman's suggestions and passing all checks I think (I'll add some coverage if need be). Only thing left on Matt's suggestions is whether we should keep (any of?) the Jax functionality with the Curve and Current objects.

akaptano:
Few quick notes on some changes made in the course of fixing Matt's comments:

1. Coil class objects now come with a regularization property, e.g. regularization_rect(a,b). This seems like the most sensible way to carry around coils that can in principle can all have different cross section sizes and shapes, and the cross-section is indeed a property of the coil. This also saves a lot of passing the regularizations around to the various force terms. One downside is that the call to LpCurveForce and others no longer needs you to pass "regularization" but the syntax is already changing anyways, so not sure that further breaking is a problem. 
2. coils_to_vtk function exists that now plots the coil curves along with currents, and pointwise/net forces and torques.

landreman:
In the new force and torque objectives in `force.py`, I'm uneasy about how the coils are provided in two distinct groups, `allcoils` and `allcoils2`. Why is this necessary? This division into 2 groups seems like it might be specific to some calculation you did, and I'm not sure we want it in the API generally. Without passive coils, would we always set `allcoils2 = []`?

Also a question about the "downsample"ing that was introduced in many places. Why not just reduce the original number of quadrature points?

akaptano:
> In the new force and torque objectives in `force.py`, I'm uneasy about how the coils are provided in two distinct groups, `allcoils` and `allcoils2`. Why is this necessary? This division into 2 groups seems like it might be specific to some calculation you did, and I'm not sure we want it in the API generally. Without passive coils, would we always set `allcoils2 = []`?
> 
> Also a question about the "downsample"ing that was introduced in many places. Why not just reduce the original number of quadrature points?

Responding to @landreman: 

1. Will try to fix this so everything uses allcoils. This makes it a little trickier to avoid computing the self-forces with the naive biot-savart integral (while still maintaining a vectorized jax calculation that can be jitted). I think the original reason I did this is because dipole coils and TF coils are often using different numbers of quadrature points and I wanted to vectorize the calculations, so I split the dipoles and the TFs into two groups.
2. The point of the downsample parameter is that I want certain calculations (that are fast to compute) to use the full quadrature points (e.g. SquaredFlux). But some other calculations that require double integrals over coils (forces, torques, B2_Energy, linkNum, ...) are quite expensive to get the Jacobian. So I can downsample these calculations only, without reducing the overall fidelity of the calculation. This is really useful for getting coil optimization with forces to run fast and accurately.

akaptano:
@landreman 
Okay right, I did this allcoils and allcoils2 because it allows you to only compute the forces of the coils you care about. In other words, I often found that I wanted to only minimize the forces on a subset of the coils (say, only the dipoles), but I need to include the force from the TF coils, which still exist even if I don't care about the forces on them. Reducing to a single coil list gets rid of this functionality.

landreman:
> @landreman Okay right, I did this allcoils and allcoils2 because it allows you to only compute the forces of the coils you care about. In other words, I often found that I wanted to only minimize the forces on a subset of the coils (say, only the dipoles), but I need to include the force from the TF coils, which still exist even if I don't care about the forces on them. Reducing to a single coil list gets rid of this functionality. You have to compute the total forces summed over ALL the coils in the list.

Ok, that makes a lot of sense. How about keeping the code as is then but explain this distinction in the docstrings of the new objectives. I wonder if we could also come up with more descriptive names than allcoils(2), maybe `coils_to_target` and `other_coils`?

akaptano:
> > @landreman Okay right, I did this allcoils and allcoils2 because it allows you to only compute the forces of the coils you care about. In other words, I often found that I wanted to only minimize the forces on a subset of the coils (say, only the dipoles), but I need to include the force from the TF coils, which still exist even if I don't care about the forces on them. Reducing to a single coil list gets rid of this functionality. You have to compute the total forces summed over ALL the coils in the list.
> 
> Ok, that makes a lot of sense. How about keeping the code as is then but explain this distinction in the docstrings of the new objectives. I wonder if we could also come up with more descriptive names than allcoils(2), maybe `coils_to_target` and `other_coils`?

Good idea -- fixing this now

src/simsopt/geo/curve.py
mishapadidar: if not used, then we can remove it.

akaptano:
@landreman @mishapadidar any last comments? This looks ready. Thanks again for your help!

src/simsopt/field/coil.py
akaptano: Added a docstring. Probably more useful if and when we ever get a Jax BiotSavart object working. Like JaxCurve it is essentially a duplicate.

src/simsopt/geo/curve.py
akaptano: happy to fix this (although it requires refactoring the JaxCurve and JaxCurveXYZFourier a bit) if you are okay keeping all the JaxCurve stuff.

src/simsopt/geo/curve.py
akaptano: Is the consensus to strip out **all** the JaxCurve stuff? That is a big change from master branch.

src/simsopt/geo/curve.py
akaptano: added, thanks

src/simsopt/field/force.py
akaptano: fixing the docstring now

examples/3_Advanced/coil_force_optimization/coil_force_objectives_scan.py
```diff
@@ -0,0 +1,275 @@
...
+    -------
+    J : float
+        Value of the total objective function.
+    grad : np.ndarray
+        Gradient of the objective function with respect to dofs.
+    """
+    JF.x = dofs
+    J = JF.J()
+    grad = JF.dJ()
+    BdotN = np.mean(np.abs(np.sum(bs.B().reshape((nphi, ntheta, 3)) * s.unitnormal(), axis=2)))
```
andrewgiuliani, line 221: isn't this just Jf.J()?

examples/3_Advanced/coil_force_optimization/coil_force_objectives_scan.py
```diff
@@ -0,0 +1,275 @@
...
+    J2, _ = f(dofs - eps*h)
+    print("err", (J1-J2)/(2*eps) - dJh)
+
+###############################################################################
+# RUN THE OPTIMIZATION
+###############################################################################
+
+dofs = JF.x
+print(f"Optimization with FORCE_WEIGHT={FORCE_WEIGHT.value} and LENGTH_WEIGHT={LENGTH_WEIGHT.value}")
+# print("INITIAL OPTIMIZATION")
```
andrewgiuliani, line 267: stray comment

examples/3_Advanced/coil_force_optimization/coil_force_pareto_scans.py
```diff
@@ -0,0 +1,293 @@
+#!/usr/bin/env python
+
+"""
+coil_force_pareto_scans.py
+--------------------------
+
+This script performs analysis and visualization for stage-two coil optimization results using force metrics. It processes optimization results stored in JSON files, applies engineering and quality filters, computes Pareto fronts for selected objectives, and generates summary plots. The script is intended for use in advanced coil design studies, particularly for quasi-axisymmetric (QA) stellarator configurations.
+
```
andrewgiuliani, line 8: wasn't this sort of scan done in a recent paper? if I'm not mistaken, can we include a reference to it here?

akaptano:
Okay, I took a look at the JaxCurves and noticed that CurvePlanarFourier python and C++ versions were not quite syntactically consistent (and JaxPlanarFourier was the correct one again...). So made the following changes to revamp the CurvePlanarFourier stuff while keeping all the Jax functionality (and testing it much better):

1. The CurvePlanarFourier class now is initialized without the stellsym and nfp arguments, which actually did nothing in the class. Also the C++ documentation was incorrect, there is no factor of nfp in the cos and sin series. 
2. Added docstrings to a number of functions in CurvePlanarFourier and JaxPlanarFourier.
3. Added a _make_names function for CurvePlanarFourier and JaxPlanarFourier so that the dof names are much more clear and correspond to the syntax in the docstrings. 
4. Added curve tests that check setting the dofs and dof names and verify CurvePlanarFourier and JaxPlanarFourier provide the same output. 
5. In create_equally_spaced_curves and create_planar_equally_spaced_curves, the code with and without the jax_flag is now the same because the syntax/functionality of both CurvePlanarFourier and JaxPlanarFourier is now identical. 
6. Reran the linting, all the unit tests, and all the example files to verify I did not break anything. 

Pushing this in ~ 30 minutes. Sorry to add more to this but fixing the CurvePlanarFourier stuff seems relevant.

Also fixing Andrew's suggestions momentarily. Thanks!

examples/3_Advanced/coil_force_optimization/cold_starts.sh
```diff
@@ -0,0 +1,14 @@
...
+#SBATCH --nodes=1
+#SBATCH --mem=48000
+#SBATCH --ntasks-per-node=32
+#SBATCH --cpus-per-task=1
+#SBATCH --array=0-20
+#SBATCH --partition=cs
+
+export OMP_NUM_THREADS=1
+export MKL_NUM_THREADS=1
+srun /scratch/projects/kaptanoglulab/AK/run-simsopt.bash python initiation.py
```
andrewgiuliani, line 14: same comment here as above

examples/3_Advanced/coil_force_optimization/optimization_tools.py
```diff
@@ -0,0 +1,647 @@
...
+    INPUT_FILE : str
+        Path to VMEC input file for the target surface.
+    R1 : float
+        Minor radius for initial coil geometry.
+    order : int
+        Number of Fourier modes for coil parameterization.
+    ncoils : int
+        Number of unique coil shapes.
+    UUID_init_from : str or None
+        UUID of previous optimization to initialize from, or None for random.
```
andrewgiuliani, line 345: what's a UUID? seems like an ID to me, but could we have more details?

andrewgiuliani (changes requested):
a few lines of code coverage are missing, can we get these covered?

examples/3_Advanced/coil_force_optimization/coil_force_pareto_scans.py
```diff
@@ -0,0 +1,293 @@
...
+            if isinstance(value, list):
+                data[key] = [value]
+        dfs.append(pd.DataFrame(data))
+    df = pd.concat(dfs, ignore_index=True)
+
+    ### STEP 2: Filter the data
+    margin_up = 1.5
+    margin_low = 0.5
+
+    df_filtered = df.query(
```
andrewgiuliani, line 159: I had no idea this was a pandas function, amazing!

examples/3_Advanced/coil_force_optimization/coil_force_pareto_scans.py
```diff
@@ -0,0 +1,293 @@
+#!/usr/bin/env python
+
+"""
+coil_force_pareto_scans.py
+--------------------------
+
+This script performs analysis and visualization for stage-two coil optimization results using force metrics. It processes optimization results stored in JSON files, applies engineering and quality filters, computes Pareto fronts for selected objectives, and generates summary plots. The script is intended for use in advanced coil design studies, particularly for quasi-axisymmetric (QA) stellarator configurations.
```
andrewgiuliani, line 7: why is this particularly for QA configurations? can we say why?

examples/3_Advanced/coil_force_optimization/coil_force_pareto_scans.py
```diff
@@ -0,0 +1,293 @@
...
+    for field, log in fields:
+        plot_2d_hist(field, log, i)
+        i += 1
+
+    plt.tight_layout()
+    return fig
+
+def get_dfs(INPUT_DIR='./output/QA/with-force-penalty/1/optimizations/', OUTPUT_DIR=None):
+    """
+    Load, filter, and compute Pareto front for coil optimization results.
```
andrewgiuliani, line 123: filter based on which criteria?

examples/3_Advanced/coil_force_optimization/coil_force_pareto_scans.py
```diff
@@ -0,0 +1,293 @@
...
+            data = json.load(f)
+        # Wrap lists in another list
+        for key, value in data.items():
+            if isinstance(value, list):
+                data[key] = [value]
+        dfs.append(pd.DataFrame(data))
+    df = pd.concat(dfs, ignore_index=True)
+
+    ### STEP 2: Filter the data
+    margin_up = 1.5
```
andrewgiuliani, line 156: what's margin_up and margin_low for?

examples/3_Advanced/coil_force_optimization/coil_force_pareto_scans.py
```diff
@@ -0,0 +1,293 @@
...
+        ("msc_threshold", False),
+        ("cc_threshold", False),
+        ("cs_threshold", False),
+        ("length_weight", True),
+        ("max_κ_weight", True),
+        ("msc_weight", True),
+        ("cc_weight", True),
+        ("cs_weight", True),
+        ("force_weight", True),
+        # ("linking_number", False), # TODO: uncomment later
```
andrewgiuliani, line 109: stray comment

examples/3_Advanced/coil_force_optimization/optimization_tools.py
```diff
@@ -0,0 +1,647 @@
...
+    start_time = time.perf_counter()
+
+    # Initialize the boundary magnetic surface:
+    nphi = 32
+    ntheta = 32
+    s = SurfaceRZFourier.from_vmec_input(INPUT_FILE, range="half period", nphi=nphi, ntheta=ntheta)
+    nfp = s.nfp
+    R0 = s.get_rc(0, 0)
+
+    # Create a copy of the surface that is closed in theta and phi, and covers the
```
andrewgiuliani, line 379: stray comments

examples/3_Advanced/coil_force_optimization/optimization_tools.py
```diff
@@ -0,0 +1,647 @@
...
+
+    # MAKE DIRECTORY FOR EXPORT
+
+    UUID = uuid.uuid4().hex  # unique id for each optimization
+    OUTPUT_DIR = OUTPUT_DIR + UUID + "/"  # Directory for output
+    os.makedirs(OUTPUT_DIR, exist_ok=True)
+
+    # EXPORT VTKS
+
+    # forces = []
```
andrewgiuliani, line 521: should this be commented out?

src/simsopt/util/permanent_magnet_helper_functions.py
```diff
@@ -351,7 +354,7 @@ def initialize_coils(config_flag, TEST_DIR, s, out_dir=''):
     return base_curves, curves, coils
 
 
-def calculate_on_axis_B(bs, s):
+def calculate_on_axis_B(bs, s, print_out=True):
     """
     Check the average, approximate, on-axis
     magnetic field strength to make sure the
```
andrewgiuliani, line 360: it's not really the magnetic axis that you're calculating the the modB average right?

andrewgiuliani: maybe `calculate_modB_on_major_radius` ?

src/simsopt/field/coil.py
andrewgiuliani: confused by `JaxCurrent`, not sure about its use case.  even with a Jax-based BiotSavart class, I don't think we can Jax through this class ? maybe I'm mistaken

src/simsopt/field/coil.py
```diff
@@ -172,36 +299,133 @@ def apply_symmetries_to_currents(base_currents, nfp, stellsym):
...
+    ----------
+    curves : list of Curve
+        List of base curves.
+    currents : list of Current
+        List of base current objects.
+    nfp : int
+        Number of field periods (rotational symmetry).
+    stellsym : bool
+        Whether to apply stellarator symmetry.
+    regularization : Regularization
```
andrewgiuliani, line 386: typo: regularizations

src/simsopt/field/coil.py
```diff
@@ -1,29 +1,45 @@
...
+        The geometric curve describing the coil shape.
+    current : Current
+        The current object describing the electric current in the coil.
+    regularization : Regularization
+        The regularization object for the coil corresponding to the coil cross section. 
+        Default is a circular cross section with radius 0.05.
     """
 
-    def __init__(self, curve, current):
+    def __init__(self, curve, current, regularization=regularization_circ(0.05)):
```
andrewgiuliani, line 39: I'm not sure about by default making a coil have a regularization, when some problems don't have this.  I'd lean towards making this be `None`

src/simsopt/field/coil.py
```diff
@@ -172,36 +299,133 @@ def apply_symmetries_to_currents(base_currents, nfp, stellsym):
...
+    coils : list of Coil
+        List of Coil objects with symmetries applied.
     """
 
     assert len(curves) == len(currents)
     curves = apply_symmetries_to_curves(curves, nfp, stellsym)
     currents = apply_symmetries_to_currents(currents, nfp, stellsym)
-    coils = [Coil(curv, curr) for (curv, curr) in zip(curves, currents)]
+    if regularizations is None:
+        regularizations = [regularization_circ(0.05) for _ in curves]
```
andrewgiuliani, line 400: I'm not sure about including a regularization in a coil when some problems don't have this, I would have thought this would default to `None`

src/simsopt/field/force.py
```diff
@@ -2,55 +2,182 @@
...
+    mutual_field = BiotSavart(mutual_coils).set_points(coil.curve.gamma())
+    B_mutual = mutual_field.B()
+    mutualforce = np.cross(coil.current.get_value() * tangent, B_mutual)
+    selfforce = self_force(coil)
+    return (selfforce + mutualforce)
+
+
+def coil_net_force(coil, other_coils):
+    """
+    Compute the net forces on a list of coils.
```
andrewgiuliani, line 64: can we include units? can we clarify that this is the integrated pointwise force on a coil?

src/simsopt/field/force.py
```diff
@@ -2,55 +2,182 @@
...
+        array: Array of torques.
+    """
+    gamma = coil.curve.gamma()
+    center = coil.curve.centroid()
+    return np.cross(gamma - center, coil_force(coil, other_coils))
+
+
+def coil_net_torque(coil, other_coils):
+    """
+    Compute the net torques on a list of coils.
```
andrewgiuliani, line 100: same comments as above

src/simsopt/field/force.py
```diff
@@ -2,55 +2,182 @@
...
     return jnp.cross(I * t, B)
 
 
-def self_force(coil, regularization):
+def self_force(coil):
     """
     Compute the self-force of a coil.
+
+    Args:
+        coil (Coil): Coil to compute the self-force on.
```
andrewgiuliani, line 137: pointwise force per unit length on a coil?

src/simsopt/field/force.py
```diff
@@ -60,120 +187,206 @@ def lp_force_pure(gamma, gammadash, gammadashdash, quadpoints, current, regulari
 
     where :math:`\vec{F}` is the Lorentz force, :math:`F_0` is a threshold force,
```
andrewgiuliani, line 188: can address issue #515 in this PR

examples/3_Advanced/coil_force_optimization/optimization_tools.py
```diff
@@ -0,0 +1,619 @@
...
+        "arclength_weight": ARCLENGTH_WEIGHT,
+        "lpcurveforce": float(lpcurveforce),
+        "lpcurvetorque": float(lpcurvetorque),
+        "squaredmeanforce": float(squaredmeanforce),
+        "squaredmeantorque": float(squaredmeantorque),
+        "JF": float(JF.J()),
+        "Jf": float(Jf.J()),
+        "gradient_norm": np.linalg.norm(JF.dJ()),
+        # "linking_number":           LinkingNumber(curves).J(), #TODO: UNCOMMENT!!!!!!!
+        "lengths": [float(J.J()) for J in Jls],
```
andrewgiuliani, line 569: why do you have to cast as a float here?

src/simsopt/field/coil.py
andrewgiuliani: a second comment about this is adding a new attribute to `Coil` might break a lot of users' jsons... a suggestion would be create a wrapper coil class like `RegularizedCoil` but I'm open to other solutions that require less code.

src/simsopt/geo/curve_planar_elliptical_cylindrical.py
```diff
@@ -0,0 +1,172 @@
...
+    CurvePlanarEllipticalCylindrical is a translated and rotated Curve in r, phi, and z coordinates.
+    It is parametrized by an ellipse with semi-major axis a and semi-minor axis b, where a lies along 
+    the R vector and b along the Z vector. This curve is then translated and rotated in cylindrical coordinates
+    to a major radius R0, toroidal angle phi, and vertical position Z0. The curve is then rotated by angles
+    r_rotation, phi_rotation, and z_rotation in the r, phi, and z directions, respectively, about the minimum 
+    Z location of the curve.
+    """
+
+    def __init__(self, quadpoints, a, b, dofs=None):
+        if isinstance(quadpoints, int):
```
andrewgiuliani, line 105: docstring

src/simsopt/geo/curve_planar_elliptical_cylindrical.py
```diff
@@ -0,0 +1,172 @@
...
+    final_gamma = cyl_to_cart(g4)  # convert back to cartesian
+    return final_gamma
+
+
+class CurvePlanarEllipticalCylindrical(JaxCurve):
+    """
+    CurvePlanarEllipticalCylindrical is a translated and rotated Curve in r, phi, and z coordinates.
+    It is parametrized by an ellipse with semi-major axis a and semi-minor axis b, where a lies along 
+    the R vector and b along the Z vector. This curve is then translated and rotated in cylindrical coordinates
+    to a major radius R0, toroidal angle phi, and vertical position Z0. The curve is then rotated by angles
```
andrewgiuliani, line 99: what does it mean to rotate something to a toroidal angle phi? is the coil's centroid moved to toroidal angle phi?

src/simsopt/geo/curve_planar_elliptical_cylindrical.py
```diff
@@ -0,0 +1,172 @@
...
+    r_rotation, phi_rotation, and z_rotation in the r, phi, and z directions, respectively, about the minimum
```
andrewgiuliani, line 100: what does it mean to rotate by these angles? Are you rotating about the local cylindrical basis axes?

src/simsopt/geo/curve.py
```diff
@@ -841,14 +937,227 @@ def wrap(data):
...
     pointData = {'idx': data}
-
     if extra_data is not None:
         pointData = {**pointData, **extra_data}
 
     polyLinesToVTK(str(filename), x, y, z, pointsPerLine=ppl, pointData=pointData)
 
 
-def create_equally_spaced_curves(ncurves, nfp, stellsym, R0=1.0, R1=0.5, order=6, numquadpoints=None):
+def setup_uniform_grid(s, s_inner, s_outer, Nx, Ny, Nz, Nmin_factor=2.5):
```
andrewgiuliani, line 946: this doesn't seem the right place for this function, maybe in the util directory?

src/simsopt/geo/curve.py
```diff
@@ -841,14 +937,227 @@ def wrap(data):
...
+    """
+    Generate a uniform 3D grid of points between two toroidal surfaces, with spacing and filtering to avoid coil 
+    overlap and respect stellarator symmetry.
+
+    This function is typically used to initialize candidate coil center locations for planar coil optimization. 
+    It computes a uniform grid in the bounding box between two surfaces, then (for nfp > 1) removes points 
+    outside the fundamental sector and those too close to the symmetry plane, to avoid overlap after symmetry operations.
```
andrewgiuliani, line 953: what's a fundamental sector?

src/simsopt/geo/curve.py
```diff
@@ -841,14 +937,227 @@ def wrap(data):
...
+    # This is probably not robust for every stellarator but seems to work
+    # reasonably well for the Landreman/Paul QA/QH in the code.
+    if nfp > 1:
+        inds = []
+        for i in range(Nx):
+            for j in range(Ny):
+                for k in range(Nz):
+                    phi = np.arctan2(Y[i, j, k], X[i, j, k])
+                    if nfp == 4:
+                        phi2 = np.arctan2(R / 1.4, X[i, j, k])
```
andrewgiuliani, line 1050: I don't understand what phi2 is? why for nfp=4 is R/1.4 and nfp=3  R/2?

src/simsopt/geo/curve.py
```diff
@@ -841,14 +937,227 @@ def wrap(data):
...
+
+    # Initialize uniform grid
+    dx = (x_max - x_min) / (Nx - 1)
+    dy = (y_max - y_min) / (Ny - 1)
+    dz = 2 * z_max / (Nz - 1)
+    Nmin = min(dx, min(dy, dz))
+
+    # This is not a guarantee that coils will not touch but inductance
+    # matrix blows up if they do so it is easy to tell when they do
+    R = Nmin / Nmin_factor
```
andrewgiuliani, line 1012: why doesn't this guarantee the coils won't overlap? if R < h/2 isn't that enough?

src/simsopt/geo/curve.py
andrewgiuliani: `JaxCurveXYZFourier` was included to illustrate how to use `JaxCurve`, on which many curve representations rely, so I do not think either should be removed from simsopt.

That said, I agree with @mishapadidar that `create_equally_spaced_curves` doesn't need to support `JaxCurveXYZFourier`

src/simsopt/field/selffield.py
```diff
@@ -53,45 +54,42 @@ def B_regularized_singularity_term(rc_prime, rc_prime_prime, regularization):
...
+def B_regularized(coil):
     """Calculate the regularized field on a coil following the Landreman and Hurwitz method"""
     return B_regularized_pure(
         coil.curve.gamma(),
         coil.curve.gammadash(),
         coil.curve.gammadashdash(),
         coil.curve.quadpoints,
         coil._current.get_value(),
-        regularization,
+        coil.regularization,
```
andrewgiuliani, line 84: seems like instead of `regularization` being a `Coil` attribute, it can be an argument to this function

src/simsopt/geo/curve.py
```diff
@@ -441,21 +477,72 @@ def __init__(self, quadpoints, gamma_pure, **kwargs):
...
         sopp.Curve.set_dofs(self, dofs)
 
     def gamma_impl(self, gamma, quadpoints):
         r"""
         This function returns the x,y,z coordinates of the curve :math:`\Gamma`.
         """
-
         gamma[:, :] = self.gamma_impl_jax(self.get_dofs(), quadpoints)
 
+    def incremental_arclength_pure(self, dofs):
```
andrewgiuliani, line 504: we've typically placed pure functions outside of the classes

src/simsopt/geo/curve.py
```diff
@@ -441,21 +477,72 @@ def __init__(self, quadpoints, gamma_pure, **kwargs):
...
+        return jnp.linalg.norm(gammadash, axis=1)
+
+    def incremental_arclength(self):
+        """
+        This function returns the incremental arclength of the curve.
+        """
+        return self.incremental_arclength_jax(self.get_dofs())
+
+    # @jit
+    def frenet_frame_pure(self, dofs):
```
andrewgiuliani, line 518: we've typically placed pure functions outside of the classes

src/simsopt/geo/curve.py
```diff
@@ -441,21 +477,72 @@ def __init__(self, quadpoints, gamma_pure, **kwargs):
...
+        t = (1./l[:, None]) * gammadash
+
+        tdash = (1./l[:, None])**2 * (l[:, None] * gammadashdash
+                                      - (inner(gammadash, gammadashdash)/l)[:, None] * gammadash
+                                      )
+        n = (1./norm(tdash))[:, None] * tdash
+        b = jnp.cross(t, n, axis=1)
+        return t, n, b
+
+    def frenet_frame(self):
```
andrewgiuliani, line 539: hm, since JaxCurve is a Curve, doesn't `frenet_frame` already exist for this class?

src/simsopt/geo/curve.py
```diff
@@ -653,6 +738,17 @@ def num_dofs(self):
         This function returns the number of dofs associated to the curve.
         """
         return self.curve.num_dofs()
+    
+    def centroid(self):
```
andrewgiuliani, line 742: hm, since JaxCurve is a Curve, doesn't `centroid` already exist for this class and is added in the PR?

src/simsopt/geo/curve.py
andrewgiuliani, line 946: suggest renaming from `setup_uniform_grid` because this function does more than just return a uniform grid of points

src/simsopt/geo/curveplanarfourier.py
```diff
@@ -73,3 +80,136 @@ def set_dofs(self, dofs):
...
+    return jnp.transpose(jnp.vstack((jnp.vstack(((1.0 - 2 * (q_norm[2] * q_norm[2] + q_norm[3] * q_norm[3])) * x_curve_in_plane
+                                                 + 2 * (q_norm[1] * q_norm[2] - q_norm[3] * q_norm[0]) * y_curve_in_plane
+                                                 + center[0], (1.0 - 2 * (q_norm[1] * q_norm[1] + q_norm[3] * q_norm[3])) * y_curve_in_plane
+                                                 + 2 * (q_norm[0] * q_norm[3] + q_norm[1] * q_norm[2]) * x_curve_in_plane
+                                                 + center[1])), 2 * (q_norm[1] * q_norm[3] - q_norm[0] * q_norm[2]) * x_curve_in_plane
+                                     + 2 * (q_norm[0] * q_norm[1] + q_norm[2] * q_norm[3]) * y_curve_in_plane
+                                     + center[2])))
+
+
+class JaxCurvePlanarFourier(JaxCurve):
```
andrewgiuliani, line 142: what's the purpose of this class, doesn't `CurvePlanarFourier` already exist?

src/simsopt/geo/curve.py
andrewgiuliani: I agree with @landreman, curious if there's a particular reason we need to support `JaxCurvePlanarFourier` in this function? Just trying to understand the use case, @akaptano.

andrewgiuliani (commented):
Hi @akaptano, thanks for doing all this code development! I think that this is great work.

One broader suggestion that I have is to further separate this PR into smaller chunks if possible.  For example, the utility files in the examples section could form their own PR themselves.  

Just a thought to make things a bit more manageable. Thanks again for your hard work!

mishapadidar (changes requested):
This is a very long diff. It contains edits to many different simsopt features. For example, this diff contains changes to permanent magnets, QFM, adds new coils objectives, add new coils types, modifieds the existing coil classes, adds new examples, adds new tests, etc. These many changes make this diff difficult and time-consuming to review -- for reference, after about an hour and a half of review time this morning, I am still not done. The best way to get your cool new features into simsopt is to _break this diff up_.

Here are some general comments in addition to my inline comments.

**Coil**
- Adding the coil `regularization` term is a significant change. The coil class in simsopt really reflects a filamentary coil, which differs from say the FiniteBuild coil class. The coils with regularization seem to be something in between, representing coils with a finite radius which are amenable to special computations. Seems like having a coil or curve class for this makes sense, i.e. CoilWithCrossSection (probably needs a better name).
- In general, can we rename `regularization` to something more descriptive such as `cross_section_shape`?

**Deprecation**
- This PR deprecates a number of features in the coil force computation. The functions have been named `..._deprecated`. Typically, we would keep around deprecated functions so that we could slowly phase them out. In this case, I think it is fine to remove them altogether. We need buy-in from other contributors.

**Examples**
- Many of the example scripts included are catered to executing large workloads. The aim of the examples is to teach users how to use simsopt. So while these scripts may be valuable to researchers, they do not exactly belong in the examples directory. 
- I suggest keeping the basic coil_forces example.
- To enhance the visibility of this work, I suggest adding the coil force terms to the list of regularization terms in the [coil optimization tutorial](https://simsopt.readthedocs.io/latest/example_coils.html#further-coil-regularization-terms).

akaptano:
Thank you _very much_ to everyone for their help here. This should be now passing all the checks again. I can submit a bunch of sub-PRs as suggested by Andrew and Misha but unfortunately this will probably number in the dozens, which seems like it might require even more effort from everyone. I have also not incorporated a bunch of code from the dipole and passive coil arrays yet -- I was still planning on two or three more PRs to get all my code merged. I can proceed with the many micro-PRs though if this PR is too much of a headache. 

Some other notes to help folks navigate this PR:
1. Misha notes (and I totally agree) that the docstrings could be improved (adding type and shape to each parameter), but this is a fix that is needed on the entire master branch, not specific to the PR here. I have also stopped running ./run_autopep because it generates so many changes with master. I am happy to fix all of this in a separate PR later.
2. There seems to be no clear consensus about keeping all the JaxCurve functionality. The main branch already has JaxCurve and JaxCurveXYZFourier classes, and I know that some developers (Antoine Baillod and others) use these JaxCurves regularly (although I'm not sure why), so there is a risk of breaking code. I will say that I have found these JaxCurve classes very useful for debugging various calculations, especially Jacobians, and this is precisely how the bug in the CurvePlanarFourier C++ jacobian calculation was eventually found. 
3. See above long comment for my thoughts on keeping the `downsample` parameter.
4. There are no real changes to the permanent magnets functionality. Andrew suggested changing `calculate_on_axis_B` to the much more correct `calculate_modB_on_major_radius` and all the permanent magnet examples were updated with the name change. 
5. There are lots of small fixes and improvements that I have made on this branch in the process of this PR that improve the code. I list some here:
- CurvePlanarFourier Jacobian is now correct
- CurveHelical, Wireframe, and other new functionality is better tested in the unit tests now
- Added function coils_to_vtk which saves the coil curves along with their currents, forces, torques, etc. 
- CurvePlanarFourier C++ and Python initialization was not correct. It accepted stellsym and nfp parameters, and then did nothing with them, so they are removed. 
- In test_biotsavart.py the test_flux_through_disk fails occasionally, even on the master branch. This is easily fixed by increasing the number of quadrature points a little. 
- Added a huge amount of improved docstrings, including on many functions that are not specific to this PR. 
- Unit tests now rigorously check all the JaxCurves pass all the curve objective tests and so forth
- New CurvePlanarEllipticalCylindrical class introduced and tested. I am not sure I understand the motivation for it, but it would be somewhat hard to replicate with the CurvePlanarFourier class because it uses a fairly different representation.
- As far as I can tell, the force and torque objectives now pass the Taylor tests, fixing the CI issues still plaguing the master branch, i.e. resolving Issue #485 .
- I rewrote the QFM jacobian calculation in a way that made more sense to me, with additional comments, and it does not change the result. But I am happy to revert this if folks do not appreciate the change. 

If we split this into more PRs, the first one still needs to be the force/torque overhaul, and the fixes to CurvePlanarFourier. Otherwise, the unit tests are liable to fail regardless. But I can strip out the new coil force examples, documentation, CurvePlanarEllipticalCylindrical class, etc. etc.

akaptano:
> a few lines of code coverage are missing, can we get these covered?

@andrewgiuliani coverage should be basically perfect now.

akaptano:
@mishapadidar @andrewgiuliani friendly monthly reminder

mishapadidar:
Hi, Alan, these are great changes. To continue with the review process, it would be helpful to break up this PR into more pieces. The idea is to try to separate the PR into self-contained pieces. We will be able to review smaller pieces more quickly. Here is a suggested breakdown.

1. `coil.py` and `curve.py` and relevant unit tests. `coil.py` file has mostly self-contained changed, like the `RegularizedCoil` class and a bunch of documentation. The `coils_to_vtk` probably needs to wait until a later PR because of dependencies. `curve.py` Seems like only small changes have been made, so might as well include it in this PR.
2. `force.py`, `selffield.py` and relevant unit tests. Also correct any examples that are broken, but don't add new examples yet. If this can somehow be split into 2 PRs, that would also be useful, but I suspect these functions all need to land at once. The goal here is to land the main coil-force functionality in a small PR.

At this point, the core functionality of this PR will have landed. The remaining bits such as `example_coils.rst`, the `coils_to_vtk` function, new examples, changes to the permanent magnet examples, or helpers can go in later PRs.

@akaptano @landreman @andrewgiuliani

src/simsopt/field/force.py
```diff
@@ -2,235 +2,1634 @@
...
+            by skipping through the array by a factor of ``downsample``,
+            e.g. curve.gamma()[::downsample, :]. 
+            Setting this parameter to a value larger than 1 will speed up the calculation,
+            which may be useful if the set of coils is large, though it may introduce
+            inaccuracy if ``downsample`` is set too large, or not a multiple of the 
+            total number of quadrature points (since this will produce a nonuniform set of points). 
+            This parameter is used to speed up expensive calculations during optimization, 
+            while retaining higher accuracy for the other objectives. 
+        regularizations (array, shape (m,)): 
+            Array of regularizations coming from finite cross-section for all m coils.
```
mishapadidar: can you put an `i.e. regularization_circ` here or similar?

src/simsopt/field/force.py
```diff
@@ -2,235 +2,1634 @@
...
+        else:
+            idxs = jnp.linspace(0, n-1, target_n).round().astype(int)
+            idxs = idxs[::downsample]
+        return arr[idxs, ...]
+    gammas = jnp.stack([subsample(g, min_npts) for g in gammas])
+    gammadashs = jnp.stack([subsample(g, min_npts) for g in gammadashs])
+    N = gammas.shape[0]
+
+    # Compute Lij, i != j
+    eps = 1e-10
```
mishapadidar: do we need the regularization? 
We should mention that the computation is regularized if coils are too close.
We could instead throw a warning if the coils are too close.

src/simsopt/field/force.py
```diff
@@ -2,235 +2,1634 @@
...
+    """
+    Pure function for computing the induced currents in a set of m passive coils with n quadrature points
+    due to a set of m' TF coils with n' quadrature points (and themselves). 
+
+    Args:
+        gammas (array, shape (m,n,3)): 
+            Array of passive coil positions for all m coils.
+        gammadashs (array, shape (m,n,3)): 
+            Array of passive coil tangent vectors for all m coils.
+        gammas_TF (array, shape (m',n',3)):
```
mishapadidar: Can we use the naming convention "active" vs "passive" coils rather than specifically say "T"? the active coil type may vary.

src/simsopt/field/force.py
```diff
@@ -2,235 +2,1634 @@
...
+            Array of TF coil tangent vectors for all m' coils.
+        currents_TF (array, shape (m',)): 
+            Array of TF coil current.
+        downsample (int): 
+            Factor by which to downsample the quadrature points 
+            by skipping through the array by a factor of ``downsample``,
+            e.g. curve.gamma()[::downsample, :]. 
+            Setting this parameter to a value larger than 1 will speed up the calculation,
+            which may be useful if the set of coils is large, though it may introduce
+            inaccuracy if ``downsample`` is set too large, or not a multiple of the
```
mishapadidar: same comment on multiple here and elsewhere

src/simsopt/field/force.py
```diff
@@ -2,235 +2,1634 @@
...
+            jnp.asarray([c.curve.gamma() for c in self.coils_to_target]),
+            jnp.asarray([c.curve.gammadash() for c in self.coils_to_target]),
+            jnp.asarray([c.current.get_value() for c in self.coils_to_target]),
+            self.downsample
+        ]
+
+        return self.J_jax(*args)
+
+    @derivative_dec
+    def dJ(self):
```
mishapadidar: docstring

src/simsopt/field/force.py
```diff
@@ -2,235 +2,1634 @@
...
+
+    return_fn_map = {'J': J, 'dJ': dJ}
+
+
+def net_fluxes_pure(gammas, gammadashs, gammas2, gammadashs2, currents2, downsample):
+    r"""
+    Calculate the net magnetic flux through a set of m coils with n quadrature points
+    due to another set of m' coils with n' quadrature points.
+
+    This function computes the total magnetic flux passing through a set of coils
```
mishapadidar: first two lines of docstring repeats itself. keep the second one without the mention of the number of quadpoints

src/simsopt/field/force.py
```diff
@@ -2,235 +2,1634 @@
...
+                lambda _: jnp.zeros(3),
+                lambda _: jnp.asarray(jnp.sum(
+                    jnp.cross(gammadashs[j], pt - gammas[j]) /
+                    (jnp.linalg.norm(pt - gammas[j] + eps, axis=1) ** 3)[:, None],
+                    axis=0
+                ) * currents[j]),
+                operand=None
+            )
+
+        def biot_savart_from_j2(j2):
```
mishapadidar: clean up subsampling logic and separate biot-savart and A function in to pures

src/simsopt/field/force.py
```diff
@@ -2,235 +2,1634 @@
...
+    def subsample(arr, target_n):
+        arr = jnp.asarray(arr)
+        n = arr.shape[0]
+        if n == target_n:
+            idxs = jnp.arange(0, n, downsample)
+        else:
+            idxs = jnp.linspace(0, n-1, target_n).round().astype(int)
+            idxs = idxs[::downsample]
+        return arr[idxs, ...]
+    gammas = jnp.stack([subsample(g, min_npts) for g in gammas])
```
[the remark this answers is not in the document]
akaptano: See above comment

src/simsopt/field/force.py
```diff
@@ -2,235 +2,1634 @@
...
+    # Compute Lij, i != j
+    eps = 1e-10
+    r_ij = gammas[None, :, None, :, :] - gammas[:, None, :, None, :] + eps
+    rij_norm = jnp.linalg.norm(r_ij, axis=-1)
+    gammadash_prod = jnp.sum(gammadashs[None, :, None, :, :] * gammadashs[:, None, :, None, :], axis=-1)
+
+    # Double sum over each of the closed curves for off-diagonal elements
+    Lij = jnp.sum(jnp.sum(gammadash_prod / rij_norm, axis=-1), axis=-1) / jnp.shape(gammas)[1] ** 2
+
+    diag_values = jnp.sum(jnp.sum(gammadash_prod / jnp.sqrt(rij_norm ** 2 + regularizations[None, :, None, None]),
```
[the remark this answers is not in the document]
akaptano: leaning against it for now since this is only calculated at this one location right now

src/simsopt/field/force.py
```diff
@@ -2,235 +2,1634 @@
...
+        J = \frac{1}{2}\sum_{i,j}I_i L_{ij} I_j
+
+    where :math:`L_{ij}` is the coil inductance matrix (positive definite),
+    and :math:`I_i` is the current in the ith coil. The units of the objective function are MJ (megajoules).
 
-    where :math:`\vec{F}` is the Lorentz force, :math:`F_0` is a threshold force,  
-    and :math:`\ell` is arclength along the coil.
+    Args:
+        coils_to_target (list of Coil, shape (m,)): 
+            List of coils contributing to the total energy.
```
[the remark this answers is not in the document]
akaptano: clarified this throughout the file

src/simsopt/field/force.py
```diff
@@ -2,235 +2,1634 @@
...
+            Net magnetic flux through each coil in the first set.
+    """
+    # Downsample if desired
+    gammas = gammas[:, ::downsample, :]
+    gammadashs = gammadashs[:, ::downsample, :]
+    gammas2 = gammas2[:, ::downsample, :]
+    gammadashs2 = gammadashs2[:, ::downsample, :]
+    rij_norm = jnp.linalg.norm(gammas[:, :, None, None, :] - gammas2[None, None, :, :, :], axis=-1)
+    # sum over the currents, and sum over the biot savart integral
+    A_ext = jnp.sum(currents2[None, None, :, None] * jnp.sum(gammadashs2[None, None, :, :, :] / rij_norm[:, :, :, :, None], axis=-2), axis=-2) / jnp.shape(gammadashs2)[1]
```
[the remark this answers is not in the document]
akaptano: I think this is the only place, omitting this change for now.

src/simsopt/field/force.py
```diff
@@ -2,235 +2,1634 @@
...
+            total number of quadrature points (since this will produce a nonuniform set of points). 
+            This parameter is used to speed up expensive calculations during optimization, 
+            while retaining higher accuracy for the other objectives. 
+    Returns:
+        float: The squared mean force.
+    """
+    all_lengths = [g.shape[0] for g in gammas] + [g2.shape[0] for g2 in gammas2]
+    min_npts = min(all_lengths)
+
+    def subsample(arr, target_n):
```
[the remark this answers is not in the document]
akaptano: removed everywhere

src/simsopt/field/force.py
```diff
@@ -2,235 +2,1634 @@
...
+    (downsampled to have the smallest number, denoted n, of quadrature points),
+    due to themselves and another set of m' coils with varying quadrature points
+    (downsampled to n quadrature points).
+
+    The objective function is
+
+    .. math:
+        J = \sum_i \left(\frac{\int \frac{d\vec{F}_i}{d\ell_i} d\ell_i}{L_i}\right)^2
+
+    where :math:`\frac{d\vec{F}_i}{d\ell_i}` is the Lorentz force per unit length,
```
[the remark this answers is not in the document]
akaptano: fixed throughout file, thanks

src/simsopt/field/force.py
akaptano: already did from above comments, thanks

andrewgiuliani (changes requested):

docs/source/example_coils.rst
```diff
@@ -374,6 +374,11 @@ many of which are illustrated in
 - :obj:`~simsopt.geo.CurveSurfaceDistance`: Useful for ensuring the minimum coil-to-plasma distance is at least a specified target value.
 - :obj:`~simsopt.geo.ArclengthVariation`: Ensures the curves are parameterized using (approximately) a uniform-arclength parameter.
 - :obj:`~simsopt.geo.LinkingNumber`: Prevents coils from becoming topologically linked to each other.
+- :obj:`~simsopt.field.LpCurveForce`: Penalizes the pointwise Lorentz force per unit length on each coil.
```
mishapadidar: thanks for this

src/simsopt/field/force.py
```diff
@@ -2,235 +2,1724 @@
...
+
+    Returns:
+        array (shape (m,m)): Full inductance matrix Lij.
+    """
+    gammas = jnp.asarray(gammas)[:, ::downsample, :]
+    gammadashs = jnp.asarray(gammadashs)[:, ::downsample, :]
+    N = gammas.shape[0]
+
+    # Compute Lij, i != j
+    eps = 1e-10  # small constant to avoid division by zero for mutual inductance between coil_i and itself
```
mishapadidar: rather than hardcoding this, can it be an argument with a default value?

src/simsopt/field/force.py
```diff
@@ -2,235 +2,1724 @@
...
+    which may have different finite-build parameters. In order to avoid buildup of optimizable 
+    dependencies, it directly computes the BiotSavart law terms, instead of relying on the existing
+    C++ code that computes BiotSavart related terms. The two sets of coils may contain 
+    coils with all different number of quadrature points and different types of cross-sections.
 
-    where :math:`\vec{F}` is the Lorentz force, :math:`F_0` is a threshold force,  
-    and :math:`\ell` is arclength along the coil.
+    Args:
+        coils_to_target (list of RegularizedCoil, shape (m,)): 
+            List of coils to use for computing LpCurveForce.
```
mishapadidar: List of coils on which the LpCurveForce is computed.

src/simsopt/field/force.py
```diff
@@ -2,235 +2,1724 @@
...
+    coils with all different number of quadrature points and different types of cross-sections.
 
-    where :math:`\vec{F}` is the Lorentz force, :math:`F_0` is a threshold force,  
-    and :math:`\ell` is arclength along the coil.
+    Args:
+        coils_to_target (list of RegularizedCoil, shape (m,)): 
+            List of coils to use for computing LpCurveForce. 
+        source_coils (list of RegularizedCoil, shape (m',)): 
+            List of coils that provide forces on the first set of coils but that
+            we do not care about optimizing their forces.
```
mishapadidar: a bit confused on "we do not care about optimizing their forces.". rephrase please. Maybe just "List of additional coils that provide forces on the coils_to_target. Forces are not computed on the source_coils"

src/simsopt/field/force.py
```diff
@@ -2,235 +2,1724 @@
...
 
     .. math::
-        J = \frac{1}{p}\left(\int \text{max}(|\vec{F}| - F_0, 0)^p d\ell\right)
+        J = \frac{1}{p}\sum_i\frac{1}{L_i}\left(\int \text{max}(|d\vec{F}/d\ell_i| - dF_0/d\ell_i, 0)^p d\ell_i\right)
+
+    where :math:`\frac{d\vec{F}_i}{d\ell_i}` is the Lorentz force per unit length, 
+    in units of MN/m, where MN = meganewtons. The units of the objective function are therefore (MN/m)^p.
+    :math:`d\ell_i` is the arclength along the ith coil,
+    :math:`L_i` is the total coil length,
+    and :math:`dF_0/d\ell_i` is a threshold force at the ith coil.
```
mishapadidar: I assume the threshold is just a constraint bound right? the notation `dF_0/d\ell_i` makes it looks the threshold is a dependent quantity. Replace with `tau_i` or similar.

src/simsopt/field/force.py
akaptano: fixed throughout, thanks

src/simsopt/field/force.py
```diff
@@ -2,235 +2,1724 @@
...
+        return jnp.sum(force_density * gammadash_norm_i, axis=0) / npts1
+
+    mean_forces = vmap(mean_force_group1, in_axes=(0, 0, 0, 0, 0))(
+        jnp.arange(n1), gammas_targets, tangents, gammadash_norms, currents_targets
+    )
+    # mean_forces is computed without mu_0/(4*pi) factor in B_mutual, so we need to multiply by (mu_0/(4*pi))^2 = 1e-14
+    # Then convert from (N/m)^2 to (MN/m)^2 by dividing by (1e6)^2 = 1e12
+    # Net factor: 1e-14 / 1e12 = 1e-26  
+    mean_forces_squared = jnp.sum(jnp.linalg.norm(mean_forces, axis=-1) ** 2)
+    return mean_forces_squared * 1e-26
```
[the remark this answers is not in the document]
akaptano: yes, fixed from above

PR #512, opened 2025-05-11
Improve CurveHelical

landreman:
This PR addresses several issues with `CurveHelical`:

* Previously `order` had been defined so the maximum mode number was `order - 1`, inconsistent with the rest of simsopt where the maximum mode number is `order` (inclusive). With this PR, the maximum mode number is changed to be `order` for consistency.
* The dofs had included a mode amplitude for the k=0 term in the sine terms, which is meaningless. This dof is now removed.
* There wasn't much description in the docstring. More detail is added here.
* Some of the curve parameter names weren't consistent with the Cary-Hanson and Hanson-Cary papers, which are a major motivation for this curve type. E.g. n0 instead of m, r0 instead of r. In this PR, these parameters are renamed to match the papers.
* The dofs previously used the default names x0, x1, ... Now, the dof names are set to `A_0`, `A_1`, ..., `B_1`, ... for better intelligibility and for consistency with Cary-Hanson and Hanson-Cary.

### DIFF
--- a/src/simsopt/geo/curvehelical.py
+++ b/src/simsopt/geo/curvehelical.py
@@ -1,69 +1,114 @@
-import jax.numpy as jnp
-from math import pi
 import numpy as np
+import jax.numpy as jnp
 from .curve import JaxCurve
 
-__all__ = ['CurveHelical']
+__all__ = ["CurveHelical"]
 
 
-def jaxHelicalfouriercurve_pure(dofs, quadpoints, order, n0, l0, R0, r0):
-    A = dofs[:order]
-    B = dofs[order:]
-    phi = quadpoints*2*pi*l0
-    m, phiV = jnp.meshgrid(jnp.arange(order), phi)
-    AcosArray = jnp.sum(A*jnp.cos(m*phiV*n0/l0), axis=1)
-    BsinArray = jnp.sum(B*jnp.sin(m*phiV*n0/l0), axis=1)
-    eta = n0*phi/l0+AcosArray+BsinArray
-    gamma = jnp.zeros((len(quadpoints), 3))
-    gamma = gamma.at[:, 0].add((R0 + r0 * jnp.cos(eta)) * jnp.cos(phi))
-    gamma = gamma.at[:, 1].add((R0 + r0 * jnp.cos(eta)) * jnp.sin(phi))
-    gamma = gamma.at[:, 2].add(-r0 * jnp.sin(eta))
+def jaxHelicalfouriercurve_pure(dofs, quadpoints, order, m, ell, R0, r):
+    """Pure function for the position vector used by CurveHelical."""
+    A = dofs[: order + 1]
+    B = jnp.concatenate(
+        (jnp.zeros(1), dofs[order + 1 :])
+    )  # Add 0 at the start for k = 0
+    phi = quadpoints * 2 * jnp.pi * ell
+    k, phi_2D = jnp.meshgrid(jnp.arange(order + 1), phi)
+    eta = m * phi / ell + jnp.sum(
+        A * jnp.cos(k * phi_2D * m / ell) + B * jnp.sin(k * phi_2D * m / ell), axis=1
+    )
+    R = R0 + r * jnp.cos(eta)
+    x = R * jnp.cos(phi)
+    y = R * jnp.sin(phi)
+    z = -r * jnp.sin(eta)
+    gamma = jnp.column_stack((x, y, z))
     return gamma
 
 
 class CurveHelical(JaxCurve):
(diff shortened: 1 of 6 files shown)


akaptano (approved):
This looks good to me, no notes!

src/simsopt/geo/curvehelical.py
```diff
@@ -1,69 +1,112 @@
...
+    Toroidal Hybrid (CTH) device at Auburn University.
+
+    The helical coil lies on a circular-cross-section axisymmetric torus with
+    major radius :math:`R_0` and minor radius :math:`r`. Following Hanson and
+    Cary, the position along the coil is written in terms of a poloidal angle
+    on the winding surface :math:`\eta` and the standard toroidal angle :math:`\phi`:
+
+    .. math::
+        \begin{align*}
+        x &= [R_0 + r \cos(\eta)] \cos(\phi), \\
```
andrewgiuliani, line 45: This is not exactly the formula for R and Z in (14a,b) of Stochasticity reduction (Cary Hanson) and it seems like the representation in that paper is more general than what is here -- if I've understood correctly can this be clarified in docstring?

src/simsopt/geo/curvehelical.py
```diff
@@ -1,69 +1,112 @@
...
-        self.n0 = n0
-        self.l0 = l0
+        self.m = m
+        self.ell = ell
         self.R0 = R0
-        self.r0 = r0
-        self.coefficients = [np.zeros((order,)), np.zeros((order,))]
+        self.r = r
+        self.coefficients = np.zeros(self.num_dofs())
+        dof_names = [f"A_{i}" for i in range(order + 1)] + [f"B_{i}" for i in range(1, order + 1)]
```
andrewgiuliani, line 95: can we have a simple unit test that the dof names are what we think they are?  I think there's already some infrastructure for this (see below).

https://github.com/hiddenSymmetries/simsopt/blob/a76cec3308f192c9a9ac25cec22e7a599a7bedca/tests/geo/test_surface.py#L448

andrewgiuliani (dismissed):
looks good to me!

src/simsopt/geo/curvehelical.py
```diff
@@ -1,69 +1,117 @@
...
-    BsinArray = jnp.sum(B*jnp.sin(m*phiV*n0/l0), axis=1)
-    eta = n0*phi/l0+AcosArray+BsinArray
-    gamma = jnp.zeros((len(quadpoints), 3))
-    gamma = gamma.at[:, 0].add((R0 + r0 * jnp.cos(eta)) * jnp.cos(phi))
-    gamma = gamma.at[:, 1].add((R0 + r0 * jnp.cos(eta)) * jnp.sin(phi))
-    gamma = gamma.at[:, 2].add(-r0 * jnp.sin(eta))
+__all__ = ["CurveHelical"]
+
+
+def jaxHelicalfouriercurve_pure(dofs, quadpoints, order, m, ell, R0, r):
```
mishapadidar, line 8: camelcase or underscore based naming?

src/simsopt/geo/curvehelical.py
```diff
@@ -1,69 +1,117 @@
...
-        quadpoints: number of grid points/resolution along the curve;
-        order:  number of fourier coefficients, i.e., the length of the array A (or B);
-        n0:  toroidal periodicity/number of field periods
-        l0:  number of :math:`2\pi` turns in :math:`\phi`
-        R0:  major radius
-        r0:  minor radius
-    '''
-
-    def __init__(self, quadpoints, order, n0, l0, R0, r0, **kwargs):
+        quadpoints: (int or array-like) Grid points (or number thereof) along the curve.
```
mishapadidar, line 79: does quadpoints go from 0 to 2pi or from 0 to 1? The functions in the documentation make me think its from 0 to 2pi which differs from our convention.

mishapadidar (commented):
left some comments regarding documentation and naming

src/simsopt/geo/curvehelical.py
landreman: I added some text to clarify this in the docstring above. The fact that angles go up to 1 in some parts of simsopt and up to 2pi in other parts is a major confusion and potential source of errors. We really should have angles go up to 2pi everywhere, for consistency with the stellarator literature. Changing all the angles to go up to 2pi would be a big undertaking and would break many people's scripts, but might be a good idea (for another PR!)

smiet: Consistency would be much appreciated, but the [0,1) convention is much more common and has my personal preference.

smiet (approved):
LGTM, no changes.

Issue #482, opened 2025-04-16
Instructions for compiling simsoptpp with CMake are outdated


kchammond:
The documentation (`docs/source/cpp.rst`) provides some helpful instructions for how to compile changes to the C++ code without having to do a full reinstall of simsopt. These have worked for me in the past, but for more recent versions, they do not work anymore.

The issue is related to the location of the shared object file for simsoptpp. According to the instructions, it is assumed that the initial compilation (e.g. `pip install -e .`) should have created a shared object file in the `src` directory. The instructions say to replace this file with a soft link to the updated shared object file generated by `make -j`. However, when I compile the latest versions with pip, no shared object file is stored in `src`. Rather, it is stored in a subdirectory of the virtual environment instance under which the compilation with pip is performed.

I will note that an additional copy of the shared object file is stored in a subdirectory of the `build` directory in the repository directory (`build/cp312-cp312-linux_x86_64`). However, this file does not appear to be used by simsopt. Replacing it with a soft link has no effect on the execution of Python scripts that call simsoptpp functions. The key file to replace seems to be the one in the virtual environment directory as described above.

The following alternate set of commands works for me, both on my Mac laptop and on a Princeton Linux cluster:

```
# Same as before:
git clone --recursive git@github.com:hiddenSymmetries/simsopt.git
cd simsopt
pip3 install -e .
mkdir cmake-build
cd cmake-build
cmake ..
make -j

# Proposed replacement:
cd [path-to-virtual-environment]/lib/python3.12/site-packages
rm simsoptpp.cpython-312-x86_64-linux-gnu.so
ln -s [path-to-simsopt-repo]/cmake-build/simsoptpp.cpython-312-x86_64-linux-gnu.so .
```

As a caveat, the name of the .so file and the subdirectory name that includes the Python version ("python3.12") will likely vary from one computer to the next.

I'd be interested if others have experienced this issue and/or have thoughts on how to address the somewhat more convoluted procedure in the documentation.

mbkumar:
@kchammond 
I believe the updates to build tools could have caused this change in behavior. The documentation should be updated to reflect this. Can you please update the docs and submit a PR?

Thanks

PR #513, opened 2025-05-11
Resolve issue 482: update instructions for recompiling simsoptpp

kchammond:
This PR is a proposed resolution to issue #482.

The previous version of the instructions for recompiling ``simsoptpp`` presumed that the shared object file for simsoptpp used by Python would be stored in the ``src`` directory. However, this is no longer the case in the latest versions, and the instructions in their original form would no achieve the desired outcome.

This commit updates the instructions to be agnostic to the location of the shared object file used by Python, using the Python command ``print(simsoptpp.__file__)`` to directly look up the correct location.

### DIFF
--- a/docs/source/cpp.rst
+++ b/docs/source/cpp.rst
@@ -52,13 +52,18 @@ First, install ``SIMSOPT`` in the usual way::
 
     git clone --recursive git@github.com:hiddenSymmetries/simsopt.git
     cd simsopt
-    pip3 install -e .
+    pip install -e .
+
+Install ``cmake`` and ``pybind11`` using ``pip`` if they are not already installed::
+
+    pip install cmake
+    pip install pybind11
 
 Then, recompile the C++ code in a new directory (``cmake_build``)::
 
     mkdir cmake-build
     cd cmake-build
-    cmake ..
+    cmake .. -Dpybind11_DIR=$(python -c "import pybind11; print(pybind11.__path__[0])")/share/cmake/pybind11/
     make -j
 
 If the compilation executed successfully, there should now be a shared object file (ending with ``.so``) for ``simsoptpp`` in the ``cmake-build`` directory. After verifying that this ``.so`` file exists, execute the following command::
--- a/docs/source/cpp.rst
+++ b/docs/source/cpp.rst
@@ -46,20 +46,23 @@ For simple computations that are compute bound, we use SIMD (`Single Instruction
 CMake
 ^^^^^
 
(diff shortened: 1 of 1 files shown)


mbkumar:
> @kchammond Thanks for fixing this. A small addition is that this process doesn't work for my Mac unless I install pybind11 and point to pybind11 explicitly.
> 
> After `pip install "pybind11[global]"` it builds correctly, but only if I tell cmake where to find pybind11: cmake .. -DCMAKE_PREFIX_PATH=/opt/miniconda3/envs/simsopt_39/lib/python3.9/site-packages/pybind11/share/cmake/pybind11

This issue is general and would show up on Linux as well. When you do `pip install .`, pip downloads `pybind11` and passes the relevant flags to CMake to build `simsoptpp`. When we run cmake manually, we have to pass all the needed options explicitly. `pybind11` is the only package used on the C++ side that is downloaded by pip. Other packages are available as git submodules.

akaptano:
> > @kchammond Thanks for fixing this. A small addition is that this process doesn't work for my Mac unless I install pybind11 and point to pybind11 explicitly.
> > After `pip install "pybind11[global]"` it builds correctly, but only if I tell cmake where to find pybind11: cmake .. -DCMAKE_PREFIX_PATH=/opt/miniconda3/envs/simsopt_39/lib/python3.9/site-packages/pybind11/share/cmake/pybind11
> 
> This issue is general and would show up on Linux as well. When you do `pip install .`, pip downloads `pybind11` and passes the relevant flags to CMake to build `simsoptpp`. When we run cmake manually, we have to pass all the needed options explicitly. `pybind11` is the only package used on the C++ side that is downloaded by pip. Other packages are available as git submodules.

If that is true, let's add this line to the build instructions, yeah?

kchammond:
Thanks for catching this. Any chance this could be resolved by executing ``pip install cmake`` before making a call to ``cmake``? I ask because I've never encountered this issue before, and I wonder if pip's installation of cmake would automatically "know" where to find other pip-installed packages?

mbkumar:
No. `pip install cmake` won't install `pybind11`.

mbkumar:
@kchammond, If you start with a bare VE, you may find pybind11 not installed.  In your current VE, you might have installed it already.

kchammond:
@mbkumar Got it, thanks. I was able to reproduce the error after reinstalling in a fresh VE. I've updated the instructions to construct the flag for cmake in a way that (hopefully) automatically finds the location of pybind11. @akaptano I'd be interested if this works with your conda installation as well.

mbkumar (approved):

mbkumar:
@kchammond  Thanks!

PR #516, opened 2025-05-14
[skip ci] Code links in the ReadTheDocs pointing to GitHub

mbkumar:
Created the framework to point the code links in the documentation on the ReadTheDocs website to the simsopt GitHub repo. 
I updated the links only in two places in https://simsopt.readthedocs.io/mbk-gh_links/cpp.html. 

One drawback is branch names are not properly generated. The code to generate the branch names is working properly on my laptop and GH runner, but not on RTD runner. Since we point to only master and versioned releases in the docs, this should be OK. I didn't check the links generated for the releases yet. We can only do that after merging the PR and creating a new version of simsopt.

The README in the docs folder contains info on how to add links to GH code.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -7,6 +7,7 @@ build:
     python: "3.11"
   apt_packages:
     - doxygen
+    - libboost-all-dev
 
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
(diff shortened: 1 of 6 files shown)


mbkumar:
The next step is populating the documentation with correct links to GitHub. We can work on that within this PR or use a new PR and work on it gradually.

Issue #427, opened 2024-06-18
Numpy2.0 migration


smiet:
The container build is failing because the SPEC and VMEC2000 compilation step are failing. 

The migration guide to numpy2.0 is given [here](https://numpy.org/devdocs/numpy_2_0_migration_guide.html#numpy-2-migration-guide). 

I will work on getting SPEC up and running with numpy 2.0, @landreman can you have a look at VMEC? from [this link](https://numpy.org/devdocs/dev/depending_on_numpy.html#numpy-2-abi-handling) it looks like this is what should be added: 

```python
[build-system]
build-backend = ...
requires = [
    # Note for packagers: this constraint is specific to wheels
    # for PyPI; it is also supported to build against 1.xx still.
    # If you do so, please ensure to include a `numpy<2.0`
    # runtime requirement for those binary packages.
    "numpy>=2.0.0rc1",
    ...
]

[project]
dependencies = [
    "numpy>=1.23.5",
]
```

landreman:
Hm, I tried that change in the vmec2000 repo, but there are various errors in the vmec2000 CI now. Looping in @mbkumar @andrewgiuliani @daringli @rogeriojorge in case anyone can help.

smiet:
Now thinking we might want to temporarily force numpy<2.0 in all runners, there are other dependencies like f90wrap that use the deprecated oldest-supported-numpy that need to be patched, as well as that a new pyoculus needs to be uploaded to pypi as we resolve all the dependencies

smiet:
Update: f90wrap needs to be adapted as well to make spec (and simsopt) run with numpy 2.0. Working with the upstream to merge a [pull-request](https://github.com/jameskermode/f90wrap/pull/194 ). 

In the meantime we can try to force the runners to use a 1.x version of numpy, as I tried in PR #429

Dependencies that still need to be patched:
- [ ] F90wrap @smiet
- [ ] SPEC @smiet 
- [ ] pyoculus @smiet 
- [ ] VMEC2000 @mkumbar @landreman @andrewgiuliani @daringli @rogeriojorge

smiet:
nptyping does not seem to have any compatibility with numpy2.0, and when installed under 2.0 raises an exception upon import. 
Currently it's dependency is set to numpy<2.0, such that installing simsopt will downgrade numpy to 1.26.

Can we replace the dependency? @mbkumar, I see you added it, what does nptyping allow us to do that numpy.typing cannot provide?

[...]

PR #517, opened 2025-05-14
Cbs/randomgen 2 numpy

smiet:
randomgen was the last import that did not play nicely with numy2.0. (see #427)

Though it does now, It is probably beter to remove dependency on a rarely maintained package, so I updated to use numpy.random.PCG64DXSM.

### DIFF
--- a/docs/source/example_coils.rst
+++ b/docs/source/example_coils.rst
@@ -440,7 +440,7 @@ the previous examples, we additionally define::
   Jals = [ArclengthVariation(c) for c in base_curves]
 
   # Objective function for the coils and its perturbations
-  rg = np.random.Generator(PCG64(seed, inc=0))
+  rg = np.random.PCG64DXSM(seed)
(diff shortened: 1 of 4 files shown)


mbkumar:
Fixed a bug in instantiating the random Generator in the examples.

docs/source/example_coils.rst
```diff
@@ -440,7 +440,7 @@ the previous examples, we additionally define::
   Jals = [ArclengthVariation(c) for c in base_curves]
 
   # Objective function for the coils and its perturbations
-  rg = np.random.Generator(PCG64(seed, inc=0))
+  rg = np.random.PCG64DXSM(seed)
```
mbkumar, line 443: This should be `np.random.Generator(np.random.PCG64DXSM(seed))`

mbkumar (changes requested):

mbkumar (dismissed):
Looks good.

Issue #478, opened 2025-03-18
surface change_resolution method causes segmentation fault on subsequent evaluation of CPP methods


smiet:
When a `Surface` object's resolution is changed by the `change_resolution` method its `sopp.Curve`  methods will access unallocated memory and cause segmentation faults. 

MWE:
```python
from simsopt.geo import SurfaceRZFourier
NFP = 5
s = SurfaceRZFourier.from_nphi_ntheta(mpol=5, ntor=5, stellsym=True, nfp=NFP)

s.darea_by_dcoeff() # successful evaluation of darea_by_dcoeff

s.change_resolution(7, 7)  # increase the resolution. 

s.darea_by_dcoeff()   # evaluation causes segmentation fault
```

causing the following error on my system:
```
munmap_chunk(): invalid pointer
[laika:07397] *** Process received signal ***
[laika:07397] Signal: Aborted (6)
[laika:07397] Signal code:  (-6)
[laika:07397] [ 0] /usr/lib/libc.so.6(+0x3dcd0) [0x7247d8153cd0]
[laika:07397] [ 1] /usr/lib/libc.so.6(+0x97624) [0x7247d81ad624]
[laika:07397] [ 2] /usr/lib/libc.so.6(gsignal+0x20) [0x7247d8153ba0]
[laika:07397] [ 3] /usr/lib/libc.so.6(abort+0x26) [0x7247d813b582]
[laika:07397] [ 4] /usr/lib/libc.so.6(+0x263bf) [0x7247d813c3bf]
[laika:07397] [ 5] /usr/lib/libc.so.6(+0xa1765) [0x7247d81b7765]
[laika:07397] [ 6] /usr/lib/libc.so.6(+0xa1abc) [0x7247d81b7abc]
[laika:07397] [ 7] /usr/lib/libc.so.6(__libc_free+0x158) [0x7247d81bca08]
[laika:07397] [ 8] /home/smiet/.pyenv/versions/3.11.11/lib/python3.11/site-packages/simsoptpp.cpython-311-x86_64-linux-gnu.so(+0x67a16) [0x7247cc467a16]
[laika:07397] [ 9] /home/smiet/.pyenv/versions/3.11.11/lib/python3.11/site-packages/simsoptpp.cpython-311-x86_64-linux-gnu.so(+0x61d5f) [0x7247cc461d5f]
[...]

PR #519, opened 2025-05-15
Cbs/change resolution fix

smiet:
This adapts the change_resolution to return a copy of the surface, fixes the issue with the cpp backend #478, and aligns it with the return signature of `SurfaceRZPseudospectral.change_resolution`.

### DIFF
--- a/src/simsopt/geo/surfacerzfourier.py
+++ b/src/simsopt/geo/surfacerzfourier.py
@@ -564,51 +564,24 @@ class SurfaceRZFourier(sopp.SurfaceRZFourier, Surface):
             else:
                 kwargs["quadpoints_phi"] = quadpoints_phi
         # create new surface in old resolution
-        surf = SurfaceRZFourier(mpol=self.mpol, ntor=self.ntor, nfp=nfp, stellsym=stellsym,
+        surf = SurfaceRZFourier(mpol=mpol, ntor=ntor, nfp=nfp, stellsym=stellsym,
                                 **kwargs)
-        surf.rc[:, :] = self.rc
-        surf.zs[:, :] = self.zs
-        if not self.stellsym:
-            surf.rs[:, :] = self.rs
-            surf.zc[:, :] = self.zc
-        # set to the requested resolution
-        surf.change_resolution(mpol, ntor)
+        for m in range(0, mpol):
+            for n in range(-ntor, ntor+1):
+                surf.set_rc(m, n, self.get_rc(m, n))
+                surf.set_zs(m, n, self.get_zs(m, n))
+                if self.stellsym:
+                    surf.set_rs(m, n, self.get_rc(m, n))
+                    surf.set_rs(m, n, self.get_rs(m, n))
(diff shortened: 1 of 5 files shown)


tests/geo/test_surface_rzfourier.py
```diff
@@ -522,33 +522,30 @@ def test_change_resolution(self):
...
                 self.assertAlmostEqual(v1, v2)
                 self.assertAlmostEqual(a1, a2)
 
-                s.change_resolution(mpol+1, ntor+1)
-                s.recalculate = True
-                v2 = s.volume()
-                a2 = s.area()
+                s3 = s2.change_resolution(mpol+1, ntor+1)
+                v2 = s3.volume()
+                a2 = s3.area()
```
akaptano, line 548: Can you add a test here that checks if it fixes Issue #478? e.g. tests that s.darea_by_dcoeff() and other surface methods can be called after changing the resolution?

akaptano (commented):
looks great, added some minor comments for better docstrings and unit tests, happy to approve after that.

src/simsopt/geo/surfacerzfourier.py
```diff
@@ -564,51 +564,32 @@ def copy(self, **kwargs):
...
+                this_nmin = 0
+            else: 
+                this_nmin = -this_nmax
+            for n in range(this_nmin, this_nmax+1):
+                surf.set_rc(m, n, self.get_rc(m, n))
+                surf.set_zs(m, n, self.get_zs(m, n))
+                if not surf.stellsym and not self.stellsym:
+                    surf.set_zc(m, n, self.get_zc(m, n))
+                    surf.set_rs(m, n, self.get_rs(m, n))
+
```
mbkumar, line 584: Instead of looping, can't you use index slicing and copy. Could be faster.

tests/mhd/test_integrated_mpi.py
```diff
@@ -167,7 +167,7 @@ def test_Residue_parallelization(self):
                 s = Spec(filename, mpi=mpi)
 
                 # Expand number of Fourier modes to include larger poloidal mode numbers:
-                s.boundary.change_resolution(6, s.boundary.ntor)
+                s.boundary = s.boundary.change_resolution(6, s.boundary.ntor)
                 # To make this example run relatively quickly, we will optimize in a
```
mbkumar, line 171: What happens to the original boundary? Can you print the Optimizable graph and ensure that the old one is  garbage collected (or atleast not used).

smiet:
Should be good to go! Confirmed that an optimizable whose surfaces we replace get garbage collected

(exept for the first instantiated surface, this I do not understand, but that should not prevent us from merging this useful change). 

@mbkumar @akaptano @mishapadidar  have a look (but also please only suggest most neccesary changes; this should be a simple PR addressing a small bug and not cause to overhaul a whole files' docstrings or the like)

mbkumar (approved):
Looks good to me.

PR #521, opened 2025-05-15
Rename calculate_on_axis_B function

akaptano:
As title describes, this pull request merely renames the function in permanent_magnet_helpers.py and makes some other little changes in the comments and docstring to make it clear.

### DIFF
--- a/src/simsopt/util/permanent_magnet_helper_functions.py
+++ b/src/simsopt/util/permanent_magnet_helper_functions.py
@@ -313,9 +313,11 @@ def initialize_coils(config_flag, TEST_DIR, s, out_dir=''):
         # qh needs to be scaled to 0.1 T on-axis magnetic field strength
         from simsopt.mhd.vmec import Vmec
         vmec_file = 'wout_LandremanPaul2021_QH_reactorScale_lowres_reference.nc'
-        total_current = Vmec(TEST_DIR / vmec_file).external_current() / (2 * s.nfp) / 8.75 / 5.69674966667
-        base_curves = create_equally_spaced_curves(ncoils, s.nfp, stellsym=True, R0=R0, R1=R1, order=order, numquadpoints=128)
-        base_currents = [(Current(total_current / ncoils * 1e-5) * 1e5) for _ in range(ncoils-1)]
+        total_current = Vmec(TEST_DIR / vmec_file).external_current() / (2 * s.nfp) / 1.311753
+        print('Total current = ', total_current)
+        base_curves = create_equally_spaced_curves(ncoils, s.nfp, stellsym=True,
+                                                   R0=R0, R1=R1, order=order, numquadpoints=64)
+        base_currents = [(Current(total_current / ncoils * 1e-5) * 1e5) for _ in range(ncoils - 1)]
         total_current = Current(total_current)
         total_current.fix_all()
         base_currents += [total_current - sum(base_currents)]
--- a/src/simsopt/util/permanent_magnet_helper_functions.py
+++ b/src/simsopt/util/permanent_magnet_helper_functions.py
@@ -273,7 +273,8 @@ def make_qfm(s, Bfield):
 def initialize_coils(config_flag, TEST_DIR, s, out_dir=''):
     """
     Initializes coils for each of the target configurations that are
-    used for permanent magnet optimization.
+    used for permanent magnet optimization. The total current is chosen
+    to achieve a desired average field strength along the major radius.
 
     Args:
         config_flag: String denoting the stellarator configuration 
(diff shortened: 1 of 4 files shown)


src/simsopt/util/permanent_magnet_helper_functions.py
```diff
@@ -310,12 +310,14 @@ def initialize_coils(config_flag, TEST_DIR, s, out_dir=''):
...
         from simsopt.mhd.vmec import Vmec
         vmec_file = 'wout_LandremanPaul2021_QH_reactorScale_lowres_reference.nc'
-        total_current = Vmec(TEST_DIR / vmec_file).external_current() / (2 * s.nfp) / 8.75 / 5.69674966667
-        base_curves = create_equally_spaced_curves(ncoils, s.nfp, stellsym=True, R0=R0, R1=R1, order=order, numquadpoints=128)
-        base_currents = [(Current(total_current / ncoils * 1e-5) * 1e5) for _ in range(ncoils-1)]
+        total_current = Vmec(TEST_DIR / vmec_file).external_current() / (2 * s.nfp) / 1.311753
+        print('Total current = ', total_current)
+        base_curves = create_equally_spaced_curves(ncoils, s.nfp, stellsym=True,
+                                                   R0=R0, R1=R1, order=order, numquadpoints=64)
+        base_currents = [(Current(total_current / ncoils * 1e-5) * 1e5) for _ in range(ncoils - 1)]
```
andrewgiuliani, line 320: what's going on here?

src/simsopt/util/permanent_magnet_helper_functions.py
```diff
@@ -310,12 +310,14 @@ def initialize_coils(config_flag, TEST_DIR, s, out_dir=''):
...
         order = 5
 
-        # qh needs to be scaled to 0.1 T on-axis magnetic field strength
+        # QH reactor scale needs to be 5.7 T average magnetic field strength
         from simsopt.mhd.vmec import Vmec
         vmec_file = 'wout_LandremanPaul2021_QH_reactorScale_lowres_reference.nc'
-        total_current = Vmec(TEST_DIR / vmec_file).external_current() / (2 * s.nfp) / 8.75 / 5.69674966667
-        base_curves = create_equally_spaced_curves(ncoils, s.nfp, stellsym=True, R0=R0, R1=R1, order=order, numquadpoints=128)
-        base_currents = [(Current(total_current / ncoils * 1e-5) * 1e5) for _ in range(ncoils-1)]
+        total_current = Vmec(TEST_DIR / vmec_file).external_current() / (2 * s.nfp) / 1.311753
```
andrewgiuliani, line 316: what's going on here? can we add a quick comment? + update the docstring to say what the mean modB on major radius is being set here?

tests/geo/test_pm_grid.py
```diff
@@ -667,13 +667,12 @@ def test_pm_post_processing(self):
         # Make QFM surfaces
         Bfield = bs + b_dipole
         Bfield.set_points(s_plot.gamma().reshape((-1, 3)))
-        #qfm_surf = make_qfm(s_plot, Bfield)
-        #qfm_surf = qfm_surf.surface
+        # qfm_surf = make_qfm(s_plot, Bfield)
```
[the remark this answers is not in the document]
akaptano: I actually think this adds coverage to permanent_magnet_helpers (which is something we discussed with Bharat) so I've added the comments back in (seems to run fine and not take too long in the unit tests)

src/simsopt/util/permanent_magnet_helper_functions.py
akaptano: Removed, and replaced by a fixed number that gets the scaling right. Updated the docstring

akaptano:
> thanks so much for splitting things up, only some really minor comments here.
> 
> It seems like you're not consistently setting up the devices with the same way, e.g.
> 
> for muse, `base_currents[0].fix_all()` is called, while for QH we scale things down and `total_current.fix_all()` is being called, which seems now like all the currents are fixed.
> 
> Can it be clarified why?

Just fixed this, now they both use the 'total_current' method. My impression has been that this is generally superior, because (I think) it is only fixing the total current through all coils (but each coil current can vary within that constraint), rather than `base_currents[0].fix_all()` explicitly fixes the current in coil 1, which might have a crappy initial value that you have now fixed forever.

src/simsopt/util/permanent_magnet_helper_functions.py
```diff
@@ -332,37 +334,36 @@ def initialize_coils(config_flag, TEST_DIR, s, out_dir=''):
...
 
     Args:
-        bs: MagneticField or BiotSavart class object.
-        s: plasma boundary surface.
+        bs (BiotSavart): MagneticField or BiotSavart class object.
+        s (Surface): plasma boundary surface.
 
     Returns:
-        B0avg: Average magnetic field strength along 
+        B0avg (array, shape (s.nphi,)): Average magnetic field strength along
```
landreman, line 366: Isn't B0avg a float and not an array?

src/simsopt/util/permanent_magnet_helper_functions.py
```diff
@@ -332,37 +334,36 @@ def initialize_coils(config_flag, TEST_DIR, s, out_dir=''):
...
-    configuration is scaled correctly.
+    Check the average magnetic field strength along the major radius
+    to make sure the configuration is scaled correctly. For highly shaped
+    stellarators, this can deviate a bit from the on-axis B field strength.
 
     Args:
-        bs: MagneticField or BiotSavart class object.
-        s: plasma boundary surface.
+        bs (BiotSavart): MagneticField or BiotSavart class object.
+        s (Surface): plasma boundary surface.
```
landreman, line 363: Specifically, state that s must be a SurfaceRZFourier - otherwise the `get_rc()` method used in this function would not exist.

akaptano:
@landreman @andrewgiuliani I think this is ready to go but I somehow dismissed Matt's old approval of this. I will merge this in when you're ready.

landreman (approved):

PR #522, opened 2025-05-15
Curveplanar jacobian fix

akaptano:
This PR fixes two issues with the CurvePlanarFourier class:
1. The C++ jacobian calculation was incorrect, and now fixed. This is very stringently tested in upcoming PR with force/torque jacobian checks. 
2. The Python and C++ initializations asked for nfp and stellsym parameters, but these parameters were not used in the actual definition of the curve. So they are removed, and wherever CurvePlanarFourier is tested, the initialization call needs to change from CurvePlanarFourier(numquadpoints, order, nfp, stellsym) to just CurvePlanarFourier(numquadpoints, order)

### DIFF
--- a/src/simsopt/geo/curveplanarfourier.py
+++ b/src/simsopt/geo/curveplanarfourier.py
@@ -1,18 +1,20 @@
 import numpy as np
-
+import jax.numpy as jnp
+from itertools import chain
 import simsoptpp as sopp
-from .curve import Curve
+from .curve import Curve, JaxCurve
 
-__all__ = ['CurvePlanarFourier']
+__all__ = ['CurvePlanarFourier', 'JaxCurvePlanarFourier']
 
 
 class CurvePlanarFourier(sopp.CurvePlanarFourier, Curve):
     r"""
     ``CurvePlanarFourier`` is a curve that is restricted to lie in a plane. The
     shape of the curve within the plane is represented by a Fourier series in
-    polar coordinates. The resulting planar curve is then rotated in three
-    dimensions using a quaternion, and finally a translation is applied. The
-    Fourier series in polar coordinates is
+    polar coordinates centered at the center of curve. 
+    The resulting planar curve is then rotated in three
+    dimensions using a quaternion, and finally a translation is applied by the center point
+    (X, Y, Z). The Fourier series in polar coordinates is
 
     .. math::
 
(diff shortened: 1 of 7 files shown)


landreman (approved):

akaptano:
> I have run into two bugs when testing the CurvePlanarFourier class. I know these are not your bugs, but I figure since we are fixing the bugs now, we should get them all!
> 
> 1. Initializing the class with `order=0` prevents other methods, such as `gamma()` from being called. An `order=0` curve should just be a circle. I think the problem is that [the loop](https://github.com/hiddenSymmetries/simsopt/blob/f30f74bcf6afccda208b16d62cd5b87d2bf73fb8/src/simsoptpp/curveplanarfourier.cpp#L29) in the `gamma` function starts at 1. The `gammadash` functions etc would also need to be corrected. Let me know if you want a second hand on the edits.
> 2. The dof names are not labeled correctly because the class is not initialized with a dof names. The dofs are currently named `x0, x1, x2, ...`.  [The initialization of the class](https://github.com/hiddenSymmetries/simsopt/blob/f30f74bcf6afccda208b16d62cd5b87d2bf73fb8/src/simsopt/geo/curveplanarfourier.py#L58) should use the `names` argument, i.e.
> 
> ```
> Curve.__init__(self, external_dof_setter=CurvePlanarFourier.set_dofs_impl,
>                            x0=self.get_dofs(), names=names)
> ```

Misha can you clarify your issue in (1)? I have not seen an issue with order=0 before and have used this pretty extensively. You are right that in the loop, the gamma function starts at order = 1, but the order = 0 case is handled right before the for loop. Secondly there is already a make_names function for (2) in this branch, fixing this problem.

mishapadidar:
You know, just ignore point 1. I think I made a mistake. Adressind item 2 would be nice though

akaptano:
> You know, just ignore point 1. I think I made a mistake. Adressind item 2 would be nice though

Great, yes this is fixed!

tests/geo/test_curve.py
```diff
@@ -802,5 +802,69 @@ def test_curverzfourier_dofnames(self):
...
+        # Test that the _make_names function returns the correct dof names for a given order
+        order = 3
+        expected_names = [
+            'rc(0)', 'rc(1)', 'rc(2)', 'rc(3)',
+            'rs(1)', 'rs(2)', 'rs(3)',
+            'q0', 'qi', 'qj', 'qk',
+            'X', 'Y', 'Z'
+        ]
+        curve = CurvePlanarFourier(32, order)
+        self.assertEqual(curve._make_names(order), expected_names)
```
mishapadidar, line 815: could we have an error message for this test?

tests/geo/test_curve.py
```diff
@@ -802,5 +802,69 @@ def test_curverzfourier_dofnames(self):
...
+
+        # repeat test with order 0
+        order = 0
+        expected_names = [
+            'rc(0)',
+            'q0', 'qi', 'qj', 'qk',
+            'X', 'Y', 'Z'
+        ]
+        curve = CurvePlanarFourier(32, order)
+        self.assertEqual(curve._make_names(order), expected_names)
```
mishapadidar, line 846: same here?

src/simsopt/geo/curveplanarfourier.py
```diff
@@ -10,9 +9,10 @@ class CurvePlanarFourier(sopp.CurvePlanarFourier, Curve):
     r"""
     ``CurvePlanarFourier`` is a curve that is restricted to lie in a plane. The
     shape of the curve within the plane is represented by a Fourier series in
-    polar coordinates. The resulting planar curve is then rotated in three
-    dimensions using a quaternion, and finally a translation is applied. The
-    Fourier series in polar coordinates is
+    polar coordinates centered at the center of curve. 
+    The resulting planar curve is then rotated in three
+    dimensions using a quaternion, and finally a translation is applied by the center point
+    (X, Y, Z). The Fourier series in polar coordinates is
```
mishapadidar, line 15: nit: use `(X, Y, Z)`

src/simsoptpp/curveplanarfourier.h
```diff
@@ -8,7 +8,7 @@ class CurvePlanarFourier : public Curve<Array> {
         CurvePlanarFourier is a curve that is restricted to lie in a plane. In
         the plane, the curve is represented using a Fourier series in plane polar coordinates:
 
-           r(phi) = \sum_{n=0}^{order} x_{c,n}cos(n*nfp*phi) + \sum_{n=1}^order x_{s,n}sin(n*nfp*phi)
+           r(phi) = \sum_{n=0}^{order} x_{c,n}cos(n*phi) + \sum_{n=1}^order x_{s,n}sin(n*phi)
```
mishapadidar, line 11: should we make this notation match the docstring, i.e. `r(\phi) = \sum_{m=0}^{\text{order}} r_{c,m}\cos(m \phi) + \sum_{m=1}^{\text{order}} r_{s,m}\sin(m \phi).`

mishapadidar (commented):
looking pretty good. Just minor comments, then i think its ready to go.

src/simsopt/geo/curveplanarfourier.py
akaptano: What do you mean? (X, Y, Z) is what is says above

akaptano:
Last call -- is this ready to merge? I will merge the current version of master in right now to ease things.

landreman (approved):

PR #530, opened 2025-06-02
ToroidalTransitStoppingCriterion: changed max_transits to double

landreman:
I've sometimes wanted to follow field lines for a number of toroidal transits that is not an integer, for field lines that are periodic in the (R,Z) plane after a number of field periods different from nfp. This isn't possible now with ToroidalTransitStoppingCriterion because the max_transits argument has type int. In this tiny PR I just change the type to double to enable this functionality.

Users who have been using the argument type `int` should notice no differences. For test coverage, I just randomly picked two of the tests that used ToroidalTransitStoppingCriterion and changed the argument from int to float. In other tests the argument remains an int, so both types are covered.


smiet (approved):
Looks good! Thanks for this contribution, I was wanting the same thing!

PR #532, opened 2025-06-20
Refactor: unify stellarator configuration loaders into a single get_data API

armulrich:
This change consolidates all of the individual get_*_data loaders into a single get_data function, streamlining the API and mirroring the style used in the [pyQSC repository](https://github.com/landreman/pyQSC/blob/cd75359ea47548d5db7ccb458c100085c04ba1bc/qsc/configurations.py#L9). It updates the test suite to call get_data instead of the old functions, preserves backward compatibility via thin wrappers, and refreshes the module’s documentation so that all configurations (“ncsx,” “hsx,” “giuliani,” and “w7x”) are described consistently in one place. It was raised by @landreman  and @mishapadidar.

### DIFF
--- a/tests/geo/test_boozersurface.py
+++ b/tests/geo/test_boozersurface.py
@@ -250,7 +250,8 @@ class BoozerSurfaceTests(unittest.TestCase):
                 c.rotmatT = c.rotmat.T
             coils = coils_via_symmetries(base_curves + curves_flipped,
                                          base_currents + currents_flipped, nfp, False)
-        current_sum = nfp * sum(abs(c.get_value()) for c in base_currents)
+        
+        current_sum = sum(abs(c.current.get_value()) for c in coils)
 
         bs = BiotSavart(coils)
 


tests/geo/test_surface_objectives.py
```diff
@@ -64,7 +64,7 @@ def test_toroidal_flux_is_constant(self):
         of the cross section (varphi = constant) across which it is computed
         """
         s = get_exact_surface()
-        curves, currents, ma = get_ncsx_data()
+        curves, currents, ma, _ = get_data("ncsx")
         nfp = 3
         coils = coils_via_symmetries(curves, currents, nfp, True)
         bs_tf = BiotSavart(coils)
```
landreman, line 70: Lines 68-70 could be eliminated now because the BiotSavart is returned directly by get_data(). This is also true of many of the other places that get_data is called.

tests/geo/test_qfm.py
```diff
@@ -49,13 +49,14 @@ def test_qfm_objective_gradient(self):
...
+                        
 
-    def subtest_qfm_objective_gradient(self, surfacetype, stellsym, get_data):
+    def subtest_qfm_objective_gradient(self, surfacetype, stellsym, config):
         np.random.seed(1)
-        curves, currents, ma = get_data()
+        curves, currents, ma, _ = get_data(config)
         nfp = ma.nfp
         coils = coils_via_symmetries(curves, currents, nfp, True)
         bs = BiotSavart(coils)
```
landreman, line 62: I believe lines 60-62 can be eliminated now since get_data returns a BiotSavart object. I believe line 63 could be replaced with `bs_tf = BiotSavart(bs.coils)` or perhaps some kind of copy() could be used.

landreman (commented):
Looks great! Thanks for working on this.

Can we also add `simsopt.configs.LHD_like` to the same get_config() function?

It might be nice to return nfp for each config as well. (I see in the tests that a hack for this is to get the nfp attribute of the magnetic axis, but this trick isn't obvious). Would it make sense to include nfp as another returned quantity?

armulrich:
> Looks great! Thanks for working on this.
> 
> Can we also add `simsopt.configs.LHD_like` to the same get_config() function?
> 
> It might be nice to return nfp for each config as well. (I see in the tests that a hack for this is to get the nfp attribute of the magnetic axis, but this trick isn't obvious). Would it make sense to include nfp as another returned quantity?

No problem @landreman ! Glad to hear it's going in the right direction.

And yes, I think adding `simsopt.configs.LHD_like` makes a lot of sense and helps clean up the configuration interface further. I'll suggest removing the standalone `get_LHD_like_data` function and instead integrating it directly into `zoo.py`, registering it as `"lhd_like"`. Also, just to note: I called the general function `get_data` rather than `get_config`. To me, it felt a bit clearer and flows better syntactically (eg. `get_data("ncsx")`). Let me know if you think there's a strong reason to prefer `get_config` instead.

As for returning `nfp`, I completely agree, it avoids the need to unwrap it from `ma.nfp`, which isn’t always obvious to users. I’d propose returning it explicitly like this from `get_data`:

```python
return curves, currents, ma, nfp, bs
```

Note: I placed `nfp` before `bs` since it conceptually follows from the magnetic axis definition, hope you agree with that.

tests/geo/test_surface_objectives.py
armulrich: I agree. Will address this.

tests/geo/test_qfm.py
armulrich: Yep, agree with this too.

armulrich:
Addressed all of the above comments in this commit.

smiet:
Some people have scripts that use on these functions. 

- Would it be an idea to keep the old functions (until the next major release) and add a deprecationwarning, similar to what is done in NumPy?

- We are also working on a QUASR loader, that automatically downloads any QUASR id. This has a different return style (QUASR does not save the axis and so cannot return it) Does it still make sense to consolidate this into a unified configuration-getting funciton? @andrewgiuliani @landreman

tests/geo/test_qfm.py
armulrich: Finished implementing this too.

src/simsopt/configs/zoo.py
```diff
@@ -2,188 +2,334 @@
...
 from pathlib import Path
 THIS_DIR = (Path(__file__).parent).resolve()
 
-__all__ = ['get_ncsx_data', 'get_hsx_data', 'get_giuliani_data', "get_w7x_data"]
+__all__ = ["get_data", "configurations"]
+configurations = ["ncsx", "hsx", "giuliani", "w7x", "lhd_like"]
 
-
-def get_ncsx_data(Nt_coils=25, Nt_ma=10, ppp=10):
+def get_data(name, **kwargs):
```
mishapadidar, line 17: what do we think about the name `get_data`? should we rename to something like `load_config`?

Similarly, we could change `name` to something more descriptive like `configuration` or `design`.

src/simsopt/configs/zoo.py
```diff
@@ -2,188 +2,334 @@
...
-    Args:
-        Nt_coils: order of the curves representing the coils.
-        Nt_ma: order of the curve representing the magnetic axis.
-        ppp: point-per-period: number of quadrature points per period
+    Parameters
+    ----------
+    name : str
+        Which configuration to load. Available values are:
+        ``"ncsx"``, ``"hsx"``, ``"giuliani"``, ``"w7x"``, ``"lhd_like"``.
+    Nt_coils : int, optional
```
mishapadidar, line 27: eventually we should consider renaming this arguments. `Nt_coils` and `ppp` are not great names.

src/simsopt/configs/zoo.py
```diff
@@ -2,188 +2,334 @@
...
+        Points-per-period: number of quadrature points per period.
+    length : int, optional
+        (*Giuliani only*) Total length for the nine‐stage optimization.
+    nsurfaces : int, optional
+        (*Giuliani only*) Number of surfaces used in the optimization.
+    numquadpoints_circular : int, optional
+        (*LHD_like only*) Number of quadrature points for the six circular coils.
+    numquadpoints_helical : int, optional
+        (*LHD_like only*) Number of quadrature points for each helical coil.
+    numquadpoints_axis : int, optional
```
mishapadidar, line 41: I could see this list of arguments becoming quite long as we add more configurations with their own set of parameters.

src/simsopt/configs/zoo.py
```diff
@@ -2,188 +2,334 @@
...
-    return (curves, currents, ma)
-
-
-def get_hsx_data(Nt_coils=16, Nt_ma=10, ppp=10):
-    """
-    Get a configuration that corresponds to the modular coils of the HSX experiment.
+    Returns
+    -------
+    tuple
+        *5-element tuple* ``(curves, currents, ma, nfp, bs)``, where:
```
mishapadidar, line 47: curves and currents should already be in the `bs` object right? perhaps we can remove them from the tuple. `nfp` should also be in the `bs` or `ma` right?

src/simsopt/configs/zoo.py
```diff
@@ -2,188 +2,334 @@
...
+        ma : :class:`CurveRZFourier`
+            The magnetic axis, of order ``Nt_ma``, with ``nfp`` field periods.
+        nfp : int
+            Number of field periods.
+        bs : :class:`BiotSavart`
+            The Biot–Savart operator built from the coil objects.  
+            
+                - For all configurations **except** ``"lhd_like"``, the coils are expanded via ``coils_via_symmetries(curves, currents, nfp, True)`` before building ``bs``.
+                
+                - For ``"lhd_like"``, each ``CurveXYZFourier`` and its ``Current`` are passed directly into ``Coil(curve, current)`` (no symmetry expansion).
```
mishapadidar, line 62: Perhaps it is worth separating the documentation a bit, so that any notes specific to configuration arguments or output are listed alongside the description of the configuration. what do you think?

src/simsopt/configs/zoo.py
```diff
@@ -2,188 +2,334 @@
...
-    ma = CurveRZFourier(numpoints, Nt_ma, nfp, True)
-    ma.rc[:] = cR[0:(Nt_ma+1)]
-    ma.zs[:] = sZ[0:Nt_ma]
-    ma.x = ma.get_dofs()
-    return (curves, currents, ma)
-
-
-def get_giuliani_data(Nt_coils=16, Nt_ma=10, ppp=10, length=18, nsurfaces=5):
-    """
+    Notes
```
mishapadidar, line 64: I think a good example to follow in the documentation is the [scipy minimize](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize.html). For scipy minimize, the `method` argument has many options, similar how we have many options for `name`. You'll notice that for each method, there is a link to a separate documentation page discussing the specific arguments and quirks of using scipy minimize with that method. For example, the Nelder Mead optimizer has its own [page](https://docs.scipy.org/doc/scipy/reference/optimize.minimize-neldermead.html#optimize-minimize-neldermead). 

We could try to reproduce this here. For each configuration we have a link to a separate documentation page that discusses the specific arguments, and the modifications to the output tuple.

armulrich: That sounds like a solid approach. We can keep the high-level overview and shared parameters in the main get_data docstring, then link out to individual pages, one per config, where we dive into the nitty gritty params and return.

src/simsopt/configs/zoo.py
armulrich: I’m fine with get_data, it’s short and to the point, but ``load_config`` would perhaps make the intent a bit clearer. Happy to switch if that feels more intuitive to everyone. @landreman perhaps you have a take on this.

src/simsopt/configs/zoo.py
mishapadidar: we should change the naming now so we dont get locked in to a bad convention.

src/simsopt/configs/zoo.py
mishapadidar: The kwargs handles this in a clean way in the code. The only problem is the documentation getting messy. One way to handle this is by setting up the documentation to be like the scipy minimize. If this is not possible, then another way is to just add sections in the documentation devoted to each config. Something like, 
```
Parameters
    ----------
    name : str
        Which configuration to load. Available values are:
        ``"ncsx"``, ``"hsx"``, ``"giuliani"``, ``"w7x"``, 
    kwargs: dict
        Configuration specific parameters. See documentation of these parameters below.


ncsx
    ----------
    blah : discussion of parameters.
```

This might be overkill for now though, since we only have a small number of configurations.

src/simsopt/configs/zoo.py
mishapadidar: hmm, okay. maybe we leave `nfp` then? I am still in favor of removing curves and currents. Lets get more opinions @andrewgiuliani @landreman

src/simsopt/configs/zoo.py
mishapadidar: Question: for `lhd_like` and all other configurations, do we always return the full set of coils? LHD is the only configuration with helical coils so perhaps the coils already represent the full coil set. If that is so, then we can remove this comment on the case-by-case behavior and just say `we return a biot savart for all coils`. If this is not the case, then maybe we should modify the behavior of `lhd_like` to return the full set of coils.

src/simsopt/configs/zoo.py
mishapadidar: If this is too difficult we can drop it.

armulrich: Ok, got it.

src/simsopt/configs/zoo.py
armulrich: Yeah, kwargs keeps the code tidy. The mess is purely in the docs. I’m fine slimming the main “Parameters” block to kwargs: dict and then listing config‑specific options in short subsections (“ncsx”, “hsx”, …) right below, like you suggest. If we later spin up separate pages a la scipy.optimize.minimize, we can move those details there and just link out. For now I’ll probaly do the lightweight subsection approach then...

armulrich:
@smiet Addressed all of the following in the latest push.  
> Looks like there are a few more changes to do before we can move forward:
> 
> * [x]  re-instate the old `get_<name>_data` scripts with a `DeprecationWarning`
> * [x]  change the naming of `ppp` and `Nt_X` (as per discussion^: [Refactor: unify stellarator configuration loaders into a single get_data API #532 (comment)](https://github.com/hiddenSymmetries/simsopt/pull/532#discussion_r2222969306))
> * [x]  Resolve the merge conflict with the updated master.
> 
> [When this PR is merged, I will adapt #425 to use this function.]


Extra notes:
- Restored old `get_*_data` functions as well as `get_LHD_like`, with deprecation warnings added to the docstrings for clarity.
- Updated `get_data` parameter naming conventions to improve readability.
Screenshot 2025-08-09 at 14 28 04
Screenshot 2025-08-09 at 14 30 33

Screenshot 2025-08-09 at 14 30 43

Screenshot 2025-08-09 at 14 28 47

Screenshot 2025-08-09 at 14 28 57

src/simsopt/configs/zoo.py
armulrich: @mishapadidar No further concerns were raised, so I’ve kept it as `get_data` for now.

src/simsopt/configs/zoo.py
armulrich: Addressed this in the latest push.

src/simsopt/configs/zoo.py
armulrich: Resolved by slimming the main `Parameters` block to `kwargs: dict` and moving config-specific arguments into short subsections for each configuration.

src/simsopt/configs/zoo.py
armulrich: Leaving tuple contents unchanged for now to preserve the old API and maintain convenience access to `curves` and `currents`. `nfp` is still included since it is not exposed directly by `bs` or `ma` (and definitions differ for `lhd_like`). No changes made in the last push. Can revisit if consensus shifts toward removing `curves`/`currents`. @mishapadidar @landreman @andrewgiuliani

smiet: I am happy with the current return tuple, but I think the description and naming should be changed to `base_curves`, `base_currents` to emphasize that this returns only the unique coils. Providing these is useful because it is not straighforward to extract these from the bs, which contains ss and nfp copies of them.

tests/configs/test_LHD_like.py
```diff
@@ -12,12 +12,11 @@ def test_axis(self):
...
-        coils, currents, axis = get_LHD_like_data()
+        curves, currents, axis, nfp, bs = get_data("lhd_like")
         # Flip the sign of current so B points towards +phi. Otherwise
         # fieldline_tracing traces towards -phi.
-        currents = -np.array(currents)
         coils = [
-            Coil(curve, Current(current)) for curve, current in zip(coils, currents)
+            Coil(curve, -1*current) for curve, current in zip(curves, currents)
         ]
         field = BiotSavart(coils)
```
smiet, line 21: use `bs` returned from get_data.

tests/field/test_coilset.py
```diff
@@ -95,10 +93,9 @@ def test_base_coils(self):
 
     def test_base_coils_setter(self):
         # Test the base_coils setter method
-        curves, currents, _ = get_ncsx_data()
-        new_base_coils = [Coil(curve, current) for curve, current in zip(curves, currents)]
-        self.coilset.base_coils = new_base_coils
-        self.assertEqual(self.coilset.base_coils, new_base_coils)
+        curves, currents, ma, nfp, bs = get_data("ncsx")
```
smiet, line 96: A CoilSet should be provided with base_coils (the unique coils). This passes but uses the setter incorrectly. 

replace with
```
        curves, currents, ma, nfp, bs = get_data("ncsx")
        new_base_coils = [Coil(curve, current) for curve, current in zip(curves, currents)]
        self.coilset.base_coils = new_base_coils
        self.assertEqual(self.coilset.base_coils, new_base_coils)
        ```

File: tests/geo/surface_test_helpers.py [hunk 11]
```diff
@@ -84,10 +84,9 @@ def get_boozer_surface(label="Volume", nphi=None, ntheta=None, boozer_type='exac
     if boozer_type == 'exact':
         assert weight_inv_modB == False
 
-    base_curves, base_currents, ma = get_ncsx_data()
-    coils = coils_via_symmetries(base_curves, base_currents, 3, True)
-    bs = BiotSavart(coils)
-    current_sum = sum(abs(c.current.get_value()) for c in coils)
+    base_curves, base_currents, ma, nfp, bs = get_data("ncsx")
+    
+    current_sum = sum(abs(c.current.get_value()) for c in bs.coils)
```
smiet, line 89: use returned currents * nfp instead of all coils in bs

tests/geo/surface_test_helpers.py
```diff
@@ -116,7 +114,7 @@ def get_boozer_surface(label="Volume", nphi=None, ntheta=None, boozer_type='exac
         lab = Volume(s, nphi=nphi, ntheta=ntheta)
         lab_target = lab.J()
     elif label == "ToroidalFlux":
-        bs_tf = BiotSavart(coils)
+        bs_tf = BiotSavart(bs.coils)
```
smiet, line 117: just use `bs` object?

tests/geo/test_boozersurface.py
```diff
@@ -24,10 +24,8 @@ def test_residual(self):
         """
 
         s = get_exact_surface()
-        curves, currents, ma = get_ncsx_data()
-        coils = coils_via_symmetries(curves, currents, 3, True)
-        bs = BiotSavart(coils)
-        bs_tf = BiotSavart(coils)
+        curves, currents, ma, nfp, bs = get_data("ncsx")
+        bs_tf = BiotSavart(bs.coils)
```
smiet, line 28: use bs object

tests/geo/test_boozersurface.py
```diff
@@ -82,11 +80,9 @@ def test_boozer_penalty_constraints_hessian(self):
...
                                                     optimize_G=False, vectorize=False):
         np.random.seed(1)
-        curves, currents, ma = get_ncsx_data()
-        coils = coils_via_symmetries(curves, currents, 3, True)
-        bs = BiotSavart(coils)
-        bs_tf = BiotSavart(coils)
-        current_sum = sum(abs(c.current.get_value()) for c in coils)
+        curves, currents, ma, nfp, bs = get_data("ncsx")
+        bs_tf = BiotSavart(bs.coils)
+        current_sum = sum(abs(c.current.get_value()) for c in bs.coils)
```
smiet, line 85: only unique currents

tests/geo/test_boozersurface.py
```diff
@@ -395,12 +387,10 @@ def subtest_convergence_cpp_and_notcpp_same(self, vectorize):
         """
         compute a surface using either the vectorized or non-vectorized subroutines
         """
-        curves, currents, ma = get_ncsx_data()
-        coils = coils_via_symmetries(curves, currents, ma.nfp, True)
-        current_sum = sum(abs(c.current.get_value()) for c in coils)
-        bs = BiotSavart(coils)
+        curves, currents, ma, nfp, bs = get_data("ncsx")
+        current_sum = sum(abs(c.current.get_value()) for c in bs.coils)
```
smiet, line 391: only unique currents

tests/geo/test_boozersurface.py
```diff
@@ -447,11 +437,9 @@ def test_boozer_penalty_constraints_cpp_notcpp(self):
     def subtest_boozer_penalty_constraints_cpp_notcpp(self, surfacetype, stellsym, optimize_G, nphi, ntheta, weight_inv_modB, mpol, ntor):
 
         np.random.seed(1)
-        curves, currents, ma = get_ncsx_data()
-        coils = coils_via_symmetries(curves, currents, 3, stellsym)
-        bs = BiotSavart(coils)
-        bs_tf = BiotSavart(coils)
-        current_sum = sum(abs(c.current.get_value()) for c in coils)
+        curves, currents, ma, nfp, bs = get_data("ncsx")
+        bs_tf = BiotSavart(bs.coils)
```
smiet, line 441: use returned bs and only unique currents

tests/geo/test_qfm.py
```diff
@@ -25,11 +25,8 @@ def test_residual(self):
         """
 
         s = get_exact_surface()
-        curves, currents, ma = get_ncsx_data()
-        nfp = 3
-        coils = coils_via_symmetries(curves, currents, nfp, True)
-        bs = BiotSavart(coils)
-        bs_tf = BiotSavart(coils)
+        curves, currents, ma, nfp, bs = get_data("ncsx")
+        bs_tf = BiotSavart(bs.coils)
```
smiet, line 29: same bs

tests/geo/test_qfm.py
```diff
@@ -49,17 +46,15 @@ def test_qfm_objective_gradient(self):
...
-    def subtest_qfm_objective_gradient(self, surfacetype, stellsym, get_data):
+    def subtest_qfm_objective_gradient(self, surfacetype, stellsym, config):
         np.random.seed(1)
-        curves, currents, ma = get_data()
-        nfp = ma.nfp
-        coils = coils_via_symmetries(curves, currents, nfp, True)
-        bs = BiotSavart(coils)
-        bs_tf = BiotSavart(coils)
+        curves, currents, ma, nfp, bs = get_data(config)
+        bs_tf = BiotSavart(bs.coils)
```
smiet, line 57: same bs

tests/geo/test_qfm.py
```diff
@@ -100,11 +95,8 @@ def test_qfm_label_constraint_gradient(self):
 
     def subtest_qfm_label_constraint_gradient(self, surfacetype, stellsym):
         np.random.seed(1)
-        curves, currents, ma = get_ncsx_data()
-        nfp = 3
-        coils = coils_via_symmetries(curves, currents, nfp, True)
-        bs = BiotSavart(coils)
-        bs_tf = BiotSavart(coils)
+        curves, currents, ma, nfp, bs = get_data("ncsx")
+        bs_tf = BiotSavart(bs.coils)
```
smiet, line 99: same bs

tests/geo/test_qfm.py
```diff
@@ -140,17 +132,15 @@ def test_qfm_penalty_constraints_gradient(self):
...
-    def subtest_qfm_penalty_constraints_gradient(self, surfacetype, stellsym, get_data):
+    def subtest_qfm_penalty_constraints_gradient(self, surfacetype, stellsym, config):
         np.random.seed(1)
-        curves, currents, ma = get_data()
-        nfp = ma.nfp
-        coils = coils_via_symmetries(curves, currents, nfp, True)
-        bs = BiotSavart(coils)
-        bs_tf = BiotSavart(coils)
+        curves, currents, ma, nfp, bs = get_data(config)
+        bs_tf = BiotSavart(bs.coils)
```
smiet, line 143: same

tests/geo/test_surface_objectives.py
```diff
@@ -64,10 +63,8 @@ def test_toroidal_flux_is_constant(self):
         of the cross section (varphi = constant) across which it is computed
         """
         s = get_exact_surface()
-        curves, currents, ma = get_ncsx_data()
-        nfp = 3
-        coils = coils_via_symmetries(curves, currents, nfp, True)
-        bs_tf = BiotSavart(coils)
+        curves, currents, ma, nfp, bs= get_data("ncsx")
+        bs_tf = BiotSavart(bs.coils)
```
smiet, line 67: same bs

smiet (changes requested):
Thank you for the work, it is looking really good! sorry for the delay in my review (vacation). 

There a few more things: 
-The `curves` and `currents` returned by `get_data` are only the unique coils, whereas `bs.coils` contains the symmetry-copies. can you make this clear by calling these `base_curves` and `base_currents` in documentation and in examples (if you are a regex ninja also in tests)
- you often create a new BiotSavart object with `bs_tf= BiotSavart(bs.coils)` these two should be identical, and it is simpler and clearer to use the `bs` that is returned by `get_data`
- The `current_sum` object that you create by summing all currents in `bs.coils` will sum the same value many times. Best practice is to only take the unique values and sum them (multiplied by 2*nfp if the absolute value is important).

examples/1_Simple/qfm.py
```diff
@@ -17,15 +17,12 @@
 print("Running 1_Simple/qfm.py")
 print("=======================")
 
-curves, currents, ma = get_ncsx_data()
-coils = coils_via_symmetries(curves, currents, 3, True)
-bs = BiotSavart(coils)
-bs_tf = BiotSavart(coils)
+curves, currents, ma, nfp, bs = get_data("ncsx")
+bs_tf = BiotSavart(bs.coils)
```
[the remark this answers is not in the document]
armulrich: NOTE: this impacts all the subsequent use same ``bs_tf = bs`` comments

armulrich:
Thanks again for the thorough feedback @smiet.  
I updated the code to consistently use `base_curves`, `base_currents`, and `all_curves`, and fixed all instances of `current_sum` so that only unique currents are counted.  

The two items I haven’t implemented (due to concerns) are:  
1. **LHD-like config:** I didn’t switch to using the returned `bs` object, since `get_data` currently builds `bs` with positive currents. Flipping the sign afterward gives a different `bs` than the one used in the test, which causes it to fail. Hence need to build separate ``bs`` object.
2. **`bs_tf` vs single `bs`:** Throughout the `get_data` usage kept the second `BiotSavart` object (`bs_tf`). A single `bs` isn’t reliable because `BiotSavart.set_points()` is mutable. `QfmResidual` and `ToroidalFlux` require different grids at the same time, and reusing one `bs` causes reshape errors. Having a duplicate `BiotSavart` avoids this conflict while still referencing the same coil set.  

If you’d like me to handle either of these two points differently, let me know.

examples/2_Intermediate/boozer.py
```diff
@@ -19,18 +19,15 @@
...
 print("================================")
 
-curves, currents, ma = get_ncsx_data()
-coils = coils_via_symmetries(curves, currents, 3, True)
-curves = [c.curve for c in coils]
-bs = BiotSavart(coils)
-bs_tf = BiotSavart(coils)
-current_sum = sum(abs(c.current.get_value()) for c in coils)
+curves, currents, ma, nfp, bs = get_data("ncsx")
+curves = [c.curve for c in bs.coils]
```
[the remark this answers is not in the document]
armulrich: implemented this below with comment

examples/2_Intermediate/boozer.py
```diff
@@ -19,18 +19,15 @@
...
+bs_tf = BiotSavart(bs.coils)
+current_sum = sum(abs(c.current.get_value()) for c in bs.coils)
```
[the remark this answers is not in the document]
armulrich: fixed below

tests/configs/test_LHD_like.py
armulrich: My concern with this outlined below (minus phi is a special case)

tests/field/test_coilset.py
armulrich: implemented this too.

tests/geo/surface_test_helpers.py
armulrich: same issue as pointed out already

tests/geo/test_boozersurface.py
armulrich: done

tests/geo/test_boozersurface.py
```diff
@@ -122,11 +118,9 @@ def subtest_boozer_penalty_constraints_gradient(self, surfacetype, stellsym,
     def subtest_boozer_penalty_constraints_hessian(self, surfacetype, stellsym,
                                                    optimize_G=False, vectorize=False):
         np.random.seed(1)
-        curves, currents, ma = get_ncsx_data()
-        coils = coils_via_symmetries(curves, currents, 3, True)
-        bs = BiotSavart(coils)
-        bs_tf = BiotSavart(coils)
-        current_sum = sum(abs(c.current.get_value()) for c in coils)
+        curves, currents, ma, nfp, bs = get_data("ncsx")
+        bs_tf = BiotSavart(bs.coils)
```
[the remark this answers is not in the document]
armulrich: implemented partly with current sum fixed but kept `bs_tf` the same due to issue outlined below

tests/geo/test_boozersurface.py
armulrich: done

tests/geo/test_qfm.py
armulrich: same as above

tests/geo/test_qfm.py
armulrich: `tf` issue outlined below

PR #538, opened 2025-07-21
Update the python version and singularity version

mbkumar:
Python 3.9 is reach EOL. Changing python version from 3.9 to 3.13 for future proofing. Also changing from singularity to apptainer for singularity builds.


mbkumar:
PyOculus had an exact pin on the numpy version to 2.0.0. This is causing the failures. Waiting for Zhisong to merge my PR on PyOculus.

PR #543, opened 2025-08-04
Bug in Surface Cross Section arguments

mishapadidar:
This PR makes two fixes to the `Surface.cross_section` function. 
1. The documentation was not compiling correctly, which made it difficult to understand what the arguments and returns were. The documentation has now been corrected. See attached.
2. The `thetas` argument can be one of `int, Array, None`.  However, when an array or list was passed in, an error would be thrown. The error was arising because of improper validation of inputs: there was no handling for the case when `thetas` is an array. In this PR, we correct the validation of the inputs to allow for `thetas` to be an array, as the documentation states it can be.

**Changes**
- Small changes have been made to the `cross_section` method and documentation.
- A basic unit test has been added to check that an array can be used as input.

**New Documenation Image** 
Screenshot 2025-08-04 at 3 18 47 PM

**Old Documentation Image**
Screenshot 2025-08-04 at 3 15 02 PM

### DIFF
--- a/src/simsopt/geo/surface.py
+++ b/src/simsopt/geo/surface.py
@@ -326,40 +326,37 @@ class Surface(Optimizable):
 
     def cross_section(self, phi, thetas=None, tol=1e-13):
         """
-        Computes the cross-section at an angle :math:`\phi` at `thetas` using bisection.
-        :math:`\phi'` follows the same conventions as `Surfaces`, i.e. :math:`\phi=0, 1` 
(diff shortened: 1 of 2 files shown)


andrewgiuliani (commented):
there's one line missing by code coverage, and it's an easy one to get covered.  Otherwise, thanks for fixing this up!

mishapadidar:
> there's one line missing by code coverage, and it's an easy one to get covered. Otherwise, thanks for fixing this up!

Updated the coverage of that line!

andrewgiuliani (approved):

PR #556, opened 2025-09-24
Update optimization examples for clearer and faster optimizations

rogeriojorge:
Newcomers to SIMSOPT might have a hard time understanding VMEC-based optimization based only on QH_fixed_resolution.py and others. There are a few more modes than most computers can optimize on a small amount of time and it prints all VMEC calculations. This streamlines it to make it easier to start doing their own optimizations.


smiet (approved):
LGTM! clearer examples are nice. 

One suggestion, change to 4 MPI threads, most users will not have a system with 9 cores.  Leaving as-is also fine with me.

PR #557, opened 2025-09-24
Overflow error in Mgrid write method

mishapadidar:
**Summary**
Using the `.write` method of the `Mgrid` class can lead to an Overflow error (see screenshot). 

**Context**
The `Mgrid` class is used to write `netcdf` files which can be used in free boundary VMEC. When setting the resolution parameters of the `Mgrid` class, `nr, nphi, nz`, to high values, the `write` method can overflow (see screenshot)

Image

As seen in the screenshot, the write method attempts to write values to an array that is allocated for _single precision_. The error occurs because the values being written cannot be represented in 32 bits.

This overflow error is problematic because it limits users to only using coarse grids which poorly represent magnetic fields.

**Fix**
The `Mgrid` class uses Scipy's `netcdf_file` class to write the file. The `netcdf_file` class relies on the `version` argument,  which defaults to `version=1`. The `version` argument sets the array format for writing, where 1 means Classic format (32 bit in our case) and 2 means 64-bit offset format. Switching the `Mgrid.write` method to rely on `netcdf_file(... version=2)` forces the values to be written to a 64-bit array, which solves the problem. 

**Question/Uncertainty**
While setting `version=2` seems to prevent the overflow error from appearing, I do not understand the fortran side of the code well enough to know if this change will lead to misrepresentation of the magnetic fields in the Mgrid files. In other words, is using `64-bit offset format` appropriate? Perhaps you have a better sense, @mbkumar @landreman.

### DIFF
--- a/src/simsopt/field/mgrid.py
+++ b/src/simsopt/field/mgrid.py
@@ -143,7 +143,7 @@ class MGrid():
             filename (str): output file name.
         '''
 
-        with netcdf_file(filename, 'w', mmap=False) as ds:
+        with netcdf_file(filename, 'w', mmap=False, version=2) as ds:
 
             # set netcdf dimensions
             ds.createDimension('stringsize', 30)
@@ -225,7 +225,7 @@ class MGrid():
             MGrid: ``MGrid`` object with data from file.
         '''
 
-        with netcdf_file(filename, 'r', mmap=False) as f:
+        with netcdf_file(filename, 'r', mmap=False, version=2) as f:
 
             # load grid
             nr = f.variables['ir'].getValue()


smiet:
I see that `simsopt` only uses the `MGrid` class in one place, in `magneticfield.py` in the `to_mgrid` function, i.e. for generating the files which will be inputs for VMEC to run on. 

The question is thus whether VMEC can handle (read in) mgrid files which have been generated with this option (and next, whether our bound version of VMEC can). This I am pretty sure of,  I know people who use insanely large MGRID resolutions. 

The input file handling of our bound VMEC should be similar to raw VMEC, as it just wraps and interfaces the Fortran routines. 

for safety, you should test if the mgrid file generated like this works with raw VMEC and on our bound version, and if it passes we can be confident there is no issue. If it fails on raw VMEC, it is an issue for them I would say.

mishapadidar:
@mbkumar I believe that the problem is that the mgrid file is being written with the incorrect version argument. I am not 100% positive if the error is arising from the float type. 

One reason to suspect this is that [DESC also writes mgrid files](https://github.com/PlasmaControl/DESC/blob/4e527932c898b6058a4134fe7e57157e37e80b78/desc/magnetic_fields/_core.py#L642) using the 64-bit offset format. 

Another is that the simsopt `.write` method cannot write files as large as the ones often used in the community, such as the ones [here](https://www.osti.gov/servlets/purl/2458165), unless we use the 64 bit offset format. 

To get a better understanding of the problem, I am consulting with the DESC developers.

mishapadidar:
**Establishing file format correctness**
I believe I have been able to establish that community has been using NETCDF files with the `64-bit offset` format by using the [ncdump](https://linux.die.net/man/1/ncdump) command. According to the `ncdump` documentation, "ncdump may also be used to determine what kind of netCDF file is used (which variant of the netCDF file format) with the -k option."

I ran the command

    ncdump -k some_file.nc

using `wout` files for W7-X, HSX, and ARIES-CS, as well as an [mgrid file from STELLOPT](https://github.com/PrincetonUniversity/STELLOPT/blob/c2ff2f1d491b44ec3e3a97ad11ed3df678e1d848/BENCHMARKS/STELLOPT_TEST/STELLCOPT/mgrid_RUN09.00000.nc). They all return the same answer:`64-bit offset`. 

This is a strong signal to suggest that we should be using the `version=2` argument in the NETCDF writer class, i.e. `netcdf_file(... version=2)`. Using the `version=1` argument, will not write a file in `64-bit offset` format.

**Test runs**
Following @smiet's recommendation, I tested that VMEC in free boundary mode can handle large mgrid files: `nphi=720, nR=nZ=544`. Here are the takeaways:
- The `Mgrid.write` operation is successful when we use the `netcdf_file(... version=2)`, as suggested in this PR. This is not the case if we use `netcdf_file(... version=1)`.
- VMEC runs and converges to `ftol=1.00E-15`
- The attached image compares the rotational transform profile produced by free boundary VMEC compared to the profile produced using BoozerSurface. They agree extremely well, suggesting that the results are accurate.

Screenshot 2025-11-25 at 1 46 09 PM

**Discussion with DESC group**
Additionally, I discussed this code change with the DESC group (Daniel Dudt and Dario Panici), since they also use `netcdf_file(... version=2)`. They couldn't recall why they also use this setting, but believe that it is the right choice.

PR #558, opened 2025-10-15
Part I of force_and_torque_overhaul breakup: Update of permanent magnet helpers, moving coil optimization functionality to its own helper file

akaptano:
I have separated out a substantial piece of PR #509 that relates to adding useful coil optimization functionality to its own util file, as well as reducing the time spent in running the permanent magnet examples, and otherwise reorganizing some functions between the permanent magnet and coil helper functions in utils/. 

I would like to get this merged in main before continuing onwards to Part II. Note that code coverage is not perfect only because I have some functionality waiting to be tested once the force and torque terms are back in there.

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -20,7 +20,7 @@ jobs:
       fail-fast: false
       matrix:
         platform: [ubuntu-24.04, macos-15]
-        python-version: ["3.9"]
+        python-version: ["3.12"]
 
     runs-on: ${{ matrix.platform }}
 
@@ -45,7 +45,7 @@ jobs:
           channel-priority: strict
 
       - name: Prepare
-        run: conda install -y conda-build conda-verify setuptools_scm anaconda-client scikit-build-core
+        run: conda install -y conda-build setuptools_scm anaconda-client scikit-build-core
 
       - name: Build and upload
         if: github.event_name == 'release' && github.event.action == 'published'
--- a/.github/workflows/docs_test.yml
+++ b/.github/workflows/docs_test.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.11"]
+        python-version: ["3.12"]
 
     steps:
     - uses: actions/checkout@v4
--- a/.github/workflows/extensive_test.yml
+++ b/.github/workflows/extensive_test.yml
@@ -35,12 +35,11 @@ jobs:
       fail-fast: false
       matrix:
         test-type: [unit, integrated]
-        # 2025-11-28: Temporarily disabling SPEC tests in this workflow due to github issue #561
         python-version: ["3.10", "3.11", "3.12", "3.13"]
         include:
-            - python-version: "3.11"
+            - python-version: "3.13"
               test-type: unit
-            - python-version: "3.11"
+            - python-version: "3.13"
               test-type: integrated
 
     steps:
@@ -186,36 +185,15 @@ jobs:
         # python -m pip install -v git+https://github.com/zhisong/pyoculus
         # python -m pip install py_spec h5py
 
-
-    - name: Run serial examples as part of integrated tests
-      if: contains(matrix.test-type, 'integrated')
+    - name: Run examples as part of integrated tests
+      if: contains(matrix.test-type, 'integrated') 
       run: |
         cd examples
         ./run_serial_examples
-
-    - name: Run simple parallel examples as part of integrated tests
-      if: contains(matrix.test-type, 'integrated')
-      run: |
-        cd examples
         ./run_parallel_examples
-
-    # - name: Run spec examples as part of integrated tests
-    #  if: contains(matrix.test-type, 'integrated')
-    #  run: |
-    #    cd examples
-    #    ./run_spec_examples
-
-    - name: Run vmec examples as part of integrated tests
-      if: contains(matrix.test-type, 'integrated')
-      run: |
-        cd examples
+        # ./run_spec_examples
         ./run_vmec_examples
-
-    # - name: Run vmec and spec examples as part of integrated tests
-    #  if: contains(matrix.test-type, 'integrated')
-    #  run: |
-    #    cd examples
-    #    ./run_spec_vmec_examples
+        # ./run_spec_vmec_examples
 
     - name: Install coverage dependencies
       if: contains(matrix.test-type, 'unit')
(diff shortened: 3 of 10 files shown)


src/simsopt/util/coil_optimization_helper_functions.py
```diff
@@ -0,0 +1,1170 @@
...
+        msc_string = ", ".join(f"{J.J():.1f}" for J in Jmscs)
+        outstr += f", Len=sum([{cl_string}])={sum(J.J() for J in Jls):.1f}, ϰ=[{kap_string}], ∫ϰ²/L=[{msc_string}]"
+        outstr += f", C-C-Sep={Jccdist.shortest_distance():.2f}, C-S-Sep={Jcsdist.shortest_distance():.2f}"
+        outstr += f", ║∇J║={np.linalg.norm(grad):.1e}"
+        print(outstr)
+        return J, grad
+
+    print("""
+    ################################################################################
+    ### Perform a Taylor test ######################################################
```
mishapadidar: Is taylor test necessary here? I see it fitting better with unit tests

src/simsopt/util/coil_optimization_helper_functions.py
```diff
@@ -0,0 +1,1170 @@
...
+
+
+def make_qfm(s, Bfield, n_iters=200):
+    """
+    Given some Bfield generated by dipoles AND a set of TF coils, 
+    compute a quadratic flux-minimizing surface (QFMS)
+    for the total field configuration on the surface s.
+
+    Args:
+        s: plasma boundary surface.
```
mishapadidar: Arg types please

src/simsopt/util/coil_optimization_helper_functions.py
```diff
@@ -0,0 +1,1170 @@
...
+                                0.0]
+                               )
+    bs.set_points(bspoints)
+    B0 = np.linalg.norm(bs.B(), axis=-1)
+    B0avg = np.mean(np.linalg.norm(bs.B(), axis=-1))
+    print("Bmag at R = ", R0, ", Z = 0: ", B0)
+    print("toroidally averaged Bmag at R = ", R0, ", Z = 0: ", B0avg)
+    return B0avg
+
+def make_Bnormal_plots(bs, s_plot, out_dir='', bs_filename="Bnormal", B_axis=None):
```
mishapadidar: this seems like a `to_vtk` function rather than a plotting function. Do we want to move this functionality into field or geo?

src/simsopt/util/coil_optimization_helper_functions.py
```diff
@@ -0,0 +1,1170 @@
...
+        init = df.sample()
+
+        # FIXED PARAMETERS
+        ARCLENGTH_WEIGHT = 0.01
+        UUID_init_from = init['UUID'].iloc[0]
+        ncoils = init['ncoils'].iloc[0]
+        order = init['order'].iloc[0]
+        R1 = init['R1'].iloc[0]
+
+        # RANDOM PARAMETERS
```
mishapadidar: why random?

src/simsopt/util/coil_optimization_helper_functions.py
```diff
@@ -0,0 +1,1170 @@
...
+        CURVATURE_THRESHOLD = perturb(init, 'max_κ_threshold')
+        MSC_THRESHOLD = perturb(init, 'msc_threshold')
+        CS_THRESHOLD = perturb(init, 'cs_threshold')
+        CC_THRESHOLD = perturb(init, 'cc_threshold')
+        FORCE_THRESHOLD = perturb(init, 'force_threshold')
+        LENGTH_TARGET = perturb(init, 'length_target')
+
+        LENGTH_WEIGHT = perturb(init, 'length_weight')
```
mishapadidar: this is a particular set of weights. Can we generalize?

tests/util/test_coil_optimization_helper_functions.py
```diff
@@ -0,0 +1,327 @@
...
+
+    def test_initial_optimizations_basic(self):
+        """Test basic functionality of initial_optimizations with real file."""
+        with ScratchDir("."):
+            # Copy required files into the temp dir
+            shutil.copy(TEST_DIR / "input.LandremanPaul2021_QA_reactorScale_lowres", ".")
+            
+            output_dir = "qa_output/"
+            
+            initial_optimizations(
```
mishapadidar: for each of these can you add a quick comment saying what the test is?

mishapadidar (commented):
Thanks for the review reminder! 

I am a still a bit confused on the use case of coil_optimization_helper_functions. The module holds mostly convenience functions for building QFMs or running stage 2 optimization. Currently, some of the functions seem to be built with a particular use case or run paradigm in mind. If these features are intended to be user facing, it would be beneficial to make the functions general enough, so users can get the most out of them. Part of my confusion with these functions is that they seem very specific how they are intended to be used or run. For example, there are a few functions to perform optimizations of various types i.e. initial_optimizations_QH, initial_optimizations. It would be best to consolidate these as much as possible. 

A general concern I have with the convenience functions is that some of them are designed for a particular _workflow_ rather than being designed for expanding _functionality_. We want to build our library to expose building blocks, rather than particular workflows. Typically, we relegate workflows to example scripts or tutorials, since adding them to the code breaks consistency: why should some workflows get “special” shortcuts while others don’t. On the other hand, some routines are highly used and might deserve a special convenience function. For example, vacuum stage two optimization solves a common problem. In that sense, maybe it makes sense to package it up. However, when doing so, the method should be built in the same style as a modular "building block" (`BoozerSurface` is a good example of this).

Another point is that some of the functions write files out, without explicitly stating that they will do so in the documentation. See [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment). In general, we should have two functions: one to do the computation, and another to write the files. Having both in one may be confusing to users, and makes testing/maintainability more difficult.

My final comment is that some of these functions may belong in other modules. For example, `make_qfm` could be in field or geo. Why not just move them now. 

For concrete to-dos:
- Lets try to consolidate the various functions that run stage two as much as possible, and remove any that are too "use-case specific".
- Try to avoid writing files unless.
- Move files from the helpers to a better home when possible.

akaptano:
> Thanks Alan. For me it looks like this code does not belong to `simsopt/util`, but rather belongs to `simsopt/field`.

Can we leave this for a future PR? It will take a substantial change to move this stuff, and we already need to do this in a separate PR at some point for the permanent magnet functionality. If you're okay with that, please make any last checks. Thanks!

mbkumar:
self_intersecting requires ground package. Perhaps a call to self_intersecting is not there in the tests in the master branch.

PR #568, opened 2025-11-26
Escape strings correctly, so tests and imports become less noisy

jurasic-pf:
Another PR escaping strings correctly, to silence syntax warnings and make the warnings surfaced during tests more meaningful/easier to filter through.


jurasic-pf:
CI failiures are due to the known SPEC CI issue

PR #569, opened 2025-11-27
Added option to load booz_xform_jax

rogeriojorge:
A new version of booz_xform solely in JAX is available in
https://github.com/uwplasma/booz_xform_jax

This was created due to some difficulties using booz_xform with numpy 2.0, and due to its eigen dependency as an external dependency.

If the user has booz_xform_jax, SIMSOPT would default to that; if the user doesn't have it, then it checks for the regular booz_xform (C++/Python version).

The test_boozer.py tests appear to be passing locally.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -9,3 +9,4 @@ sources =
     /**/lib/python3.9/site-packages
     /**/lib/python3.10/site-packages
     /**/lib/python3.11/site-packages
+    /**/lib/python3.12/site-packages
--- a/.github/workflows/extensive_test.yml
+++ b/.github/workflows/extensive_test.yml
@@ -35,8 +35,10 @@ jobs:
       fail-fast: false
       matrix:
         test-type: [unit, integrated]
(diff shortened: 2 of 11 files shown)


mishapadidar (commented):
Hi Rogerio, 
Thanks for writing the new jax version of boozxform. It seems useful. 

1. Can you add some information to the documentation about this? Specifically, on the [main page](https://simsopt.readthedocs.io/latest/) and the [installation](https://simsopt.readthedocs.io/latest/installation.html) page of the documentation can you add links to the jax version of boozxform and a one or two sentence blurb about how they differ?
2. The way the code is implemented now, Simsopt requires boozxform and jax-boozxform to supply identical methods. It would be bad, for instance, if a new feature is added to one of the packages and not the other. To this end, it may be best to merge the jax version of boozxform into the same repository as boozxform. This way they can be maintained simultaneously. Lets get other stakeholders to chime in here, @landreman

PR #589, opened 2026-01-29
Virtual casing branch

akaptano:
Elizabeth Paul's branch that implements virtual casing evaluation of the B field at arbitrary locations, instead of just restricted to the plasma boundary. Going to do some improvements before this is ready for a look/merge.

Big benefit of this functionality is that we could generate Poincare plots, QFM, VMEC metrics like quasi-symmetry, etc. for plasma surfaces that require B_plasma contributions for accuracy.

### DIFF
--- a/examples/2_Intermediate/vc_field.py
+++ b/examples/2_Intermediate/vc_field.py
@@ -11,8 +11,6 @@ This example requires that the python ``virtual_casing`` module is installed.
 """
 from simsopt.field import VmecVirtualCasingField
 from simsopt.mhd import Vmec
-from simsopt.configs import get_data
-from simsopt.geo import SurfaceRZFourier
 from pathlib import Path
 import matplotlib.pyplot as plt
 import numpy as np
--- a/src/simsopt/field/virtualcasingmagneticfield.py
+++ b/src/simsopt/field/virtualcasingmagneticfield.py
@@ -4,14 +4,12 @@ import logging
 
 from nptyping import NDArray
 import numpy as np
-from scipy.spatial import cKDTree
 
 from .magneticfield import MagneticField
-from .._core.json import GSONDecoder
 
 from virtual_casing import VirtualCasing
 from ..mhd.vmec import Vmec
-from ..geo.surface import best_nphi_over_ntheta
+#from ..geo.surface import best_nphi_over_ntheta
 from ..mhd.vmec_diagnostics import B_cartesian
 from .._core.optimizable import Optimizable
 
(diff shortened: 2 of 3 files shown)


src/simsopt/field/magneticfieldclasses.py
```diff
@@ -16,10 +16,187 @@
...
+        vcasing.setup(
+            digits, nfp, stellsym,
+            src_nphi, src_ntheta, gamma1d,
+            src_nphi, src_ntheta,
+            trgt_nphi, trgt_ntheta)
+        
+        # Store target grid positions for on-surface detection
+        trgt_gamma = trgt_surf.gamma()
+        
+        vc = cls()
```
smiet: the init of the class is run here, which is unspecified, and therefore defaults to MagneticField, It is a bit confusing to follow like this, a pure `__init__`  would be clearer.

akaptano:
Adding some additional unit and convergence tests in a moment.

Regarding more benchmarking of the possible bugs:

1. It is still the case that the time and memory requirements of compute_external_B_offsurf -> infinity if the evaluation points are very close to the plasma surface. The workaround is roughly that if |evaluation_points - plasma_surface_points| < tolerance, the method switches over. I tried fancier things but they tended to be more finicky (e.g. signed_distance_to_surface is tricky because sometimes a point is close to the interpolated surface, but not that close to a target plasma point). 

2. It remains the case that compute_external_B_offsurf and compute_external_B only agree up to some small error, even as resolution gets extremely large. The problem seems to be that the resolution of VMEC is limited -- the residual errors improve with mpol and ntor increasing, but for most plasma boundaries, VMEC gives a failure at some point at these are increased, so hard to say. It is not clear to me if this is good enough for purposes like Poincare plots and so forth.
Screenshot 2026-01-30 at 2 03 18 PM

akaptano:
@ejpaul Are there applications where we need close evaluation of the field near the surface? The unit tests now use the offsurf function with a default tolerance of 0.01 m before it switches to just using the field on the nearest plasma point (we could implement a simple interpolation scheme though). This 0.01m can be reduced if necessary, at the cost of computational expense. But it seems like for the purpose of Poincare plots, this might be suitable? We often don't evaluate field lines right at the boundary, unless there is stochasticity.

dmalhotra:
I've fixed a bug in the off-surface evaluation in the virtual-casing package. I've also added optional parameters to specify the maximum upsample order for near integration to limit the maximum memory use.

dmalhotra:
> @dmalhotra thanks for the fix! A heads up that the currently pinned BIEST versuib will not work, so if people try to install the latest virtual-casing package right now, I think it will fail. You will need to update BIEST to the latest on your end, and commit that change to the master branch of virtual-casing.

@akaptano the virtual-casing package is already using the latest version of BIEST. After `git pull`, you’ll need to run `git submodule update --init --recursive` to update submodules.

dmalhotra:
> which routine should we use to evaluate the field outside a surface due to currents inside it?

I've added the routines to evaluate the field due to internal currents.

Issue #595, opened 2026-02-09
from ground.base import get_context


andrewgiuliani:
the proposed fix that was merged does not work on my machine

https://github.com/hiddenSymmetries/simsopt/blob/21117aa8d0175e157f4cec42f92eff74b8c07755/src/simsopt/geo/surface.py#L13

smiet:
@mbkumar our fix was no good. Can we revert and fix the ground (and Bentley ottman) versions? Also see @andrewgiuliani comment on #585

PR #597, opened 2026-02-09
Revert "Attempt to deal with strange issue with no ground package installed."

andrewgiuliani:
Reverts hiddenSymmetries/simsopt#585

PR #585 skips unit tests for code in production:

```
Test to verify the various optimization algorithms that compute ... skipped 'Libraries to check whether self-intersecting or not are missing'
```

I'm working on forcing ground to use the most recent version that works to address issue #595


mbkumar:
Will look into it

mbkumar (approved):
Looks good to me.

PR #601, opened 2026-03-18
Pyoculus migration

smiet:
`pyoculus` repo moved from private ownership by @zhisong to github organization `pyoculus` for future maintainability.

Organization has owners @smiet  and @zhisong. 
Organization structure allows sharing and changing of ownership of repo and future maintainability, as well as automating pypi uploads (wip on pyoculus end)


mbkumar (changes requested):
Can you also bring back the SPEC related functionality back along with this in a single PR? It does not make sense to have these changes when SPEC is effectively disabled.

PR #613, opened 2026-03-27
Add Boozer quasi-isodynamic workflow

rogeriojorge:
## Summary
- add the Boozer-surface quasi-isodynamic objective `NonQuasiIsodynamicRatio` and the supporting helper machinery in `surfaceobjectives.py`
- add a `boozerQI.py` example that stays close to `boozerQA.py` while targeting the milestone-1 QI residual
- add regression coverage for the legacy QI helper math, the new objective, and the reduced-runtime example path
- include small doc and docstring fixes that reduce strict-docs noise encountered while validating the new example page

## Validation
- `python -m pytest tests/geo/test_boozer_qi_math.py tests/geo/test_surface_objectives.py -k 'boozer_qi_math or nonQIratio or boozer_qi_example_reduced_runtime' -q`
- `python examples/2_Intermediate/boozerQI.py` with reduced-runtime environment settings completed successfully
- broader Boozer/QI regression bar: 30 passed, 13 deselected, 45 subtests passed
- extra confidence pass completed on the current branch, including reduced-runtime convergence and Boozer-line comparison plots

Confidence plots from this branch:

Boozer QI convergence

Boozer QI Boozer-line comparison

## Notes
- A fully warning-free global docs build was not treated as a gate for this branch.
- The broader Boozer regression excludes `test_boozer_surface_optimisation_convergence` because it requires the optional `ground` package for `is_self_intersecting`.
- The local Python environment has an editable-install hook that can drift back to a different checkout; the example and QI-specific tests were made more hermetic against that drift.


mishapadidar:
Hi Rogerio, 

Thanks for adding the QI metric for boozer surface. 
Before we begin review, can you remove all files that aren't directly related to the new metric. It would be best to have a clean PR just related to the new QI metric i.e. the new metric, the unit tests, example. 

Does this function run on QUASR 0104183? I noticed that Alan Goodman's QI metric would fail when using the Vmec version of this configuration. 

Misha

src/simsopt/geo/surfaceobjectives.py
```diff
@@ -878,6 +879,401 @@ def dJ_by_dsurfacecoefficients(self):
...
+                                                  self.boozer_surface.res['iota'], phi_shift)
+        residual = _qi_residual_vector_single_surface(modB_lines, phi_values, self.nBj, self.nphi_out)
+        return float(np.dot(residual, residual))
+
+    def _finite_difference_gradient(self):
+        base_x = self.biotsavart.x.copy()
+        gradient = np.zeros_like(base_x)
+
+        for index in range(base_x.size):
+            step = self.fd_abs_step + self.fd_rel_step * max(1.0, abs(base_x[index]))
```
andrewgiuliani, line 1251: it's taking finite differences, which defeats the purpose somewhat of using Boozer surfaces.  It needs to use adjoints to get the gradient.

andrewgiuliani (changes requested):

andrewgiuliani:
what QI metric does this implement?

rogeriojorge:
Hi all, I didn't mean to ask for review yet; it was in draft mode. Let me close this PR until it is ready for review.

PR #624, opened 2026-05-06
Desc Integration

mishapadidar:
This PR adds a Simsopt interface for the [DESC](https://desc-docs.readthedocs.io/) ideal MHD equilibrium code.
One new file is introduced, `simsopt/mhd/desc.py`, with unit tests `tests/mhd/test_desc.py`.

DESC is integrated through a runtime_checkable Python Protocol that specifies the minimum interface a DESC Equilibrium object must satisfy to be used within Simsopt (attributes `surface, pressure, current, iota, Psi`, and methods `solve`/`compute`). This allows duck-typing checks (`isinstance(eq, DescEquilibriumProtocol)`) without hard-coupling to a specific DESC version.

A Simsopt Optimizable, `DescOptimizable`, wraps a DESC Equilibrium and exposes it to the Simsopt optimization framework. 

All child-object DOFs flow through Simsopt's standard depends_on / recompute_bell mechanism, so any change to boundary shape or profiles automatically flags the equilibrium as stale. The class also provides utilities for converting between Simsopt and DESC surface/profile representations, loading from DESC/VMEC input and HDF5 files, rescaling equilibria, and exporting to VMEC (`to_vmec, to_wout, to_vmec_input`).

### DIFF
--- a/src/simsopt/mhd/desc.py
+++ b/src/simsopt/mhd/desc.py
@@ -1,3 +1,11 @@
+# coding: utf-8
+# Copyright (c) HiddenSymmetries Development Team.
+# Distributed under the terms of the MIT License
+
+"""
+This module provides a class that handles the DESC equilibrium code.
+"""
+
 import logging
 import os
 import re
@@ -10,6 +18,8 @@ logger = logging.getLogger(__name__)
 
 from simsopt.mhd import ProfileSpline, ProfilePolynomial, Vmec
 from simsopt._core import Optimizable
+
+# TODO: remove this import after merging the surface conversion.
 from constellaration.geometry.surface_utils_desc import (
     to_desc_fourier_rz_toroidal_surface,
     from_desc_fourier_rz_toroidal_surface,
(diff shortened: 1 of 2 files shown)


jurasic-pf:
Great idea, I think the protocol is a nice choice for loose coupling

src/simsopt/mhd/desc.py
```diff
@@ -0,0 +1,765 @@
...
+    """An Optimizable object for interfacing with the DESC code to solve Ideal MHD equilibria.
+
+    In addition to passing in a Surface object representing the plasma boundary, the edge toroidal
+    flux, the pressure profile, and one of the iota or current profile should be prescribed. Both
+    rotational transform and current profiles can be passed in. When solve() is called, one of the
+    profiles must be chosen to be used.
+
+    The profiles should represent p(rho), I(rho) (or iota(rho)) where rho is Desc's normalized radial
+    coordinate; the profiles should not represent derivatives such as p'(rho), I'(rho). rho
+    relates to toroidal flux psi as rho = (psi / psi_edge)^2, so it is standard for profiles
```
YigitElma, line 91:
```suggestion
    relates to toroidal flux psi as rho = sqrt(psi / psi_edge), so it is standard for profiles
```

src/simsopt/mhd/desc.py
```diff
@@ -0,0 +1,765 @@
...
+            eq (Equilibrium, optional): Optionally pass in a Desc Equilibrium object which will be used
+                for evaluation etc. If None, a new Equilibrium object will be created. Defaults to None.
+            **kwargs: Keyword arguments passed to DESC Equilibrium() constructor. See the documentation of
+                Equilibium object for an up to date list of the keyword arguments.
+
+        Return:
+            None
+        """
+
+        boundary_desc = self.surface_to_desc(self.boundary)
```
YigitElma, line 418: These can be moved inside the `if eq is None` block

src/simsopt/mhd/desc.py
```diff
@@ -0,0 +1,765 @@
...
+        self.eq = eq
+
+        scale_pressure = kwargs.get("scale_pressure", False)
+        self.need_to_run_code = not scale_pressure
+
+        return
+
+    def solve(self, *args, **kwargs):
+        """Call the Equilibium.solve() method of the Desc Equilibrium
+        Any args, kwargs will be passed to the solve() method.
```
YigitElma, line 529: It might be good to mention in the docs that the equilibrium won't be solved if `need_to_run_code=False`. Apart from optimization, if a user interacts with `self.eq` in some way, this may introduce some inconsistency that requires re-solve.

YigitElma (commented):
I just had a quick look at it. A couple of minor suggestions.

I am not very familiar with `simsopt,` but it looks like `need_to_run_code` is set to `True` during optimization(?) if necessary. I think it would be nice to mention in the docs that the consistency between `Desc.eq` and `Desc.boundary`, etc., is held via this parameter. If a user calls `desc` functions on the equilibrium object, that might cause some inconsistencies later.

Issue #625, opened 2026-05-07
Dictionary-like ordering of ancestors in `Optimizable._get_ancestors`


IssraAli:
For an `Optimizable` that inherits from multiple ancestors of the same class, the order of the DOFs (i.e., the order of DOF names in `Optimizable.dof_names` following dictionary-like/lexographic ordering (i.e., a10 would precede a9). This is harmful when more than 10 ancestors of the same type in total are created, particularly when accessing `Optimizable.x` to set variables: 

```Python
from simsopt.geo import CurveXYZFourier
from simsopt._core import Optimizable

class BunchOfCurves(Optimizable):
    """
    Create an optimizable set of curves, given a list of the orders
    of each curve, or a list of curves. 
    """
    def __init__(self, order_list, curve_list=None):
        if curve_list is None:
            curve_list=[]
            for order in order_list:
                curve = CurveXYZFourier(quadpoints=30, order=order)
                curve_list.append(curve)
        Optimizable.__init__(self, depends_on=curve_list)

orderlist = [1, 2, 2, 2, 1]
curveset0 = BunchOfCurves(order_list=orderlist) #['CurveXYZFourier1:...', ... , 'CurveXYZFourier5:...']
curveset1 = BunchOfCurves(order_list=orderlist) #['CurveXYZFourier10:...', 'CurveXYZFourier6:...', ... , 'CurveXYZFourier9:...']
curveset1.x = curveset0.x # does not set curve DOFs in ascending order!
```
The above example features a class `BunchOfCurves` containing `CurveXYZFourier`s of different resolution; the resolutions are specified with the list `order_list`. In the final line, `CurveXYZFourier10` in `curveset1` would get the DOFs of `CurveXYZFourier1` from `curveset0`, and 'CurveXYZFourier6' would get values from (the starting half of) 'CurveXYZFourier2'. More intuitive behaviour would be if all of the DOFs were assigned in ascending order. Though the above example seems contrived, this is an actual issue I ran into during my implementation of splines (#578, in progress), where I have "cross sections" of spline DOFs (with different numbers of DOFs) as ancestors of the same `SurfaceBSpline` object.

PR #626, opened 2026-05-07
Address dictionary-like naming of ancestors

IssraAli:
Addressing issue #625

For an `Optimizable` that inherits from multiple ancestors of the same class, the order of the DOFs (i.e., the order of DOF names in `Optimizable.dof_names` follows dictionary-like/lexographic ordering (i.e., a10 would precede a9). This is harmful when more than 10 ancestors of the same type in total are created, particularly when accessing `Optimizable.x` to set variables: 

```Python
from simsopt.geo import CurveXYZFourier
from simsopt._core import Optimizable

class BunchOfCurves(Optimizable):
    """
    Create an optimizable set of curves, given a list of the orders
    of each curve, or a list of curves. 
    """
    def __init__(self, order_list, curve_list=None):
        if curve_list is None:
            curve_list=[]
            for order in order_list:
                curve = CurveXYZFourier(quadpoints=30, order=order)
                curve_list.append(curve)
        Optimizable.__init__(self, depends_on=curve_list)

orderlist = [1, 2, 2, 2, 1]
curveset0 = BunchOfCurves(order_list=orderlist) #['CurveXYZFourier1:...', ... , 'CurveXYZFourier5:...']
curveset1 = BunchOfCurves(order_list=orderlist) #['CurveXYZFourier10:...', 'CurveXYZFourier6:...', ... , 'CurveXYZFourier9:...']
curveset1.x = curveset0.x # does not set curve DOFs in ascending order!
```
The above example features a class `BunchOfCurves` containing `CurveXYZFourier`s of different resolution; the resolutions are specified with the list `order_list`. In the final line, `CurveXYZFourier10` in `curveset1` would get the DOFs of `CurveXYZFourier1` from `curveset0`, and 'CurveXYZFourier6' would get values from (the starting half of) 'CurveXYZFourier2'. More intuitive behaviour would be if all of the DOFs were assigned in ascending order.

### DIFF
--- a/src/simsopt/_core/optimizable.py
+++ b/src/simsopt/_core/optimizable.py
@@ -47,7 +47,7 @@ except ImportError:
 
 log = logging.getLogger(__name__)
 
-__all__ = ['Optimizable', 'make_optimizable', 'natural_key', 'load', 'save',
+__all__ = ['Optimizable', 'make_optimizable', 'load', 'save',
            'OptimizableSum', 'ScaledOptimizable']
 
 
@@ -917,9 +917,23 @@ class Optimizable(ABC_Callable, Hashable, GSONable, metaclass=OptimizableMeta):
         for parent in self.parents:
             ancestors += parent.ancestors
         ancestors += self.parents
+
+        def _natural_key(text):
+            """
+            Return a key to be used in natural sorting of strings containing
+            integers.
+            Args:
+                s:
+                    String, containing integers
+            Returns:
+                Key string
+            """
+            return [int(s) if s.isdigit() else s.lower()
+                    for s in re.split(r'(\d+)', text)]
+
         return sorted(
             dict.fromkeys(ancestors),
-            key=lambda a: natural_key(a.name)
+            key=lambda a: _natural_key(a.name)
         )
 
     @property
@@ -1614,18 +1628,6 @@ class Optimizable(ABC_Callable, Hashable, GSONable, metaclass=OptimizableMeta):
                 contents = f.read()
             return cls.from_str(contents, fmt="json")
 
-def natural_key(text):
-    """
-    Return a key to be used in natural sorting of strings containing
-    integers.
-    Args:
-        s:
-            String, containing integers
-    Returns:
-        Key string
-    """
-    return [int(s) if s.isdigit() else s.lower()
-            for s in re.split(r'(\d+)', text)]
 
 
 def load(filename, *args, **kwargs):
--- a/src/simsopt/_core/optimizable.py
+++ b/src/simsopt/_core/optimizable.py
@@ -47,7 +47,7 @@ except ImportError:
 
 log = logging.getLogger(__name__)
 
-__all__ = ['Optimizable', 'make_optimizable', 'load', 'save',
+__all__ = ['Optimizable', 'make_optimizable', 'natural_key', 'load', 'save',
            'OptimizableSum', 'ScaledOptimizable']
 
 
@@ -1614,7 +1614,7 @@ class Optimizable(ABC_Callable, Hashable, GSONable, metaclass=OptimizableMeta):
                 contents = f.read()
             return cls.from_str(contents, fmt="json")
 
-def natural_key(s):
+def natural_key(text):
     """
     Return a key to be used in natural sorting of strings containing
     integers.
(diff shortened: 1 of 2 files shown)


mishapadidar:
does this need review? @IssraAli

IssraAli:
Hi Misha, yes, this needs review.

mishapadidar:
@mbkumar, @landreman mentioned that you had dealt with a sorting issue in the past. Is this related?

mishapadidar:
Hi Issra, thanks for the changes. please add unit tests. Please write a more thorough description of the PR (you could copy some of the info from the issue)

mbkumar:
To be honest, the ordering issue was not considered, because I never thought dofs of an optimization graph would be copied from one object to another object.

mishapadidar:
One problem i foresee is in loading existing optimizable values from a json. In what order, would the dofs of an existing file be loaded?

mbkumar:
> One problem i foresee is in loading existing optimizable values from a json. In what order, would the dofs of an existing file be loaded?

In the json, the Optimizable objects along with their local dofs are stored. So ordering of global dofs is never a consideration. The graph ordering of the Optimizable objects is preserved in the json generated by simsopt.

In the issue raised, global dofs of one Optimizable object are assigned to another Optimizable object. I never thought of that use case. Typically global dofs are used for the optimization run and used only for the final object.

IssraAli:
> In the issue raised, global dofs of one Optimizable object are assigned to another Optimizable object. I never thought of that use case. Typically global dofs are used for the optimization run and used only for the final object.

It came up for me when I was running optimizations using other packages (for example, BoTorch) that accept a numpy array as the variable optimized over. In the optimization loop, I set the dofs using a line like `myopt.x = x_new`

mbkumar:
> 
> It came up for me when I was running optimizations using other packages (for example, BoTorch) that accept a numpy array as the variable optimized over. In the optimization loop, I set the dofs using a line like `myopt.x = x_new`

```
x = myopt.x # where x is global dof array
# pass x for optimizer and get x_new
myopt.x = x_new
```
This loop should work.

mbkumar:
But I like the updated sorting which is better.

src/simsopt/_core/optimizable.py
```diff
@@ -1610,6 +1614,18 @@ def from_file(cls, filename: str):
...
+    """
+    Return a key to be used in natural sorting of strings containing
+    integers.
+    Args:
+        s:
+            String, containing integers
+    Returns:
+        Key string
+    """
+    return [int(text) if text.isdigit() else text.lower()
```
mbkumar: Swap `s` and `text`. Using short variables in loop comprehension makes it easy to read.

mbkumar (changes requested):

tests/core/test_optimizable.py
```diff
@@ -928,6 +928,32 @@ def test_full_dof_names(self):
...
+    
+        # Check at parent level
+        identity_list = [Identity() for i in range(11)]
+        adder_list = [Adder(n=11) for i in range(11)]
+        test_obj3 = OptClassWithParents(val=10, depends_on= (identity_list + adder_list))
+        dofs_names_ints = [list(map(int, re.findall(r'\d+', s))) for s in test_obj3.dof_names]
+        unique_ancestor_namelist = []    
+        for dof_name in test_obj3.dof_names:
+            ancestor_name = re.split(':', dof_name)[0]
+            if ancestor_name not in unique_ancestor_namelist:
```
mishapadidar, line 943: add a comment or two to disambiguate test case

src/simsopt/_core/optimizable.py
```diff
@@ -47,7 +47,7 @@
 
 log = logging.getLogger(__name__)
 
-__all__ = ['Optimizable', 'make_optimizable', 'load', 'save',
+__all__ = ['Optimizable', 'make_optimizable', 'natural_key', 'load', 'save',
```
mbkumar: `natural_key` should not be exported. Please remove this.

src/simsopt/_core/optimizable.py
```diff
@@ -1614,7 +1614,7 @@ def from_file(cls, filename: str):
                 contents = f.read()
             return cls.from_str(contents, fmt="json")
 
-def natural_key(s):
+def natural_key(text):
```
mbkumar: Make this a private function `_natural_key`.

src/simsopt/_core/optimizable.py
```diff
@@ -1610,6 +1614,19 @@ def from_file(cls, filename: str):
                 contents = f.read()
             return cls.from_str(contents, fmt="json")
 
+def natural_key(text):
```
mbkumar: Put`_natural_key` inside `_get_ancestors`.

mbkumar (changes requested):

mbkumar (approved):

PR #627, opened 2026-05-07
Vmec Protocol and Integration

mishapadidar:
This PR generalizes the interface between Simsopt and the various Vmec solvers (Vmec2000, Vmec++, Vmec-Jax). The framework relies on Protocols to define the minimal requirements for a Vmec solver to be compatible with Simsopt, and to determine how Simsopt relays information to Vmec.

### DIFF
--- a/src/simsopt/mhd/vmec.py
+++ b/src/simsopt/mhd/vmec.py
@@ -110,6 +110,55 @@ def array_to_namelist(arr, aux_s=False):
 #                        if ns_index = 0 and numsteps = 0 (see below), vmec will
 #                        control its own run history
 
+@runtime_checkable
+class SurfaceRZFourierProtocol(Protocol):
+    # also maintained on VMEC side
+    rbc: dict # (m,n): value
+    rbs: dict
+    zbc: dict
+    zbs: dict
+    nfp: int
+    stellsym: bool
+
+@runtime_checkable
+class ProfileProtocol(Protocol):
+    # also maintained on VMEC side
+    name: str
+    x: Array
+    y: Array
+
+@runtime_checkable
+class VmecProtocol(Protocol):
+    """A Protocol for desc.equilibrium.Equilibrium objects. This Protocol determines
+    the basic set of attributes and methods that a DESC Equilibrium must have in order
+    to be used within Simsopt.
+
+    Running,
+        ```
+        from desc.equilibrium import Equilibrium
+        eq = Equilibrium(...)
+        isinstance(eq, DescEquilibriumProtocol)
+        ```
+    will check the DESC Equilibrium object has the attributes and methods defined by the DescEquilibriumProtocol.
+    If False, then `eq` does not have the necessary structure to be used within Simsopt.
+    This check should be implemented by all methods that rely directly (though not indirectly)
+    on the DESC Equilibrium object.
+    """
+
+    surface: SurfaceRZFourierProtocol
+    pressure: ProfileProtocol
+    current: Optional[ProfileProtocol]
+    iota: Optional[ProfileProtocol]
+
+    # TODO: need access to settings (phiedge, ...)
+    vmec_input: Any
+
+    # needed for compute_geometry
+    wout: Any
+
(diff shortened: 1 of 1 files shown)


src/simsopt/mhd/vmec.py
```diff
@@ -110,6 +112,63 @@ def array_to_namelist(arr, aux_s=False):
...
+        ```
+    will check the Vmec object has the attributes and methods defined by the VmecProtocol.
+    If False, then `eq` does not have the necessary structure to be used within Simsopt.
+    This check should be implemented by all methods that rely directly (though not indirectly)
+    on the Vmec object.
+    """
+
+    surface: SurfaceRZFourierProtocol
+    pressure: ProfileProtocol
+    current: Optional[ProfileProtocol]
```
jurasic-pf, line 160: Let's be explicit about which takes precedence if both are set. I would recommend current taking precedence.

src/simsopt/mhd/vmec.py
```diff
@@ -110,6 +112,63 @@ def array_to_namelist(arr, aux_s=False):
...
 #                        control its own run history
 
+@runtime_checkable
+class SurfaceRZFourierProtocol(Protocol):
+    """A Protocol for facility the interaction between Vmec solvers (Vmec2000, Vmec++, Vmec-Jax, ...). 
+    This Protocol determines how SurfaceRZFourier object data is passed from Simsopt to Vmec solvers.
+    This protocol should be maintained here and also in the solver.
+    """
+    # also maintained on VMEC side
+    rbc: dict # (m,n): value
```
jurasic-pf, line 122: @jons-pf pointed out that there's of course an ambiguity between `n, m` and `m, n`. How about explicit keying?
```json
"rbc":[
{"n":   0,"m":   0,"value": 0.100000000000000000E+01},
{"n":   2,"m":   0,"value":-0.588235294117647120E-01},
...
```

src/simsopt/mhd/vmec.py
```diff
@@ -110,6 +112,63 @@ def array_to_namelist(arr, aux_s=False):
...
+@runtime_checkable
+class ProfileProtocol(Protocol):
+    """A Protocol for facility the interaction between Vmec solvers (Vmec2000, Vmec++, Vmec-Jax, ...). 
+    This Protocol determines how Simsopt Profile object data is passed from Simsopt to Vmec solvers.
+    This protocol should be maintained here and also in the solver.
+    """
+    # also maintained on VMEC side
+    name: str
+    x: Union[np.ndarray, list]
+    y: Union[np.ndarray, list]
```
jurasic-pf, line 138: Given our discussion, maybe callable is fine after all.
we know that GVEC, DESC, vmecpp support arbitrary callable and for VMEC2000 we can provide a lossless conversion

src/simsopt/mhd/vmec.py
```diff
@@ -110,6 +112,63 @@ def array_to_namelist(arr, aux_s=False):
...
+    on the Vmec object.
+    """
+
+    surface: SurfaceRZFourierProtocol
+    pressure: ProfileProtocol
+    current: Optional[ProfileProtocol]
+    iota: Optional[ProfileProtocol]
+
+    # needed access to settings (phiedge, ...)
+    vmec_input: Any
```
jurasic-pf, line 164: Basically this is the only part that makes this VMEC specific now, which is nice.
This could be any "solver_input" for other solvers

src/simsopt/mhd/vmec.py
jurasic-pf: That's how we handle it in `vmecpp` input json for that reason

src/simsopt/mhd/vmec.py
mishapadidar: @jurasic-pf precedence should be controlled via `ncurr` right?

jurasic-pf:
Unfortunately I'm quite busy until end of July, but I'll happily pick this project back up afterwards. Just to provide context, this has not been forgotten

PR #630, opened 2026-05-11
Added BoozXform Protocol and BoozerOutput

mishapadidar:
**Summary**
This PR introduces a BoozXform Protocol to simplify connecting Simsopt to multiple booz_xform modules (booz_xform, booz_xform-jax). This PR also introduces the `BoozerOutput` object which captures the output of the booz_xform. Simsopt classes should access the `BoozerOutput` class rather than the `Boozer.bx` object.

**New Classes**
Two new public classes are introduced in `simsopt.mhd.boozer`:

- `BoozXFormProtocol`: a `@runtime_checkable` Protocol describing the structural interface required of any Booz_xform-like object. Enables duck-typed alternatives to be validated with `isinstance`.
- `BoozerOutput`: a `@dataclass` collecting all output quantities produced by a booz_xform run (Fourier coefficients, rotational transform, surface grids, etc.). 

**Changes to Boozer**
- `bx` is now a validated property: the setter checks `isinstance(value, BoozXFormProtocol)` and raises `TypeError` on incompatible objects.
- `output` is a new property. It raises `RuntimeError` if accessed before `run()` has been called. `run()` populates it with a `BoozerOutput` at the end of each transformation.

**Downstream updates**
`Quasisymmetry.J()`, `boozermagneticfield.py`, and `bootstrap.py` are updated to read results from `boozer.output` instead of `boozer.bx` directly.

**Tests**
Unit tests are added for `BoozXFormProtocol` (protocol satisfaction and rejection), `BoozerOutput` (construction for symmetric and asymmetric configs), and the bx / output properties.

### DIFF
--- a/src/simsopt/mhd/boozer.py
+++ b/src/simsopt/mhd/boozer.py
@@ -8,7 +8,8 @@ Boozer coordinates, and an optimization target for quasisymmetry.
 """
 
 import logging
-from typing import Union, Iterable
+from dataclasses import dataclass
+from typing import Union, Iterable, Protocol, runtime_checkable, Any, Optional
 
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


rogeriojorge (approved):
Looks good to me! Proceeding to the full CI for testing

src/simsopt/mhd/boozer.py
```diff
@@ -31,7 +32,116 @@
...
+                       zmns: Any,
+                       lmnc: Any,
+                       lmns: Any,
+                       bmnc: Any,
+                       bmns: Any,
+                       bsubumnc: Any,
+                       bsubumns: Any,
+                       bsubvmnc: Any,
+                       bsubvmns: Any) -> None:
+        """Load VMEC equilibrium data into the Booz_xform object."""
```
mbkumar, line 88: Can you use `numpy.typing.ArrayLike` in place of `Any`?

src/simsopt/mhd/boozer.py
```diff
@@ -31,7 +32,116 @@
...
+    zmns_b: Any
+    numns_b: Any
+    gmnc_b: Any
+
+    # --- Fourier coefficients (asymmetric; 0x0 arrays when asym=False) ---
+    bmns_b: Any
+    rmns_b: Any
+    zmnc_b: Any
+    numnc_b: Any
+    gmns_b: Any
```
mbkumar, line 144: Same here. Try using using `npt.ArrayLike` in place of `Any`.

mbkumar:
Do you expect users to change bx during the simulation? If not why don't you supply that in the `__init__` for Boozer class?

src/simsopt/mhd/boozer.py
```diff
@@ -57,6 +167,41 @@ class Boozer(Optimizable):
...
+
+        Returns a :class:`BoozerOutput` containing all output quantities (Fourier
+        coefficients, rotational transform, surface grids, etc.) produced by the
+        last call to :meth:`run`. Raises ``RuntimeError`` if :meth:`run` has not
+        yet been called.
+        """
+        if self._output is None:
+            raise RuntimeError(
+                "Boozer.run() has not been called yet. Call run() before accessing output.")
+        return self._output
```
mbkumar, line 199: Instead of throwing error, calling `self._bx.run()` is a better idea.

src/simsopt/mhd/boozer.py
```diff
@@ -57,6 +167,41 @@ class Boozer(Optimizable):
...
+
+    @output.setter
+    def output(self, value: Optional['BoozerOutput']) -> None:
```
mbkumar, line 202: This setter does not make sense to me if you incorporate automatic calling to self._bx.run() in the property definition.

src/simsopt/mhd/boozer.py
```diff
@@ -57,6 +167,41 @@ class Boozer(Optimizable):
...
+        structured :class:`BoozerOutput` object after :meth:`run` has been called.
+        """
+        return self._bx
+
+    @bx.setter
+    def bx(self, value: BoozXFormProtocol) -> None:
+        if not isinstance(value, BoozXFormProtocol):
+            raise TypeError(
+                f"bx must implement BoozXFormProtocol, got {type(value)}")
+        self._bx = value
```
mbkumar, line 185: Getter and setter for `bx` don't make much sense. You don't change BoozXForm during the simulation.
It is better to do the `isinstance` check  in the `__init__` itself.

src/simsopt/mhd/boozer.py
```diff
@@ -31,7 +32,116 @@
...
+        ...
+
+    def run(self) -> None:
+        """Execute the Boozer coordinate transformation on ``compute_surfs``."""
+        ...
+
+
+@dataclass
+class BoozerOutput:
+    """All output quantities produced by a ``Booz_xform`` run.
```
mbkumar, line 98: This class does not make any sense to me. In the `Boozer` class, all you are doing is copying/grouping the data into this class from the BoozXForm object. Instead, you can define the needed data variables as part of the BoozXForm protocol.

mbkumar (commented):
I don't understand the need for `BoozerOutput` class. Type hinting could be improved. Setter & getter for bx also do not make sense for me.

PR #631, opened 2026-05-16
pyproject edits, fixing example

wingos80:
- Edited pyproject.toml to have fixed pyoculus source where distutils is deprecated, updated required python version to allow project to build with this pyoculus version
- Added uv fields to be git-ignored
- Fixed example docstring and imports

I fixed the pyoculus version to commit [d71914e](https://github.com/pyoculus/pyoculus/commit/d71914eb20a5e9b3ab5059fefeaa3199c69001f6) for now, I wanted to pin it to a release tag on that repo but the latest one still depended on distutils. If another release is made on pyoculus then it would be wise to pin it to that version. Python 3.9 was also the oldest version I could find which allowed building with this new pyoculus requirement, hopefully this is not too big of an issue to upgrade the python requirement.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,10 @@ pip-wheel-metadata/
 .coverage
 .tox
 *fort.9
+
+# uv related
+.venv/
+uv.lock
 	
 # f90wrap related
 src.*/
--- a/examples/1_Simple/logger_example.py
+++ b/examples/1_Simple/logger_example.py
@@ -1,14 +1,14 @@
 #!/usr/bin/env python3
 
 import logging
-from simsopt.util.log import initialize_logging
+from simsopt.util.logger import initialize_logging
 
 """
 Example file for transparently logging both MPI and serial jobs
 """
 
 # Serial logging
-initialize_logging(filename='serial.log')
+initialize_logging(filename='examples/serial.log')
 print("Running 1_Simple/logger_example.py")
 print("==================================")
 for i in range(2):
--- a/examples/1_Simple/tracing_fieldlines_NCSX.py
+++ b/examples/1_Simple/tracing_fieldlines_NCSX.py
@@ -4,7 +4,7 @@
 This example demonstrates how to use SIMSOPT to compute Poincare plots.
 
 This script uses the NCSX coil shapes available in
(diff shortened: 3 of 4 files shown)


wingos80:
I think a maintainer could review if affixing pyoculus source to its' repo is good to do moving forward, but perhaps @smiet could also pitch in on this matter as well.

smiet:
Hi @wingos80, for now SPEC and pyoculus are removed from the tests, they are being overhauled and API breaking changes are being implemented. 

Best would be to use the pyoculus release tagged as  0.3.3, does this (pypi.org/project/pyoculus/0.3.3/) work? That would be better than affixing to a specific commit hash. I made this release hoping the build would be stable for the forseeable future as it uses meson-build. It is also available on pypi so just pyoculus==0.3.3 should work. 

(guess I was doing this around the time you made this PR) LMK if you encounter any issues. 

edit: Re python requirement, 3.8 is deprecated so I am happy to see it go.

wingos80:
Thx for taking a look @smiet, affixing to a version label sounds more clean as a package version, yeah i think when i looked at the pyoculus repo the latest tag was v0.3.2 and it didn't have that commit switching to meson build yet, so i specified that commit in the pyproject instead.

Also I originally git-ignored uv.lock, but in hindsight it seems a bit weird to not find a lock/requirements file for this project, so I think it would be good to include it afterall so that there's atleast one set of fixed dependencies.

uv.lock
jurasic-pf, line 1: I don't think the uv.lock file should be committed, likely unintentional since you've also added it to gitignore which I appreciate.

wingos80: I originally thought to not include it, but then a bit more thinking made me think it would be useful to freeze some version of simsopt in a lock file to make the versions more reproducible. But on further research looks like other projects have problems with proper tracking of the lock files and it causes more headache than it's worth, so i also think removing it would be a better idea.

wingos80: by further research i ofcourse mean reading [one stack exchange post](https://stackoverflow.com/questions/61037557/should-i-commit-lock-file-changes-separately-what-should-i-write-for-the-commi) ;)

Issue #487, opened 2025-05-03
Some SIMSOPT coil objectives become computationally intractable as number of coils increases


akaptano:
There are now a number of force and torque calculations (more coming in the pull request here https://github.com/hiddenSymmetries/simsopt/pull/483) which rely on creating a BiotSavart object in the objective. This becomes a huge problem when there are many of these LpCurveForce objects.  Let's take the class LpCurveForce in force.py as an example. I have attached a minimum working example and corresponding plot that visualizes the fundamental issue. Note that the number of function calls (black line) made by SIMSOPT (just during Jf.x = dofs) **is on the log scale on the right-side of the plot!!!**

Basically, the number of optimizable objects, and all their child processes, AND "weak references" to these processes are growing extremely rapidly and cause the code to come to a halt. Of course, for most applications you will not need so many coils (I discovered this while using dipole arrays) but this is already a significant slowdown with just 8 coils! 

In the meantime, I have used a class MixedLpCurveForce in https://github.com/hiddenSymmetries/simsopt/pull/483 that avoids creating a new BiotSavart object in order to compute forces and torques for dipole coil arrays. But presumably there is a better fix by more carefully handling optimizable objects and their dependencies. 

<img width="996" alt="Image" src="https://github.com/user-attachments/assets/5bdd8e81-635f-4c69-b765-7e645768c85d" />

Pasted below for ncoils = 64 is a print out showing that by far the top three function calls are all coming just from setting the degrees of freedom (Jf.x = dofs). This in principle should just be setting a few hundred floating point numbers, and it is taking tens of millions of function calls and tens of seconds to execute!

<img width="1413" alt="Image" src="https://github.com/user-attachments/assets/f00b3eac-a50e-47ae-92ca-386c4dafcbf8" />

```
from pathlib import Path
from simsopt.geo import create_equally_spaced_curves,SurfaceRZFourier
from simsopt.field import Current, coils_via_symmetries, BiotSavart, LpCurveForce
from simsopt.objectives import SquaredFlux, Weight
from simsopt.field.selffield import regularization_circ
import matplotlib.pyplot as plt
import time
import cProfile
import pstats
import io
ncoils_list = [2, 4, 8, 16, 32, 64]
timings_x = []
timings_J = []
timings_dJ = []
timings_x1 = []
timings_J1 = []
timings_dJ1 = []
timings_x2 = []
timings_J2 = []
timings_dJ2 = []
num_calls1 = []
num_calls2 = []
[...]

PR #634, opened 2026-06-11
Refactored coil optimization helper

wingos80:
The `coil_optimization()` and `vacuum_stage_II_optimization()` from `coil_optimization_helper_functions.py` used an anti-pattern which gives rise to the edge-explosion kind of behaviour seen in [issue 487](https://github.com/hiddenSymmetries/simsopt/issues/487).

To construct the `LpCurveForce` optimizable, the old helper functions passed a single coil as the target coil and the complete list of source coils as the source coils. It then continues to construct additional `LpCurveForce` optimizables this way, passing target coils one at a time, resulting (roughly) in number of edges scaling quadratically to number of coils. Simply passing the entire target coil list in construction of one `LpCurveForce` lowers the scaling to $2N$. 

**Note** while it's faster to create one instance with all target coils passed at once rather than summing instances of each target coil, the two forms are only numerically equivalent when every target coil also appears among the source coils. Because force on one target coil counts contribution from other target coils + source coils. I.e. it doesn't make sense in all problems to pass all target coils at once.  

I ran the `util.test_coil_optimization_helper_functions.TestCoilOptimization` test as a quick sanity check that nothing broke.

In summary, the changes are:
1. Rewriting list comprehensions in `coil_optimization()` and `vacuum_stage_II_optimization()` to pass the complete list of target coils instead of looping over each target.
2. Edited docstrings in `force.py` to suggest more efficient construction of `Optimizable` objects

Would be great if @akaptano, @smiet, or other maintainers could take a look at this.

### DIFF
--- a/src/simsopt/field/force.py
+++ b/src/simsopt/field/force.py
@@ -922,7 +922,11 @@ class SquaredMeanForce(Optimizable):
     To minimize function calls, pass ``target_coils`` as a single list and instantiate this class **once**.
     Creating one instance per coil (``sum(SquaredMeanForce([c], sources, ...) for c in coils)``)
     builds an :math:`O(N^2)` :class:`Optimizable` dependency graph, making
-    ``JF.x = dofs`` extremely slow as ``N`` grows
+    ``JF.x = dofs`` extremely slow as ``N`` grows.
+    This single-instance form is also only numerically equivalent to the
+    one-instance-per-coil form when every target coil also appears in the source
+    coils, since physically the force on each target includes the field from the
+    other target coils.
 
     Args:
         target_coils (list of Coil or RegularizedCoil, shape (m,)): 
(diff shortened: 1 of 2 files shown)


andrewgiuliani:
Woah! This is a great :) congrats, will take a look too

src/simsopt/field/force.py
```diff
@@ -919,6 +919,15 @@ class SquaredMeanForce(Optimizable):
...
     (with barely any quadrature points) and a set of TF coils (with many quadrature points).
 
+    To minimize function calls, pass ``target_coils`` as a single list and instantiate this class **once**.
+    Creating one instance per coil (``sum(SquaredMeanForce([c], sources, ...) for c in coils)``)
+    builds an :math:`O(N^2)` :class:`Optimizable` dependency graph, making
+    ``JF.x = dofs`` extremely slow as ``N`` grows.
+    This single-instance form is also only numerically equivalent to the
+    one-instance-per-coil form when every target coil also appears in the source
+    coils, since physically the force on each target includes the field from the
+    other target coils.
```
andrewgiuliani, line 929: not following this comment, are the target coils' field also included in the force calculation?

andrewgiuliani (changes requested):

src/simsopt/field/force.py
wingos80: this comment is actually wrong and worded badly. I was trying to say that passing all target coils at once through a single list rather than passing them one-at-a-time isn't always equivalent, a better comment would be:

The single-instance form and one-instance-per-coil form are not always equivalent: the force on a target coil includes the field from the other target coils, so with one-instance-per-coil that target-target contribution is missing unless the other coils are passed as sources.

Maybe that's better?

PR #633, opened 2026-05-23
Use conda-build command in workflow

krystophny:
## Summary

This PR fixes the Conda workflow after the current runner stopped exposing `conda build` as a subcommand.

Changes:
- call the `conda-build` executable installed by the prepare step
- keep the existing upload tokens and `--no-test` build mode

### Test fails on main
```bash
$ gh run view 26345424899 --log-failed
...
conda: error: argument COMMAND: invalid choice: 'build'
```

### Test passes after fix
```bash
$ micromamba create -y -p /tmp/simsopt-conda-build-check -c conda-forge python=3.12 conda-build
$ micromamba run -p /tmp/simsopt-conda-build-check conda-build --help
usage: conda build [-h] ... [--token TOKEN] ... [--no-test] ... RECIPE_PATH
$ git diff --check upstream/master..HEAD
```

PR #632, opened 2026-05-23
Fix PR workflow failures

krystophny:
## Summary

This PR fixes three small PR-check failures.

Changes:
- skip DockerHub login unless the Docker workflow runs on `master` or a version tag
- run the branch Docker build for `pull_request` events
- call the `conda-build` executable installed by the Conda prepare step
- allow 0.05% Codecov project drift for workflow-only PRs with unchanged coverable lines

### Test fails on main
```bash
$ gh run view 26345413505 --log-failed
...
docker Login to DockerHub ... ##[error]Password required

$ gh run view 26345424899 --log-failed
...
conda: error: argument COMMAND: invalid choice: 'build'

$ gh api repos/hiddenSymmetries/simsopt/commits/1e714454281e6de8bc7e52792ad203369efc1310/check-runs --jq '.check_runs[] | select(.name=="codecov/project") | .output.title'
90.23% (-0.02%) compared to 1b0cc3a
```

### Test passes after fix
```bash
$ gh run view 26363970991 --log | rg -C 6 "Run conda-build --no-test|BUILD START"
Run conda-build --no-test conda.recipe
BUILD START: ['simsopt-v1.10.6+406_gea519000-py311h656b026_0.conda']

$ micromamba run -p /tmp/simsopt-conda-build-check bash -lc 'command -v conda-build && conda-build --help | sed -n "1,12p"'
/tmp/simsopt-conda-build-check/bin/conda-build
usage: conda build [-h] ... [--token TOKEN] ... [--no-test] ... RECIPE_PATH

$ git diff --check upstream/master..HEAD

$ $HOME/code/prompts/scripts/check-writing-slop.py codecov.yml .github/workflows/conda.yml .github/workflows/docker.yml
PASS: no writing-slop candidates at threshold medium
```

PR #636, opened 2026-07-09
Fix conda workflow: call conda-build binary from active env

krystophny:
Conda `build`/`build-on-PR` jobs fail on current runners because `conda build` is no longer exposed as a conda subcommand (`conda: error: argument COMMAND: invalid choice: 'build'`), and the `conda-build` executable was installed into base only (via `setup-miniconda` `conda-build-version` and `conda install --name base`), so the build steps (running in the activated env) could not find it.

## Changes
- Call the `conda-build` executable instead of the `conda build` subcommand in all build/upload steps.
- Install `conda-build` into the active environment during Prepare; drop the base-only `conda-build-version` pin.

## Background
This is the fix from #633, which was closed as "folded into #632" but never actually landed in the conda workflow — `master` still carries the broken `conda build` subcommand form.

## Verification
- `master` conda `build`/`build-on-PR` currently fail (`invalid choice: 'build'` on current runners).
- After this change the workflow calls the installed `conda-build` binary from the active env. Full confirmation comes from this PR's own conda CI run.

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -52,20 +52,20 @@ jobs:
         if: github.event_name == 'release' && github.event.action == 'published'
         run: |
           conda config --set anaconda_upload yes
-          conda build --token ${{ secrets.ANACONDA_TOKEN }} conda.recipe
+          "$CONDA/bin/conda-build" --token ${{ secrets.ANACONDA_TOKEN }} conda.recipe
 
       - name: Build and upload to test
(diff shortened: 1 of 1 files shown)


krystophny:
@mbkumar review please — conda workflow fix (call the `conda-build` binary from the active env). This is the fix from the closed #633 that never actually landed in the workflow.

mbkumar (approved):
Thanks for the fixes.

PR #637, opened 2026-07-09
Stabilise codecov project/patch gate

krystophny:
The `codecov/project` check fails on every PR: `extensive_test.yml` uploads one coverage session while `tests.yml` is restricted to non-master PRs, so the `master` baseline reports two sessions and the PR head reports one. The resulting fixed ~2-line gap on a ~17896-line report trips the default 0% project tolerance.

## Change
Set a 0.5% project / 1% patch tolerance (`informational: false`) with a standard comment layout in `codecov.yml`, so session-count noise no longer fails CI. Config-only.

## Note
The behavioural tests originally bundled here were split into #638 (they are independent of this gating change).

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,4 +3,15 @@ coverage:
     project:
       default:
         target: auto
-        threshold: 0.05%
+        threshold: 0.5%
+        informational: false
+    patch:
+      default:
+        target: auto
+        threshold: 1%
+        informational: false
+
+comment:
+  layout: "reach, diff, flags, files"
+  behavior: default
+  require_changes: false


krystophny:
@akaptano could you review? Test/config only — core coverage lift and codecov project/patch stabilisation.

krystophny:
Scope narrowed: this PR is now the codecov-gate change only (`codecov.yml`). The behavioural tests moved to #638. @akaptano this one is config-only and unblocks the `codecov/project` check on all open PRs.

mbkumar (approved):

PR #638, opened 2026-07-09
Add behavioural tests for core dev, logging, and LHD-like shim

krystophny:
Adds behavioural unit tests for previously-uncovered public surfaces. Test-only — no source, build, or config changes.

## Coverage
- `tests/core/test_dev.py`: `SimsoptRequires` (unmet condition raises `RuntimeError`, met path calls through, metadata preserved) and `deprecated` (default `FutureWarning` vs `DeprecationWarning`; replacement-name message for function / property / classmethod / staticmethod; appended message; return value + metadata preserved).
- `tests/util/test_logger.py`: `initialize_logging()` non-MPI — default console + rotating-file handlers, custom filename, level override, and the `mpi=True` warning when mpi4py is absent. (The warning is observed via `logging.warning`, since `initialize_logging` reconfigures the root logger through `dictConfig`, which removes any handler `assertLogs` would install.)
- `tests/configs/test_LHD_like.py`: `get_LHD_like_data()` legacy shim emits `DeprecationWarning`, returns the documented 3-tuple shape, and stays numerically consistent with `get_data('lhd_like')`.

Each test exercises the public entry point and asserts observable behaviour. Verified via this PR's CI.

### DIFF
--- a/tests/configs/test_LHD_like.py
+++ b/tests/configs/test_LHD_like.py
@@ -1,8 +1,12 @@
 import unittest
+import warnings
+
 import numpy as np
 
 from simsopt.configs import get_data
+from simsopt.configs.LHD_like import get_LHD_like_data
 from simsopt.field import BiotSavart, Coil
+from simsopt.geo import CurveXYZFourier
 import simsoptpp as sopp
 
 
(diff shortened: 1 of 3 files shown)


krystophny:
@akaptano could you review? Test-only PR — behavioural coverage for `SimsoptRequires`/`deprecated` (`_core/dev.py`), `initialize_logging` (`util/logger.py`), and the `get_LHD_like_data` deprecation shim. Split out of #637 so the codecov-gate change can merge independently.

krystophny:
Dropped `test_warning_message_includes_replacement_staticmethod`: it exercised the same `isinstance(replacement, (classmethod, staticmethod))` branch as the classmethod test, so it added no coverage. Remaining tests each drive a public entry point and assert observable behaviour.

mbkumar (approved):
@krystophny 

Thanks for the tests!

PR #640, opened 2026-07-15
B field rect section v2

t-atane:
I implemented the full B field for self-field in the winding pack according to the paper of Dr. Landreman
(https://iopscience.iop.org/article/10.1088/1741-4326/adb04e/meta).

Currently, the simsopt repository only includes B_regularized in src/simsopt/field/selffield.py (and also in the coil class in coil.py). I added B_0, B_kappa and B_b. 

There are 3 main additions (the rest is due to autopep and ruff):
in selffield.py : the direct implementation of the different B field computation (from line 174)
in coil.py: the three B field contributions are added in the class "RectangularRegulaziedCoil" (from line 266)
in test_selffieldforces.py : added some assert to ensure that the tests pass for the new B fields (line 566 to 589)

This code has already been used by at IPP (Garching, Germany) for a paper about Ic-optimisation in the winding pack (currently in writing).

I remain available for any question.

Kind regards

### DIFF
--- a/src/simsopt/field/coil.py
+++ b/src/simsopt/field/coil.py
@@ -7,20 +7,29 @@ from simsopt.geo.curvexyzfourier import CurveXYZFourier
 from simsopt.geo.curve import RotatedCurve
 import simsoptpp as sopp
 
-__all__ = ['Coil', 'RegularizedCoil', 'CircularRegularizedCoil', 'RectangularRegularizedCoil',
-           'Current', 'coils_via_symmetries',
-           'load_coils_from_makegrid_file',
-           'apply_symmetries_to_currents', 'apply_symmetries_to_curves',
-           'coils_to_makegrid', 'coils_to_focus', 'coils_to_vtk'
-           ]
+
+__all__ = [
+    "Coil",
+    "RegularizedCoil",
+    "CircularRegularizedCoil",
+    "RectangularRegularizedCoil",
+    "Current",
+    "coils_via_symmetries",
+    "load_coils_from_makegrid_file",
+    "apply_symmetries_to_currents",
+    "apply_symmetries_to_curves",
+    "coils_to_makegrid",
+    "coils_to_focus",
+    "coils_to_vtk",
+]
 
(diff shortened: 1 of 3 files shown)


mbkumar:
@t-atane  Can you please fix the linter errors?

t-atane:
I mischanged the function coils_to_vtk in coil.py, which was not intentional. I thus restored the original function as it was. From all the unittests this was the only error I saw

t-atane:
Are there other problems you were talking about @mbkumar ?

t-atane:
@mbkumar The pointed linting errors are not in the files I changed (I only changed src/simsopt/coil.py and selffield.py), so I am pretty confused. Do you want me to fix them anyway ?

mbkumar:
@t-atane  No need. I realized the errors thrown by the linter could be due to new rule enforcements by ruff. I'll investigate it.

src/simsopt/field/coil.py
```diff
@@ -249,19 +267,92 @@ class RectangularRegularizedCoil(RegularizedCoil):
...
     def __init__(self, curve, current, a, b):
         from .selffield import regularization_rect
+
         regularization = regularization_rect(a, b)
         self.a = a
         self.b = b
-        RegularizedCoil.__init__(self, curve, current, regularization)    
+        RegularizedCoil.__init__(self, curve, current, regularization)
+
+    def B_total(self, u, v, a, b):
```
mishapadidar, line 286: I may misunderstand, this can already be computed by building a BiotSavart object, not?

t-atane: Dear mishapadidar,

Thanks for the reply.
I would say no, because the Biot-Savart object does not take into account all finite-build effects in the winding pack itself.

Or could you please show the code from biotsavart that computes this ?

Thank you,
t-atane

mishapadidar: Okay, best if @akaptano gives review on this, since he wrote most of the current force/torque work.

Issue #641, opened 2026-07-18
VirtualCasing.B_external_normal_extended uses wrong reflection indices on cell-centred grids


jungdaesuh:
## Problem

With `use_stellsym=True`, `VirtualCasing.from_vmec()` evaluates
`B_external_normal` on a cell-centred half-period grid, but its full-torus
extension uses endpoint-based `np.flip` indexing:

<https://github.com/hiddenSymmetries/simsopt/blob/51b22454f5a0faffbfc712b4a9a8d45738f12781/src/simsopt/mhd/virtual_casing.py#L272-L275>

For reflected row `N + r`, the correct source row is `N - 1 - r`. The current
implementation effectively uses `N - r`, shifting the reflected half-period by
one toroidal row.

Affects master at `51b22454`, release `v1.10.6`, and versions since PR #353.

## Reproducer

```python
import numpy as np

nfp, nphi, ntheta = 5, 85, 64
phi_half = (np.arange(nphi) + 0.5) / (2 * nfp * nphi)
phi_full = (np.arange(2 * nfp * nphi) + 0.5) / (2 * nfp * nphi)
theta = np.arange(ntheta) / ntheta

def field(phi, theta):
    return (
        np.sin(2*np.pi*(2*theta[None, :] - nfp*phi[:, None]))
        + 0.3*np.sin(2*np.pi*(3*theta[None, :] + 2*nfp*phi[:, None]))
    )

bn = field(phi_half, theta)
truth = field(phi_full, theta)

wrapped = np.hstack((bn, bn[:, [0]]))
wrapped = np.vstack((wrapped, -np.flip(np.flip(wrapped, 0), 1)[0]))
flipped = -np.flip(np.flip(wrapped, 0), 1)
current = np.concatenate([
    np.concatenate((bn, flipped[:-1, :-1])) for _ in range(nfp)
])

idx_theta = (-np.arange(ntheta)) % ntheta
corrected = np.tile(
    np.concatenate((bn, -bn[::-1][:, idx_theta])), (nfp, 1)
)

print(np.max(np.abs(current - truth)))
print(np.max(np.abs(corrected - truth)))
```

Output:

```text
0.05912870955174658
1.0574874309554616e-14
```

## Proposed fix

Use `-Bexternal_normal[::-1][:, (-np.arange(trgt_ntheta)) % trgt_ntheta]`
for the stellarator-symmetric reflection. When `use_stellsym=False`, the array
[...]

PR #645, opened 2026-07-22
Fix B_external_normal_extended

landreman:
This PR addresses issue #641. The `B_external_normal_extended` data in the virtual casing object was incorrect. Usually for finite-beta coil optimization we only use `B_external_normal` which is on a half field period, whereas `B_external_normal_extended` is meant to be the same data copied several times to cover the full torus. So I don't think this bug affected previous coil optimizations, thankfully. The implemented change follows the recommended solution in #641. Test coverage was added, and the new assertion fails for the old virtual_casing.py but passed for the corrected version.

### DIFF
--- a/src/simsopt/mhd/virtual_casing.py
+++ b/src/simsopt/mhd/virtual_casing.py
@@ -328,6 +328,11 @@ class VirtualCasing:
             trgt_nphi.description = 'Number of grid points in the toroidal angle phi for resulting B_external'
             trgt_nphi.units = 'Dimensionless'
 
+            trgt_nphi_extended = f.createVariable('trgt_nphi_extended', 'i', tuple())
+            trgt_nphi_extended.data[()] = self.trgt_nphi_extended
+            trgt_nphi_extended.description = 'Number of grid points in the toroidal angle phi for resulting B_external (extended to full torus)'
+            trgt_nphi_extended.units = 'Dimensionless'
+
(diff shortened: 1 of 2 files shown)


landreman:
@rogeriojorge This PR changes some lines you wrote in #353. See if you agree with the proposed change.

src/simsopt/mhd/virtual_casing.py
```diff
@@ -269,10 +271,17 @@ def from_vmec(cls, vmec, src_nphi, src_ntheta=None, trgt_nphi=None, trgt_ntheta=
...
+            # Rotate the first half-period by 180 degrees and flip sign, being careful that the theta grid
+            # includes theta=0 but not theta=1. See github issue #641.
+            vc.trgt_nphi_extended = 2 * nfp * trgt_nphi
+            idx_theta = (-np.arange(trgt_ntheta)) % trgt_ntheta
+            vc.B_external_normal_extended = np.tile(
+                np.concatenate((Bexternal_normal, -Bexternal_normal[::-1][:, idx_theta])), (nfp, 1)
+            )
+        else:
+            vc.trgt_nphi_extended = nfp * trgt_nphi
+            vc.B_external_normal_extended = np.tile(Bexternal_normal, (nfp, 1))
```
rogeriojorge, line 284: Clever!

rogeriojorge (approved):
Looks good to me!

Issue #644, opened 2026-07-22
Fix independent optional vector-potential components in MGrid


krystophny:
# Fix independent optional vector-potential components in `MGrid`

## Summary

`MGrid.add_field_cylindrical()` documents `ar`, `ap`, and `az` as independently
optional.  The presence checks for `ap` and `az` are crossed, so supplying
only `A_phi` omits `ap_001` and writes `az_001=NaN` instead.

Audited current upstream revision:
`51b22454f5a0faffbfc712b4a9a8d45738f12781`.

## Source contradiction

At `src/simsopt/field/mgrid.py:123-126` the code is equivalent to

```python
if ap is not None:
    self.az_arr.append(az)
if az is not None:
    self.ap_arr.append(ap)
```

The values appear in the correct NetCDF components when all three are
supplied, which lets the usual `MagneticField.to_mgrid(include_potential=True)`
path pass.  The independently optional public API is nevertheless broken.

## Executable discriminator

With one grid cell and only `ap=5` supplied, the written file contains

```text
ap_001 absent
az_001 present with NaN
```

The same defect is symmetric for an `az`-only input.  The external reproducer
also confirms that full `(A_R,A_phi,A_Z)=(4,5,6)` and magnetic
`(B_R,B_phi,B_Z)=(1,2,3)` preserve their component order, so this issue is
limited to the crossed presence guards.

## Proposed resolution

Guard and append each component with its own variable:

```python
if ap is not None:
    self.ap_arr.append(ap)
if az is not None:
    self.az_arr.append(az)
```

## Acceptance criteria

- `ar`-, `ap`-, and `az`-only inputs each write only the corresponding NetCDF
  variable with finite values.
- All-three input remains bit-for-bit component-consistent.
- Mixed availability across current groups is either validated/rejected or
  explicitly supported; list length alone must not silently misalign groups.

landreman:
Resolved by #646

PR #646, opened 2026-07-22
Fix issue 644: p vs z mixup in mgrid

landreman:
Fixes issue #644

### DIFF
--- a/src/simsopt/field/mgrid.py
+++ b/src/simsopt/field/mgrid.py
@@ -121,9 +121,9 @@ class MGrid():
         if ar is not None:
             self.ar_arr.append(ar)
         if ap is not None:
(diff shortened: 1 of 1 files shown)


src/simsopt/field/mgrid.py
```diff
@@ -121,9 +121,9 @@ def add_field_cylindrical(self, br, bp, bz, ar=None, ap=None, az=None, name=None
         if ar is not None:
             self.ar_arr.append(ar)
         if ap is not None:
-            self.az_arr.append(az)
-        if az is not None:
             self.ap_arr.append(ap)
```
mishapadidar, line 124: oof. thats a bug. good catch

mishapadidar (approved):

PR #650, opened 2026-08-16
Add Windows and Python 3.14~3.15 wheels in CI

lzpel:
This PR add Windows and Python 3.14~3.15 wheels

- Update pypa/cibuildwheel@v2.22.0 into pypa/cibuildwheel@v4.2.0, which extends the
    default build targets up to CPython 3.15
- Add an MSVC branch to the compiler flag selection. It has to come first because cibuildwheel inherits CI=true from the runner, which would otherwise send Windows through the -O3 -march=westmere path that cl.exe rejects. Append instead of overwrite so the CMake defaults (/EHsc) are kept, and add /D_USE_MATH_DEFINES for M_PI, /permissive- for the `and`/`not` alternative operators, and /bigobj for the large pybind11 translation units.
- Hand boost's bootstrap.sh to bash explicitly. ExternalProject runs its commands directly rather than through a shell, so a bare ./bootstrap.sh is not executable on Windows. The runner ships Git Bash and gcc, which is what b2's --guess-toolset needs.
- Replace the two VLAs in WireframeField::compute with std::vector; MSVC does not support variable length arrays.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -123,7 +123,7 @@ else()
             libs/range libs/iterator libs/concept_check libs/detail libs/function_types
             libs/lexical_cast libs/container libs/move libs/smart_ptr libs/multi_array
             libs/functional libs/function libs/type_index libs/container_hash libs/bind
-        CONFIGURE_COMMAND bash ./bootstrap.sh --prefix=<PREFIX>
+        CONFIGURE_COMMAND sh ./bootstrap.sh --prefix=<PREFIX>
         BUILD_COMMAND ./b2 headers --prefix=${boost_DIR}
         BUILD_IN_SOURCE 1
         INSTALL_COMMAND ""
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,7 +40,13 @@ configure_file(config.h.in config.h)
 include_directories(${CMAKE_CURRENT_BINARY_DIR})
 
 include(CheckCXXCompilerFlag)
-IF(DEFINED ENV{CI})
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -101,7 +103,7 @@ else()
             libs/range libs/iterator libs/concept_check libs/detail libs/function_types
             libs/lexical_cast libs/container libs/move libs/smart_ptr libs/multi_array
             libs/functional libs/function libs/type_index libs/container_hash libs/bind
-        CONFIGURE_COMMAND ./bootstrap.sh --prefix=<PREFIX>
+        CONFIGURE_COMMAND bash ./bootstrap.sh --prefix=<PREFIX>
```
mbkumar: Use `sh` in place of `bash`, which is more portable.

mbkumar (changes requested):
Thanks for the contribution! Minor change is needed because `bash` is not as portable as `sh`.

lzpel:
@mbkumar Thank you for the review. I fixed it.

mbkumar:
@lzpel  Can you also update the test workflow to run tests on windows?

lzpel:
Sorry, I gave up on running the test workflow on windows
Making the VMEC2000 and SPEC tests work on windows require too many changes to review

mbkumar:
@lzpel Those tests should be skipped on Windows. SPEC and VMEC are not expected to run on Windows. We used to test against those packages not being available. That framework is still in place. You can skip running the examples also.

PR #651, opened 2026-08-17
XSIMD update

mbkumar:
Updated the xtensor stack (xtensor, xsimd, xtensor-python, xtl) to the latest releases. 
Updated the C++ standard to 20 to conform with the xtensor standard.

Minor modifications were needed in the C++ code due to the API changes in XSIMD.

### DIFF
--- a/src/simsoptpp/biot_savart_impl.h
+++ b/src/simsoptpp/biot_savart_impl.h
@@ -1,7 +1,7 @@
 #include "simdhelpers.h"
 #include "vec3dsimd.h"
 #include <stdexcept>
-#include "xtensor/xlayout.hpp"
+#include "xtensor/core/xlayout.hpp"
 
 using namespace std;
 // When compiled with C++17, then we use `if constexpr` to check for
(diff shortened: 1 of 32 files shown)


src/simsoptpp/biot_savart_impl.h
```diff
@@ -138,28 +138,47 @@ void biot_savart_kernel(AlignedPaddedVec& pointsx, AlignedPaddedVec& pointsy, Al
...
+            for(int idx=0; idx<9; idx++)
+                d2B_dXdX_i[idx].store_aligned(d2B_dXdX_i_x[idx], d2B_dXdX_i_y[idx], d2B_dXdX_i_z[idx]);
+        }
+
         int jlimit = std::min(simd_size, num_points-i);
         for(int j=0; j<jlimit; j++){
-            B(i+j, 0) = fak * B_i.x[j];
-            B(i+j, 1) = fak * B_i.y[j];
-            B(i+j, 2) = fak * B_i.z[j];
+            B(i+j, 0) = fak * B_i_x[j];
```
andrewgiuliani, line 161: any reason why we're getting rid of the magnetic field structure B.x, B.y, B.z etc? fine with me

src/simsoptpp/biot_savart_vjp_impl.h
```diff
@@ -101,12 +115,12 @@ void biot_savart_vjp_kernel(AlignedPaddedVec& pointsx, AlignedPaddedVec& pointsy
...
-                res_grad_dgamma_by_dphi_ptr[3*j+0] += xsimd::hadd(res_grad_dgamma_by_dphi_add.x);
-                res_grad_dgamma_by_dphi_ptr[3*j+1] += xsimd::hadd(res_grad_dgamma_by_dphi_add.y);
-                res_grad_dgamma_by_dphi_ptr[3*j+2] += xsimd::hadd(res_grad_dgamma_by_dphi_add.z);
-                res_grad_gamma_ptr[3*j+0] += xsimd::hadd(res_grad_gamma_add.x);
-                res_grad_gamma_ptr[3*j+1] += xsimd::hadd(res_grad_gamma_add.y);
-                res_grad_gamma_ptr[3*j+2] += xsimd::hadd(res_grad_gamma_add.z);
+                res_grad_dgamma_by_dphi_ptr[3*j+0] += xsimd::reduce_add(res_grad_dgamma_by_dphi_add.x);
+                res_grad_dgamma_by_dphi_ptr[3*j+1] += xsimd::reduce_add(res_grad_dgamma_by_dphi_add.y);
+                res_grad_dgamma_by_dphi_ptr[3*j+2] += xsimd::reduce_add(res_grad_dgamma_by_dphi_add.z);
+                res_grad_gamma_ptr[3*j+0] += xsimd::reduce_add(res_grad_gamma_add.x);
```
andrewgiuliani, line 121: here we didn't get rid of the .x, .y, .z construct?

andrewgiuliani (approved):
minor comments, looks fine otherwise, thanks Bharat.

src/simsoptpp/biot_savart_impl.h
mbkumar: We are not getting rid of it. The newer versions of xsimd do not support array indexing for `xsimd::simd_type`, so we can not use `B_i.x[j]`. Wherever indexing for `xsimd::simd_type` object occurs, a workaround is used. A temporary memory-aligned `double tmp_array[simdcount]` array is created, and   `xsimd::simd_t.store_aligned(tmp_array)` call is made to store the simd_type object into the traditional array. Then the array indexing is used on the traditional array object.

Before the for loop, we have this line 
```cpp
B_i.store_aligned(B_i_x, B_i_y, B_i_z);
```
That line stores three `xsimd::simd_type` objects into three C arrays, allowing us to use array indexing.

src/simsoptpp/biot_savart_vjp_impl.h
mbkumar: No need to use the workaround here because indexing is not used on `simd_type` objects.

src/simsoptpp/biot_savart_impl.h
andrewgiuliani: thanks for explaining, all good on my end

PR #656, opened 2026-08-23
Preserve the inclusive highest NormalField mode

### LINKED ISSUE #643: Preserve the inclusive highest mpol row in NormalField array methods
# Preserve the inclusive highest `mpol` row in `NormalField` array methods

## Summary

`NormalField` documents arrays of shape `(mpol+1, 2*ntor+1)`, and its scalar
getters accept `m=mpol`.  However, all array getters and setters slice
`[0:mpol]`, excluding that last row.  The real-space reconstruction therefore
silently drops every `m=mpol` harmonic.

Audited current upstream revision:
`51b22454f5a0faffbfc712b4a9a8d45738f12781`.

## Executable discriminator

```python
nf = NormalField(nfp=1, stellsym=False, mpol=3, ntor=2)
nf.set_vns(3, 1, 7.0)

assert nf.vns.shape == (4, 5)
assert nf.get_vns(3, 1) == 7.0
print(nf.get_vns_asarray().shape)  # current result: (3, 5)
print(abs(nf.get_real_space_field()).max())  # current result: 0.0
```

The source locations are
`src/simsopt/field/normal_field.py:401-508`.  The same exclusive slice occurs
in `get_vns_asarray`, `get_vnc_asarray`, `set_vns_asarray`, and
`set_vnc_asarray`; the zero fallback for `vnc` also allocates `mpol` rather
than `mpol+1` rows.

This is convention-relevant because these are SPEC normal-field Fourier
coefficients.  Losing the boundary mode changes the represented normal field,
not just metadata.

## Proposed resolution

- Slice poloidal rows as `0:mpol+1` in all four array methods.
- Allocate the zero `vnc` fallback with `mpol+1` rows.
- Clarify that the `mpol` method argument is an inclusive maximum mode, as it
  already is elsewhere in the class.

## Acceptance criteria

- Round-trip arrays have shape `(mpol+1, 2*ntor+1)`.
- Nonzero sine and cosine coefficients at `m=mpol`, including both signs of
  `n`, survive scalar-to-array-to-real-space round trips.
- The real-space regression uses the declared
  `sin(m*theta-n*Nfp*phi)` / `cos(...)` phase and the SPEC oriented-area sign.

krystophny:
Fixes #643

Treat mpol as the inclusive maximum in all NormalField array getters and setters, including the stellarator-symmetric cosine fallback. The regression covers both signs of n, array round trips, and the SPEC real-space phase and sign.

Validation:
- NormalField suite: 19 passed, 10 skipped
- independent nfp=1 and nfp=3 reconstruction oracles passed
- ruff passed on changed files

### DIFF
--- a/src/simsopt/field/normal_field.py
+++ b/src/simsopt/field/normal_field.py
@@ -400,7 +400,8 @@ class NormalField(Optimizable):
 
     def get_vns_asarray(self, mpol=None, ntor=None):
         """
-        Return the vns as a single array
+        Return the vns as a single array. ``mpol`` is the inclusive maximum
+        poloidal mode.
         """
         if mpol is None:
             mpol = self.mpol
@@ -414,11 +415,12 @@ class NormalField(Optimizable):
 
         vns = self.vns
 
-        return vns[0:mpol, self.ntor-ntor:self.ntor+ntor+1]
+        return vns[0:mpol+1, self.ntor-ntor:self.ntor+ntor+1]
 
     def get_vnc_asarray(self, mpol=None, ntor=None):
         """
-        Return the vnc as a single array
+        Return the vnc as a single array. ``mpol`` is the inclusive maximum
+        poloidal mode.
         """
         if mpol is None:
             mpol = self.mpol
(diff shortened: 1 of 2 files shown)


krystophny:
Inclusive highest SPEC-mode preservation is implemented. CI is 24/24 passing with the expected `upload_pypi` skip, and the round-trip/reconstruction checks pass. Ready for review.

Chris&GPT

jurasic-pf (approved):

PR #657, opened 2026-08-26
periodic fieldlines

andrewgiuliani:
## `PeriodicFieldLine` for computing periodic field lines of a coil set

Adds `simsopt.geo.PeriodicFieldLine`, which solves for a periodic field line of a `BiotSavart` field, i.e. a closed field line that is a fixed point of the field-line return map (the magnetic axis, or an X-point of an island chain).

- **`src/simsopt/geo/periodicfieldline.py`** — solves the field-line residual `γ'/L − B/|B| = 0` using a `CurveXYZFourierSymmetries` parametrization with a Newton iteration. Supports `ntor > 1` field lines that close only after several toroidal transits, and provides a coil/current `vjp` for use in single stage optimization.

- **Examples** (`examples/1_Simple/`):
  - `periodicfieldline_QA.py` — magnetic axis (`ntor = 1`) of `STAR_Lite-A_low`.
  - `periodicfieldline_QH.py` — magnetic axis (`ntor = 1`) and two X-points (`ntor = 2`) of a QH configuration in [QUASR ID=2228222](https://quasr.flatironinstitute.org/model/2228222).


src/simsopt/geo/periodicfieldline.py
```diff
@@ -0,0 +1,197 @@
...
+
+    dres_dB = dres_dB.reshape((-1, 3, 3)) # the final equation if stellsym==False is just the curve label, which doesn't depend on B
+    lm_cons = lmask.reshape((-1, 3))
+
+    lm_times_dres_dB = np.sum(lm_cons[:, :, None] * dres_dB, axis=1).reshape((-1, 3))
+    lm_times_dres_dcoils = biotsavart.B_vjp(lm_times_dres_dB)
+    return lm_times_dres_dcoils
+
+
+class PeriodicFieldLine(Optimizable):
```
Rolopez77, line 66: Could this class get a docstring describing the supported curve type and point-count requirement? The number of residual entries selected by the solver is computed from `curve.order`, so the implementation appears to assume a `CurveXYZFourierSymmetries` with exactly `2 * order + 1` evaluation points. For example, an order-8 curve works with 17 points, while supplying 18 points currently results in a boolean-mask `IndexError`. If this point count is required, could the constructor check it and raise a more descriptive `ValueError`?

It would also help to document that the Newton solve requires an initial curve sufficiently close to the desired periodic field line. Could this class also be exported through `simsopt.geo` and added to the API documentation?

src/simsopt/geo/periodicfieldline.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def __init__(self, biotsavart, curve, options=None):
+        super().__init__(depends_on=[biotsavart])
```
Rolopez77, line 69: Could you clarify how changes to the supplied curve are intended to interact with the cached result? Using the public STAR_Lite-A_low configuration, I first obtained a successful solution and then changed one curve coefficient by 1e-2. need_to_run_code remained False, and calling run_code() again returned None, even though the actual residual increased from 4.44e-15 to 2.19e-2.

Also saw a related retry behavior after an unsuccessful solve with maxiter=1. Increasing newton_maxiter and calling run_code() again returned None rather than retrying.

Since only biotsavart is registered as a dependency here, should changes to the supplied curve or solver options invalidate the cached result?

src/simsopt/geo/periodicfieldline.py
```diff
@@ -0,0 +1,197 @@
...
+        mask = self.get_stellsym_mask()
+
+        if length is None:
+            length = CurveLength(self.curve).J()
+
+        x = np.concatenate((curve.get_dofs(), [length]))
+        i = 0
+
+        r, J, _ = field_line_residual(curve, length, self.biotsavart)
+        b = r[mask]
```
Rolopez77, line 169: Using the public STAR_Lite-A_low configuration, I increased the first coil current by 1% while retaining a curve with stellsym=True. The selected residual equations converged to 2.28e-15 and the solver returned success=True, while the full residual remained 1.60e-3.

What should the required symmetry be here? If stellsym=True requires the magnetic field to have the same symmetry, could that requirement be documented or checked? Should success also require the full residual to be below tolerance?

src/simsopt/geo/periodicfieldline.py
```diff
@@ -0,0 +1,197 @@
...
+            curve.set_dofs(x[:-1])
+            length = x[-1]
+            i += 1
+            r, J, _ = field_line_residual(curve, length, self.biotsavart)
+            b = r[mask]
+            J = J[mask]
+
+        P, L, U = lu(J)
+        res = {
+            "residual": r, "jacobian": J, "iter": i, "success": norm <= tol, "length": length, "PLU": (P, L, U),
```
Rolopez77, line 189: Reproduced a final-iteration status edge case using the public STAR_Lite-A_low configuration at commit b06761e. A normal solve converges in three Newton updates. Repeating the same solve with newton_maxiter=3 produces a final masked and full residual of 4.44e-15 for tol=1e-12, but reports success=False.

Seems like  the norm is evaluated before the Newton update and is not recomputed after the final residual is calculated when the loop exits at maxiter.

Rolopez77 (commented):
I reviewed commit `b06761e` at both the numerical and source/API levels.

- The PR-specific suite passes: 8 tests plus 3 subtests
- The current downstream project suite passes: 175 tests
- Both new examples run successfully
- The public STAR_Lite-A_low example converges successfully in three Newton updates
- An independent finite-difference check agrees with the derivative used by the solver

Left inline comments on a reproduced final-update success-flag issue, the behavior of cached results when the curve or solver options change, the required relationship between curve and magnetic-field symmetry, and the public API/input documentation.

src/simsopt/geo/periodicfieldline.py
andrewgiuliani: addressed, thanks!

src/simsopt/geo/periodicfieldline.py
andrewgiuliani: ok, agreed here, but BoozerSurfaces function in the same manner (modifying boozer_surface.surface.x does not uncache anything).  I need to think of a cleaner solution than adding curve to the depends_on list, because this modifies the dofs of the PeriodicFiedLine class.

I'll try and address this retry behaviour that you're pointing out in a follow up PR, since BoozerSurface should function in the same manner.

src/simsopt/geo/periodicfieldline.py
andrewgiuliani: I suspect here you have made the magnetic field nonstellsym, while the periodicfieldline is stellsym, so this is expected behaviour.  I've added something to the docstring to highlight that the user should ensure the biotsavart field has the requisite symmetries.  

I have changed the requirement that success show the full residual is below tolerance.

src/simsopt/geo/periodicfieldline.py
andrewgiuliani: good catch! :) this should be addressed in the latest commit

src/simsopt/geo/periodicfieldline.py
```diff
@@ -0,0 +1,278 @@
...
+
+    to zero, where :math:`\boldsymbol\gamma` is the curve, :math:`L` is the
+    length of the field line and :math:`\mathbf B` is the magnetic field. The
+    residual vanishes when the tangent of the curve is everywhere parallel to
+    :math:`\mathbf B`, so that the curve is a field line, and when the curve is
+    parametrized proportionally to arclength. The degrees of freedom are the
+    curve coefficients together with :math:`L`.
+
+    Args:
+        biotsavart: the :obj:`~simsopt.field.BiotSavart` magnetic field.
```
mishapadidar: nit: User args notation like
`biotsavart (BiotSavart): ...`
see example in [coil](https://simsopt.readthedocs.io/latest/simsopt_user.field.html#simsopt.field.Coil) class

src/simsopt/geo/periodicfieldline.py
```diff
@@ -0,0 +1,278 @@
...
+            options['newton_tol'] = 1e-13
+        if 'newton_maxiter' not in options:
+            options['newton_maxiter'] = 40
+        self.options = options
+
+        
+    def recompute_bell(self, parent=None):
+        self.need_to_run_code = True
+    
+    def run_code(self, length):
```
mishapadidar, line 204: doc please

src/simsopt/geo/periodicfieldline.py
```diff
@@ -0,0 +1,278 @@
...
+            return mask
+
+        mask = np.ones((2*order+1, 3), dtype=bool)
+        if stellsym:
+            mask[0, 0] = False
+            mask[order+1:, :] = False
+        mask = mask.flatten()
+        return mask
+
+    def minimize_boozer_penalty_constraints_LBFGS(self, tol=1e-3, maxiter=1000, length=None, limited_memory=True, verbose=False):
```
mishapadidar, line 241: doc please

src/simsopt/geo/periodicfieldline.py
```diff
@@ -0,0 +1,278 @@
...
+
+        self.res = resdict
+        self.need_to_run_code = False
+
+        if verbose:
+            print(f"{method} solve - {resdict['success']}  iter={resdict['iter']}, length={resdict['length']:.8f}, ||grad||_inf = {np.linalg.norm(resdict['gradient'], ord=np.inf):.3e}", flush=True)
+        return resdict
+
+
+    def solve_residual_equation_exactly_newton(self, tol=1e-10, maxiter=10, length=None, verbose=False):
```
mishapadidar, line 289: doc please :)

src/simsopt/geo/periodicfieldline.py
```diff
@@ -0,0 +1,278 @@
...
+        self.options = options
+
+        
+    def recompute_bell(self, parent=None):
+        self.need_to_run_code = True
+    
+    def run_code(self, length):
+        if not self.need_to_run_code:
+            return
+        res = self.solve_residual_equation_exactly_newton(length=length, tol=self.options['newton_tol'], maxiter=self.options['newton_maxiter'], verbose=self.options['verbose'])
```
mishapadidar: It looks like `run_code` function only calls newton here. I think it makes sense to be able to call either solver through `run_code`. For instance you could make the call signature
```run_code(self, length, solver, options)```

**Review by mishapadidar — COMMENTED** [review]
Looks good, just minor changes. 
Just a friendly reminder, please make sure the documentation compiles correctly when finalizing.

File: src/simsopt/geo/periodicfieldline.py (hunk 7 above)
Comment (andrewgiuliani) [reply src/simsopt/geo/periodicfieldline.py:241]: done

File: src/simsopt/geo/periodicfieldline.py (hunk 6 above)
Comment (andrewgiuliani) [reply src/simsopt/geo/periodicfieldline.py:204]: done

File: src/simsopt/geo/periodicfieldline.py (hunk 5 above)
Comment (andrewgiuliani) [reply src/simsopt/geo/periodicfieldline.py]: done, but let me know if I missed something

File: src/simsopt/geo/periodicfieldline.py (hunk 9 above)
Comment (andrewgiuliani) [reply src/simsopt/geo/periodicfieldline.py]: done, thanks

File: src/simsopt/geo/periodicfieldline.py (hunk 8 above)
Comment (andrewgiuliani) [reply src/simsopt/geo/periodicfieldline.py:289]: done thanks

**Comment by andrewgiuliani** [comment]
below are screenshots of compiled documentation.  Let me know if this is what you are looking for @mishapadidar 

image
image
image

File: src/simsopt/geo/periodicfieldline.py [hunk 10]
```diff
@@ -0,0 +1,346 @@
...
+
+        self.res = resdict
+        self.need_to_run_code = False
+
+        if verbose:
+            print(f"{method} solve - {resdict['success']}  iter={resdict['iter']}, length={resdict['length']:.8f}, ||grad||_inf = {np.linalg.norm(resdict['gradient'], ord=np.inf):.3e}", flush=True)
+        return resdict
+
+
+    def solve_residual_equation_exactly_newton(self, tol=1e-10, maxiter=10, length=None, verbose=False):
```
mishapadidar, line 289: Looks like we write our own newton loop here. If we use [scipy's newton](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.newton.html) method we may be able to streamline the code (alternatively, we could just extract the newton loop from this function). For instance, we may be able to reduce the code down to one solve method `solve_residual_equation`, and `options` could just be 
`options={'maxiter':..., 'tol':...}, where we point users to the scipy page for the options. It would be pretty clean if we can make this work.

src/simsopt/geo/periodicfieldline.py
```diff
@@ -0,0 +1,346 @@
...
+            length = CurveLength(self.curve).J()
+
+        x = np.concatenate((curve.get_dofs(), [length]))
+        i = 0
+
+        r, J, _ = field_line_residual(curve, length, self.biotsavart)
+        b = r[mask]
+        J = J[mask]
+
+        norm = 1e6
```
mishapadidar: i think we may be able to drop this initialization

mishapadidar (commented):
Just one more review comment. I ask whether we can consolidate the two solve functions into one function by separating out the newton into its own function or using scipy's.

src/simsopt/geo/periodicfieldline.py
andrewgiuliani: done

src/simsopt/geo/periodicfieldline.py
andrewgiuliani: thanks for the suggestion, and you're likely right, but it's a simple 10ish lines of code that I have used quite a bit and am confident in its behaviour.  I also like doing iterative refinement for each newton step because it improves accuracy, which won't be done by scipy's newton.  I'd also like to keep this class uniform with BoozerSurface.  

That said, I'm not against making these modifications later, but I'd like to do the for both periodicfieldluine and BoozerSurface classes.

mishapadidar (approved):

